PR #92, opened 2013-07-29
Prototype threaded writing interface and code for diginorm and filter-abund.

ctb:


ctb:
I think we can close this now ;)

PR #141, opened 2013-08-30
Documentation on how to think about comparing assemblies

ctb:


ctb:
Meh. Ongoing, but not really something we need in our docs ;).

PR #218, opened 2013-11-15
Sparse median

camillescott:
Open pull req for sparse median diginorm.

Thus far there is a new c++ function, and a new python script making use of it.

Algorithm is essentially the same; pseudocode:

```
if sparse_median_count(read) < COV_CUTOFF:
    consume(read)
else:
    continue
```

TODO:
* [ ] performance testing ipython notebook
* [x] implement version of normal median without std dev and avg calcs
* [ ] tests!
* [x] fold normalize-by-sparse-median into normalize-by-median, add a flag

Other thoughts, @ctb @mr-c?


ctb:
ping @camillescott should this be closed now?

ctb:
Done better in #862.

PR #333, opened 2014-03-19
initial stab at exception handling around file operations

ctb:


tests/test_hashbits.py
```diff
@@ -547,3 +548,223 @@ def test_simple_median():
...
+
+def test_save_load_tagset_trunc():
+    ht = khmer.new_hashbits(32, 1, 1)
+
+    outfile = utils.get_temp_filename('tagset')
+
+    ht.add_tag('A' * 32)
+    ht.add_tag('G' * 32)
+    ht.save_tagset(outfile)
+    ht.save_tagset('/tmp/goodversion-k32.tagset')
```
mr-c: Hmm.. I don't think this line should be in the test.

Issue #60, opened 2013-05-22
Allow k > 32 in khmer


ctb:
There have been a number of requests to enable k > 32 in khmer code.  This should be relatively straightforward (@fishjord has a cyclic hash implementation in Java that he sent around) but has not so far been a priority.  This is an issue so that we can track such requests in the future and keep it somewhere on the list.

Note, we need to figure out how to deal with _revhash issues; there are a few places in the code where exact hashes are kept and converted from numbers back into strings.  Not a blocker, just a detail issue.

mr-c:
See also https://twitter.com/beconstant/status/555750390323765248

alexjironkin:
+1

ctb:
Also see: https://twitter.com/ChristianHeimes/status/593007790861475840

ctb:
Now enabled by #1511 and friends.

Issue #32, opened 2013-03-07
Create 'iter_read_pairs' Method for 'ReadParser' Objects


emcd:
The logic to support read pairs in 'normalize-by-median.py' works, but read pairs could be guaranteed from the C++ side, which would remove much of the verification and bookkeeping logic on the Python side. In the vein of Python 2.x dictionary iterators, I think that a 'iter_read_pairs' method might be appropriate. If we do this, then probably we want to alias the existing iterator function with 'iter_reads' (makes selection of the appropriate method with 'getattr' a little nicer, for those, such as myself, who write code that way).

Another approach would be to create a 'PairedReadParser' family of classes, but I think that this is overkill.

PR #624, opened 2014-10-03
Feature/cyclic hash

wrightmhw:
Adding in Cyclic Hash for K > 32 as per issues #60.

### DIFF
--- a/lib/kmer_hash.cc
+++ b/lib/kmer_hash.cc
@@ -85,7 +85,7 @@ HashIntoType _revcyclichash(const char * kmer_string, const WordLength k)
     // Our list of random numbers for each base value
     // Hard coding in random vals from array
     // Random values generated from www.random.org
-    HashIntoType compliment_vals[] = {491989698, 127908739, 392233993, 303837760};
+    HashIntoType complement_vals[] = {491989698, 127908739, 392233993, 303837760};
     //std::vector<HashIntoType> compliment_random_base_vals (vals, vals + sizeof(vals) / sizeof(vals));
 
     HashIntoType hashvalue = 0;
@@ -99,21 +99,21 @@ HashIntoType _revcyclichash(const char * kmer_string, const WordLength k)
         switch (kmer_string[i]) {
         case 'A':
         case 'a':
-            hashvalue ^= compliment_vals[0];
+            hashvalue ^= complement_vals[0];
 	    break;
         case 'C':
         case 'c':
-            hashvalue ^= compliment_vals[1];
+            hashvalue ^= complement_vals[1];
 	    break;
         case 'G':
         case 'g':
-            hashvalue ^= compliment_vals[2];
+            hashvalue ^= complement_vals[2];
 	    break;
         case 'T':
         case 't':
         case 'U':
         case 'u':
-            hashvalue ^= compliment_vals[3];
+            hashvalue ^= complement_vals[3];
 	    break;
         default:
             throw khmer_exception("Unknown nucleotide, not one of AaCcGgTtUu.");
--- a/lib/kmer_hash.cc
+++ b/lib/kmer_hash.cc
@@ -38,7 +38,7 @@ HashIntoType _hash(const char * kmer, const WordLength k,
 }
 
 // Bitwise Shifting
-HashIntoType _cyclichash(std::string kmer_string, const WordLength k)
+HashIntoType _cyclichash(const char * kmer_string, const WordLength k)
 {
     // Our list of random numbers for each base value
     // Hard coding in random vals from array
@@ -49,7 +49,7 @@ HashIntoType _cyclichash(std::string kmer_string, const WordLength k)
     HashIntoType hashvalue = 0;
 
     // Iterate through the kmer, hashing each
-    for (int i = 0; i < kmer_string.length(); i++) {
+    for (int i = 0; i < strlen(kmer_string); i++) {
         // A circular bitshift (http://en.wikipedia.org/wiki/Circular_shift)
         hashvalue = (hashvalue << 1 | hashvalue >> (k - 1));
 
@@ -58,25 +58,29 @@ HashIntoType _cyclichash(std::string kmer_string, const WordLength k)
         case 'A':
         case 'a':
             hashvalue ^= vals[0];
+	    break;
         case 'C':
         case 'c':
             hashvalue ^= vals[1];
+	    break;
         case 'G':
         case 'g':
             hashvalue ^= vals[2];
+	    break;
         case 'T':
         case 't':
         case 'U':
         case 'u':
             hashvalue ^= vals[3];
+	    break;
         default:
-            throw khmer_exception();
+            throw khmer_exception("Unknown nucleotide, not one of AaCcGgTtUu.");
         }// of Switch
     }
     return hashvalue;
 }
 
-HashIntoType _revcyclichash(std::string kmer_string, const WordLength k)
+HashIntoType _revcyclichash(const char * kmer_string, const WordLength k)
 {
     // Our list of random numbers for each base value
     // Hard coding in random vals from array
@@ -87,7 +91,7 @@ HashIntoType _revcyclichash(std::string kmer_string, const WordLength k)
     HashIntoType hashvalue = 0;
 
     // Iterate through the kmer backwards, hashing each base
-    for (int i = kmer_string.length(); i > -1; i--) {
+    for (int i = strlen(kmer_string) - 1; i >= 0; i--) {
         // A circular bitshift (http://en.wikipedia.org/wiki/Circular_shift)
         hashvalue = (hashvalue << 1 | hashvalue >> (k - 1));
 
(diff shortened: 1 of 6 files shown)


mr-c:
Code compiles and runs; only 10 errors and 198 failures when running the tests!

mr-c:
Now 2 test errors and 206 failures; progress!

ctb:
:)

lib/kmer_hash.cc
```diff
@@ -31,23 +31,95 @@ HashIntoType _hash(const char * kmer, const WordLength k,
...
+}
 
-    _h = h;
-    _r = r;
+HashIntoType _revcyclichash(const char * kmer_string, const WordLength k)
+{
+    // Our list of random numbers for each base value
+    // Hard coding in random vals from array
+    // Random values generated from www.random.org
+    HashIntoType compliment_vals[] = {491989698, 127908739, 392233993, 303837760};
```
ctb: complement, not compliment :)

tests/test_sandbox_scripts.py
```diff
@@ -109,6 +109,10 @@ def test_sweep_reads():
 
     print os.listdir(in_dir)
 
+    assert os.path.exists(out1)
```
ctb: do these belong in a separate PR?

mr-c: I can pull them out separately if you want.

ctb: On Wed, Mar 11, 2015 at 03:52:14PM -0700, Michael R. Crusoe wrote:
> > @@ -109,6 +109,10 @@ def test_sweep_reads():
> >  
> >      print os.listdir(in_dir)
> >  
> > +    assert os.path.exists(out1)
> 
> I can pull them out separately if you want.

Just curious why they're part of this PR, which promises to be rather
far reaching - yes, might be better to split off the cleanup changes.

mr-c: Done: #889

mr-c:
Some thoughts from an unsent email:

Currently we use a 2-bit DNA encoding scheme as the hash function. It has the advantages of being 'perfect', reversible, and collision free. Its disadvantage the output scales in size with the k-mer size and currently the codebase limits this to 64 bits, or a k-mer size of 32.

The cyclic hash has the advantage of a fixed size and its speed scales linearly with the k-mer size. However it is not reversible and collisions can (and will) happen.

Other hashing algorithms (like smhasher) may have a speed advantage but will also have the collision and irreversibility problem. And most won't be able to produce hashes for all the kmers in a string without fully hashing each kmer -- the 2bit kmerIterator uses a bit shift and I see how to write a version for the cyclic hash that also reuses the previous hash.

It is important to remind ourselves that we don't actually care about the size of the output of the hash function as it is used to index into the bloom filter/count-min sketch tables by taking the modules of the table sizes.

Therefore we could quadruple the size of the hash type which would support k-mer sizes up to 128 bases without increasing the memory requirements.

ctb:
Additional note: we only care about reversibility for our sparse graph
implementation.  So if we allow multiple hash functions, selectable
at object creation, that would give us the flexibiltiy we need.

mr-c:
@ctb Okay. What about the collision issue?

ctb:
I think collisions only matter where we need reversibility. No?

mr-c:
A collision at the hashing stage guarantees a collision at the bit setting or count incrementing stage regardless of bloom filter or count-min sketch table number or sizes. Frankly we could skip hashing and take the numerical value of the string and get the table locations that way.

ctb:
On Tue, Mar 31, 2015 at 03:55:43AM -0700, Michael R. Crusoe wrote:
> A collision at the hashing stage guarantees a collision at the bit setting or count incrementing stage regardless of bloom filter or count-min sketch table number or sizes. Frankly we could skip hashing and take the numerical value of the string and get the table locations that way.

"Taking the numerical value of the string" *is* hashing, but it's non-random
and could lead to neighborhood structure.

I'm missing something here...? collisions at the hash level shouldn't matter
more than collisions at the CMS/Bloom filter stage, unless they're non-random.

mr-c:
On Tue, Mar 31, 2015 at 1:58 PM C. Titus Brown <notifications@github.com>
wrote:

> On Tue, Mar 31, 2015 at 03:55:43AM -0700, Michael R. Crusoe wrote:
> > A collision at the hashing stage guarantees a collision at the bit
> setting or count incrementing stage regardless of bloom filter or count-min
> sketch table number or sizes. Frankly we could skip hashing and take the
> numerical value of the string and get the table locations that way.
>
> "Taking the numerical value of the string" *is* hashing, but it's
> non-random
> and could lead to neighborhood structure.
>

Yep. Though the 2-bit representation is also non-random.

I'm missing something here...? collisions at the hash level shouldn't matter
> more than collisions at the CMS/Bloom filter stage, unless they're
> non-random.
>

If the hashing function produces the same output for two inputs it will
cause the same location to be chosen in the Bloom filter or Count-min
sketch for both inputs in all tables regardless of the parameters chosen.

If a collision happens due to the size of the bloom filter or count-min
sketch then the best case scenario is that it will happen only in one of
the tables due to their different sizes and the modulus function. (The
worst case being that the table is so small and/or crowded that collisions
occur in every table with a different colliding k-mer).

ctb:
On Tue, Mar 31, 2015 at 04:23:43AM -0700, Michael R. Crusoe wrote:
> On Tue, Mar 31, 2015 at 1:58 PM C. Titus Brown <notifications@github.com>
> wrote:
> 
> > On Tue, Mar 31, 2015 at 03:55:43AM -0700, Michael R. Crusoe wrote:
> > > A collision at the hashing stage guarantees a collision at the bit
> > setting or count incrementing stage regardless of bloom filter or count-min
> > sketch table number or sizes. Frankly we could skip hashing and take the
> > numerical value of the string and get the table locations that way.
> >
> > "Taking the numerical value of the string" *is* hashing, but it's
> > non-random
> > and could lead to neighborhood structure.
> >
> 
> Yep. Though the 2-bit representation is also non-random.

Yep.  Good reason to move away from it :)

> I'm missing something here...? collisions at the hash level shouldn't matter
> > more than collisions at the CMS/Bloom filter stage, unless they're
> > non-random.
> >
> 
> If the hashing function produces the same output for two inputs it will
> cause the same location to be chosen in the Bloom filter or Count-min
> sketch for both inputs in all tables regardless of the parameters chosen.
> 
> If a collision happens due to the size of the bloom filter or count-min
> sketch then the best case scenario is that it will happen only in one of
> the tables due to their different sizes and the modulus function. (The
> worst case being that the table is so small and/or crowded that collisions
> occur in every table with a different colliding k-mer).

Yes, but collisions from the hash function should be many, many orders of
magnitude less frequent than collisions from the limited size of the Bloom
filter.  For k=20, collisions from a random hash function will only
dominate if we have a hash table that's ~TB in size, right?

So unless there's structure in the hash function, there is no reason to
worry about this.

mr-c:
But we aren't talking about k=20 we're talking about k=32..150

ctb:
Hash collisions decrease or stay the same with increasing k, depending on what you're doing. For our applications they (at worse) stay the same.

But maybe I'm missing something. @camillescott, @luizirber?

--

camillescott:
If you're worried about exact collisions between smhasher's 128-bit output and our 4^k output, the odds of smhasher's output even falling with the same range as our 2-bit rep are *astronomically small*... 5.4210109e-20 when K=32, to be precise. And this is an *extreme* upper bound, to say the very least (used out of laziness ;)).

In general, the odds of the direct collisions you describe are terribly terribly tiny. When we talk about collisions in hashtables, we're usually talking about the actual placement algorithm, not the conversion from string to integer. Simple modulus is probably the worst placement method, but that's part of the point of the bloom filter -- so long as there isn't collision prior to placement, it works as intended, and we already know that collision prior to placement is extremely unlikely.

mr-c:
I worried about collisions between different k-mers using the cyclic hash
(64 bits) with a k-mer space of 4^k using k values up to 100 or 150.

camillescott:
Ah, I see -- it was unclear what sort of collisions you were talking about before (I read it as collisions between different hashing methods, not collisions in the cyclic impl). Then yes, if cyclic hash has a >64-bit domain and a 64-bit codomain, it necessarily will have collisions. I believe @ctb is confusing increasing k-mer diversity within a sample, which reduces collisions if all hashtable parameters are held constant, with a random sampling of a larger k-mer space.

ctb:
On Tue, Mar 31, 2015 at 09:31:46AM -0700, Camille Scott wrote:
> Ah, I see -- it was unclear what sort of collisions you were talking about before (I read it as collisions between different hashing methods, not collisions in the cyclic impl). Then yes, if cyclic hash has a >64-bit domain and a 64-bit codomain, it necessarily will have collisions. I believe @ctb is confusing increasing k-mer diversity within a sample, which reduces collisions if all hashtable parameters are held constant, with a random sampling of a larger k-mer space.

Update: talked with Camille, verified that we were on the same page, neither
of us sees a likely problem.

mr-c:
Glad you two chatted, unclear as to what you both don't see a problem with: switching to longer 2bit encoding as the hash function or the use of collision generating 64bit hash functions with k>32?

ctb:
Collisions, primarily. Reversibility is nice to have but not imo that high a priority for k > 32.

--

mr-c:
*shrug* I'll continue with my multi-hash refactor.

ctb:
(Can anyone tell me how #32 is relevant?)

mr-c:
Can't see any relevance from where I sit, no.

ctb:
Superseded by #1511 and related work.

PR #666, opened 2014-11-20
Master 'khmer' script which dynamically runs sub-commands

tseemann:
@ctb - here is the 'master' khmer command as requested. 

I have *not* integrated it into the install system. It expects to be in the same folder as the various *.py tools.

### DIFF
--- a/scripts/khmer
+++ b/scripts/khmer
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+# figure out where this script is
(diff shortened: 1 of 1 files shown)


mr-c:
Thanks @tseemann. Here's the script that will be shipping with the Debian package: http://anonscm.debian.org/cgit/debian-med/khmer.git/tree/debian/bin/khmer

tseemann:
Using that ```.``` as your `` <<HERE``` syntax threw me at first, but looks good!

PR #704, opened 2014-12-19
recast std::bad_alloc as a Python exception

### LINKED ISSUE #111: Put useful error message in when large hash tables fail to alloc.

mr-c:
Will fix #111

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,12 @@ CPPCHECK=ls lib/*.cc khmer/_khmermodule.cc | grep -v test | cppcheck -DNDEBUG \
 	 --quiet -Ilib -Ithird-party/bzip2 -Ithird-party/zlib \
 	 -Ithird-party/smhasher
 
+UNAME := $(shell uname)
+ifeq ($(UNAME),Linux)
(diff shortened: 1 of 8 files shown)


mr-c:
- [ ] Is it mergable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

mr-c:
Jenkins, test this please

mr-c:
@luizirber @ctb @camillescott Ready for review and merge

Makefile
```diff
@@ -162,7 +168,7 @@ diff-cover.html: coverage-gcovr.xml coverage.xml
 		--html-report diff-cover.html
 
 nosetests.xml: FORCE
-	./setup.py nosetests --with-xunit
+	./setup.py nosetests --with-xunit --attr ${TESTATTR}
```
ctb: does this mean we no longer need setup.cfg?

mr-c: setup.cfg provides other default values and acts as a fallback if the Setuptools nosetests command is run outside of the Makefile

PR #730, opened 2015-01-17
minor fixes to sandbox/correct-errors.py

ctb:


ctb:
- [ ] Is it mergable?
- [ ] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ctb:
Ready for review & merge.

sandbox/correct-errors.py
```diff
@@ -42,7 +42,12 @@ def output_single(read, new_sequence):
     if hasattr(read, 'accuracy'):
         accuracy = read.accuracy[:len(sequence)]
 
+        # in cases where sequence _lengthened_, need to truncate it to
+        # match the quality score length.
+        sequence = sequence[:len(accuracy)]
```
mr-c: Oi. This will lead to a shift between some bases and their corresponding quality value.

We should probably insert dummy quality values for the inserted bases.

ctb: On Mon, Jan 19, 2015 at 09:46:20AM -0800, Michael R. Crusoe wrote:
> > @@ -42,7 +42,12 @@ def output_single(read, new_sequence):
> >      if hasattr(read, 'accuracy'):
> >          accuracy = read.accuracy[:len(sequence)]
> >  
> > +        # in cases where sequence _lengthened_, need to truncate it to
> > +        # match the quality score length.
> > +        sequence = sequence[:len(accuracy)]
> 
> Oi. This will lead to a shift between some bases and their corresponding quality value.
> 
> We should probably insert dummy quality values for the inserted bases.

The shift was happening before, and causing bad FASTQ to be generated.
Probably the best way to deal with this is to output FASTA rather than
fake quality scores.

ctb:
@mr-c I'd like to get this merged; it will undoubtedly come up again in the context of error correction, but for now it causes not just semantic breakage but syntactic breakage as well. Comments?

ChangeLog
```diff
@@ -1,3 +1,9 @@
+2015-01-17  Titus Brown  <titus@idyll.org>
```
mr-c: Update the date please

mr-c:
Sure

Issue #732, opened 2015-01-18
User experience upgrades for khmer


ctb:
This issue is to track desired user experience improvements.  Most of these break I/O formats or file naming, so probably belong in khmer 2.0.  The question is, do we try to get 'em all in to khmer 2.0?

1. better filename naming, happening automatically - see #81.

1. output to gz/bz2 automatically, based on input format but overridable - see #505, with defaults determined by input format. ref #700 also.

1. exact kmer counting/automatic memory usage determination -
    a constant source of confusion for users and a constant documentation/tutorial challenge :).

    two complementary fixes: first, #347: count k-mers w/hyperloglog #233 #257 to choose the sizes of
    our high-efficiency data structures appropriately. This is trading computation/time for memory.
    Note that it will not work properly for streaming approaches like diginorm, where we can't know the
    final number of k-mers until after we do all the hard work!

   second, implement dynamically sized k-mer counting data structures, either exact or inexact or both.  For example,
   use google sparsehash #689 to count k-mers exactly.  Exact counting will be much less memory
   efficient than probabilistic data structures, but it's not clear that matters for many applications like
   diginorm, where streaming/sublinear memory is already a massive win.  Worth noting that exact
   counting would also be a useful addition to our Python API, and might also help overcome the
   concerns of people who dislike inexact counting (despite publications showing that
   you shouldn't worry about it :).

1. handling “broken paired” format in input - this affects streaming (#700 #654 #393) and may also need to be addressed in #655.

1. handle both new and old formats for FASTQ sequences (#23 #305 #606).

1. Allow k > 32 - tracked in #60. Once HyperLog Log #257 is integrated, we will have murmurhash in the
    codebase; perhaps this can be used in addition to/instead of #624?

Anything I'm forgetting? The above list covers most of the complaints I've heard :)

chuckpr:
Some notes from my experience...

It would be great to have more or specifically more centralized documentation for diagnostic performance figures. This info can be found in various sources on the web but a single location where users can see the types of figures that can be used to assess the performance of  diginorm or partitioning would be useful. It would be nice to see a collection of example figures from report files generated by ```normalize-by-median.py``` for instance. What does successful normalization look like from real-world data? What does it look like when parameters need to be tweaked? 

More documentation/examples/tutorials for the Python API as opposed to the scripts.

[...]

PR #744, opened 2015-01-27
max memory parameter specification 

ctb:
see #732, @macmanes comment

### DIFF
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -56,9 +56,9 @@ def build_hash_args(descr=None, epilog=None):
 
 
 def update_memory_parameters(parser, args):
-    env_n_tables = os.environ.get('KHMER_N_TABLES', DEFAULT_N_TABLES)
-    env_tablesize = os.environ.get('KHMER_MIN_TABLESIZE',
-                                   DEFAULT_MIN_TABLESIZE)
+    env_n_tables = int(os.environ.get('KHMER_N_TABLES', DEFAULT_N_TABLES))
+    env_tablesize = float(os.environ.get('KHMER_MIN_TABLESIZE',
+                                   DEFAULT_MIN_TABLESIZE))
 
     # can't allow all three! (a, b, c)
     if args.n_tables and args.min_tablesize and args.max_memory:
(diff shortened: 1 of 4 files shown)


khmer/khmer_args.py
```diff
@@ -52,9 +52,36 @@ def build_hash_args(descr=None, epilog=None):
...
+
+    if args.max_memory is None:
+        return
+    elif (args.n_tables != env_n_tables or \
+          args.min_tablesize != env_tablesize):
+        raise Exception("ERROR: max_memory specified along with other "
+                        "memory parameters (-x or -N)")
+
+    n_tables = DEFAULT_N_TABLES
+    max_memory = args.max_memory
```
mr-c: why not just use the full name?

khmer/khmer_args.py
```diff
@@ -52,9 +52,36 @@ def build_hash_args(descr=None, epilog=None):
...
+
+    if parser.hashtype == 'counting':
+        tablesize = int(max_memory / float(n_tables))
```
mr-c: -> env_n_tables

khmer/khmer_args.py
```diff
@@ -46,15 +43,61 @@ def build_hash_args(descr=None, epilog=None):
...
+
+    # if none are specified, set to defaults. (not a, not b, not c)
+    if not (args.n_tables or args.max_memory or args.min_tablesize):
+        args.n_tables = env_n_tables
+        args.min_tablesize = env_tablesize
+
+    # just tablesize and max memory (a, b, not c)
+    if args.n_tables is None and (args.max_memory and args.min_tablesize):
+        if parser.hashtype == 'counting':
+            tablesize = int(args.max_memory / float(args.min_tablesize))
```
mr-c: this should be `args.n_tables = int(args.max_memory / float(args.min_tablesize))` not tablesize

khmer/khmer_args.py
```diff
@@ -46,15 +43,61 @@ def build_hash_args(descr=None, epilog=None):
...
+        elif parser.hashtype == 'hashbits':
+            tablesize = int(8 * args.max_memory / float(args.min_tablesize))
```
mr-c: this should be `args.n_tables = int(8 * args.max_memory / float(args.min_tablesize))` not tablesize

khmer/khmer_args.py
```diff
@@ -46,15 +43,61 @@ def build_hash_args(descr=None, epilog=None):
     parser.add_argument('--ksize', '-k', type=int, default=env_ksize,
                         help='k-mer size to use')
     parser.add_argument('--n_tables', '-N', type=int,
-                        default=env_n_tables,
+                        default=None,
```
mr-c: This is effectively changing the command line API. Now a test is failing because we call 'normalize-by-median' with a `--min-tablesize` but without `--n_tables` as it relied on the default value.

ctb:
Redoing for 2.0.

Issue #754, opened 2015-02-04
Move trim-low-abund into scripts/


ctb:
trim-low-abund from #601 should be upgraded.

This will involve:
* load-save counting table
* reading and handling broken paired
* adding script tests

Good issue for sprint #751 and helps address broken paired in #732.

ctb:
Resolved in #759.

Issue #733, opened 2015-01-18
Deal with "broken-paired" input/output, for better streaming.


ctb:
As part of #732 handling broken paired input, a stopgap measure might be to add an option to ONLY output successfully paired reads when trimming/truncating.

A bit more explanation: right now, filter-abund pays no attention to paired-end reads.  This means that if PE reads are input, pairs or orphaned /1 or /2 reads can be output.  This will break downstream scripts that expect pure paired reads, necessitating various post-processing of upstream script output.  This is tolerable (but ugly) when not doing streaming, but horrible for streaming.  How do we fix this?

I see a few options:

first, we could have scripts that post-process output and discard or sidetrack single-ended reads, and make sure that only paired reads remain.  This is a good stopgap measure, and is in line with all of the streaming work.

second, we could integrate that kind of code into existing scripts, so that e.g. filter-abund provides the option of discarding (or sidetracking) all orphan reads.  Getting the command line options right for this is a bit of a UX nightmare tho ;).

third, we could alter all of the scripts to properly accept "broken paired" input. This is probably the right long-term solution, but every time I think about how to do it in code, my brain locks up.

@camillescott any thoughts on doing the third with your multithreading input work?

camillescott:
I'm not entirely sure what the difference between 1 and 3 here is. In what way would the scripts "accept" broken paired input? The obvious solution to me seems to be to check the pairs as we parse them, and spit orphans out into their own file -- this is kind of the standard way most utilities that respect paired reads seem to handle it (when they do handle it, instead of just core dumping). I see no reason this couldn't be supported by the multithreading work. Am I missing something here @ctb?

mr-c:
Broken paired should be easy; wasn't that the whole point of interleaved paired files in the first place?

ctb:
On Mon, Jan 19, 2015 at 11:44:51AM -0800, Camille Scott wrote:
> I'm not entirely sure what the difference between 1 and 3 here is. In what way would the scripts "accept" broken paired input? The obvious solution to me seems to be to check the pairs as we parse them, and spit orphans out into their own file -- this is kind of the standard way most utilities that respect paired reads seem to handle it (when they do handle it, instead of just core dumping). I see no reason this couldn't be supported by the multithreading work. Am I missing something here @ctb?

Probably just a brainfart on my part.  A simple implementation and support
in your PR (vice semantic versioning) would be wunderbar.

camillescott:
[...]

PR #759, opened 2015-02-07
"broken" paired-read support for a few scripts

### LINKED ISSUE #758: extract-paired-reads doesn't use argparse correctly
see sys.argv usage! oops.

ctb:
Ref #733. Should address #754. Fixes #758, too.


ctb:
OK, preliminary work done; tests still needed:

- [ ] test support for Casava 1.8 format in check_is_pair
- [ ] put in unit and functional tests for trim-low-abund behavior; cross check with filter-abund.

Would appreciate comments from @camillescott and/or @mr-c on the 'broken_paired_read' function - I don't like returning four things, but the only other option is returning a namedtuple. Thoughts?

Function here: 
https://github.com/ged-lab/khmer/pull/759/files#diff-9605187e9bf6b45a2bc03ce9876faf62R49

Note the unavoidable increase in complexity in certain functions :( --
https://github.com/ged-lab/khmer/pull/759/files#diff-d759ac95b2067ed9067857ffc791324dR143

This could be simplified by moving to a callback model as @camillescott and I have discussed.

ctb:
- [ ] Is it mergable?
- [ ] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ctb:
Jenkins, please test this!

ctb:
test this please, Jenkins.

ctb:
test this please, Jenkins.

ctb:
Ready for review. Hey @bocajnotnef @jessicamizzi, how about it?

ctb:
(Would welcome @standage comments)

khmer/utils.py
```diff
@@ -18,9 +18,63 @@ def print_error(msg):
...
+        subpart2 = name2.split('/', 1)[0]
+
+        assert subpart1
+        if subpart1 == subpart2:
+            return True
+
+    return False
+
+
+def broken_paired_reader(screed_iter):
```
standage: I had to scroll down to the script to figure out exactly what this function returns. It would be nice to have an *in situ* description. And @ctb, if you're concerned about returning 4 values, you may consider dropping the `ispair` return value and using the `read2 == None` check instead. Not a huge improvement, but 3 < 4.

ctb:
Updated as per @standage request for docstring. Thinking about the rest ;).

sandbox/trim-low-abund.py
```diff
@@ -130,36 +154,77 @@ def main():
...
-                trim_seq, trim_at = ht.trim_on_abundance(seq, CUTOFF)
-                if trim_at >= K:
-                    trimfp.write(output_single(read, trim_at))
-                    wrote_reads += 1
-                    wrote_bp += trim_at
-                    if trim_at != len(read.sequence):
+                seq1 = read1.sequence.replace('N', 'A')
+                seq2 = read2.sequence.replace('N', 'A')
+
+                med1, _, _ = ht.get_median_count(seq1)
```
bocajnotnef: This is something I've seen in other code but still am not sure what it is--What are the underscores, here?

tests/test_functions.py
```diff
@@ -110,3 +112,60 @@ def test_extract_hashbits_info():
...
+    assert check_is_pair(read1, read2)
+
+
+def test_check_is_pair_4():
+    read1 = FakeFQRead(name='seq/1', accuracy='###', sequence='AAA')
+    read2 = FakeFastaRead(name='seq/2', sequence='AAA')
+
+    try:
+        check_is_pair(read1, read2)
+        assert False                    # expect fail
```
bocajnotnef: Shouldn't this be

```assert check_is_pair(read1, read2)```?

File: sandbox/trim-low-abund.py (hunk 2 above)
Comment (ctb) [inline sandbox/trim-low-abund.py]: "Ignore this variable"

File: tests/test_functions.py (hunk 3 above)
Comment (ctb) [inline tests/test_functions.py]: No, checking that it should raise exception.

Comment (bocajnotnef) [inline tests/test_functions.py]: That the assert raises an exception? Why would it not?

Comment (ctb) [inline tests/test_functions.py]: Ahh, I see now - both the expected exception and the "this failed" exception are assertion errors.  Let me fix that and see if it makes more sense afterwards ;)

(Briefly, check_is_pair is supposed to fail here, so this test should succeed if and only if check_is_pair fails. But it's broken.)

File: khmer/utils.py [hunk 4]
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
...
+    Checks to see if the two sequence records are left and right pairs
+    of each other, respectively.  Returns True or False as appropriate.
+
+    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
+    and 'seq::... 2::...'.
+    """
+    is_fastq = False
+    if hasattr(record1, 'accuracy'):
+        if not hasattr(record2, 'accuracy'):
+            raise ValueError("both records must be same type (FASTA or FASTQ)")
```
mr-c: This logic doesn't work if the first record is FASTQ and the second is FASTA

khmer/utils.py
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
     print >>sys.stderr, msg
 
 
+def check_is_pair(record1, record2):
```
mr-c: Shouldn't this go in screed?

scripts/extract-paired-reads.py
```diff
@@ -22,34 +22,7 @@
...
-        return ">%s\n%s\n>%s\n%s\n" % (read1.name, read1.sequence, read2.name,
-                                       read2.sequence)
-
-
-def output_single(read):
-    if hasattr(read, 'accuracy'):
-        return "@%s\n%s\n+\n%s\n" % (read.name, read.sequence, read.accuracy)
-    else:
-        return ">%s\n%s\n" % (read.name, read.sequence)
+from khmer.utils import (broken_paired_reader, write_record, write_record_pair)
```
mr-c: parens aren't needed on a single line

khmer/utils.py
ctb: On Tue, Feb 10, 2015 at 09:02:12AM -0800, Michael R. Crusoe wrote:
> > @@ -18,9 +18,89 @@ def print_error(msg):
> >      print >>sys.stderr, msg
> >  
> >  
> > +def check_is_pair(record1, record2):
> 
> Shouldn't this go in screed?

Yes, but we live in an imperfect world, and I don't want to wait on screed
releases until that's a systematic and ongoing thing :).  And, of course, we
can move the functionality there more easily once it's been refactored out of
(or into) all the places it needs to be.

"Perfection is the enemy of a good start."

--t
-- 

khmer/utils.py
ctb: Oops, and that's why we do code review - thanks, fixed!

scripts/extract-paired-reads.py
ctb: fixed!

ctb:
Ready for re-review (and merge?)

bocajnotnef:
LGTM. @mr-c, final review?

tests/test_functions.py
```diff
@@ -110,3 +112,71 @@ def test_extract_hashbits_info():
...
+
+
+def test_check_is_pair_3():
+    read1 = FakeFQRead(name='seq 1::', accuracy='###', sequence='AAA')
+    read2 = FakeFQRead(name='seq 2::', accuracy='###', sequence='AAA')
+
+    assert check_is_pair(read1, read2)
+
+
+def test_check_is_pair_4():
```
luizirber: Since we are using nose for tests, what about using the [raises][1] decorator?

``` python
from nose.tools import raises

@raises(ValueError)
def test_check_is_pair_4():
    read1 = FakeFQRead(name='seq/1', accuracy='###', sequence='AAA')
    read2 = FakeFastaRead(name='seq/2', sequence='AAA')

    check_is_pair(read1, read2)
```

Again using nose, this time the [assert_raises][2] context manager:

``` python
from nose.tools import assert_raises

def test_check_is_pair_4():
    read1 = FakeFQRead(name='seq/1', accuracy='###', sequence='AAA')
    read2 = FakeFastaRead(name='seq/2', sequence='AAA')

    with assert_raises(ValueError):
        check_is_pair(read1, read2)
```

Both seem more readable than using the ```try/assert False``` approach.

[1]: http://nose.readthedocs.org/en/latest/testing_tools.html#nose.tools.raises
[2]: http://stackoverflow.com/a/19310416

mr-c: Does that behave the way we expect if the exception is not raised?

luizirber: If the assertion is not raised you get an AssertionError and the test fails:

```python
>>> from nose.tools import assert_raises
>>> with assert_raises(IOError):
...     pass
...
  File "<stdin>", line 2, in <module>
  File "/System/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/unittest/case.py", line 116, in __exit__
    "{0} not raised".format(exc_name))
AssertionError: IOError not raised
```

(from the original code I suspect that was the intention, assert check_is_pair raises an ValueError?)

ctb: I rule this an unnecessary conversation :).  Any changes to function
or syntax that need to be made??

khmer/utils.py
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
...
+    is_fastq = False
+    if hasattr(record1, 'accuracy') or hasattr(record2, 'accuracy'):
+        if not (hasattr(record1, 'accuracy') and hasattr(record2, 'accuracy')):
+            raise ValueError("both records must be same type (FASTA or FASTQ)")
+        is_fastq = True
+
+    name1 = record1.name
+    name2 = record2.name
+
+    if is_fastq and ' ' in name1:                        # handle '@name 1:rst'
```
camillescott: why do we only check casava format if fastq?

khmer/utils.py
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
...
+    of each other, respectively.  Returns True or False as appropriate.
+
+    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
+    and 'seq::... 2::...'.
+    """
+    is_fastq = False
+    if hasattr(record1, 'accuracy') or hasattr(record2, 'accuracy'):
+        if not (hasattr(record1, 'accuracy') and hasattr(record2, 'accuracy')):
+            raise ValueError("both records must be same type (FASTA or FASTQ)")
+        is_fastq = True
```
camillescott: this is erroneous for `khmer.ReadParser`, which *always* has an accuracy attribute, even for FASTA (it just leaves the field as an empty string)

khmer/utils.py
ctb: On Tue, Feb 10, 2015 at 02:09:33PM -0800, Camille Scott wrote:
> > +    Checks to see if the two sequence records are left and right pairs
> > +    of each other, respectively.  Returns True or False as appropriate.
> > +
> > +    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
> > +    and 'seq::... 2::...'.
> > +    """
> > +    is_fastq = False
> > +    if hasattr(record1, 'accuracy') or hasattr(record2, 'accuracy'):
> > +        if not (hasattr(record1, 'accuracy') and hasattr(record2, 'accuracy')):
> > +            raise ValueError("both records must be same type (FASTA or FASTQ)")
> > +        is_fastq = True
> > +
> > +    name1 = record1.name
> > +    name2 = record2.name
> > +
> > +    if is_fastq and ' ' in name1:                        # handle '@name 1:rst'
> 
> why do we only check casava format if fastq?

Because you only get CASAVA format in FASTQ files.

The specific worry here is that people will have overloaded FASTA headers
with personal identifiers, and we will incorrectly sort.  I suppose we
could just barf if it's FASTA and looks like Casava.

I was already on the fence about supporting /1 and /2 for FASTA at all, but
there are still people out there with that format, I bet.

--titus

khmer/utils.py
ctb: On Tue, Feb 10, 2015 at 02:10:33PM -0800, Camille Scott wrote:
> > @@ -18,9 +18,89 @@ def print_error(msg):
> >      print >>sys.stderr, msg
> >  
> >  
> > +def check_is_pair(record1, record2):
> > +    """
> > +    Checks to see if the two sequence records are left and right pairs
> > +    of each other, respectively.  Returns True or False as appropriate.
> > +
> > +    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
> > +    and 'seq::... 2::...'.
> > +    """
> > +    is_fastq = False
> > +    if hasattr(record1, 'accuracy') or hasattr(record2, 'accuracy'):
> > +        if not (hasattr(record1, 'accuracy') and hasattr(record2, 'accuracy')):
> > +            raise ValueError("both records must be same type (FASTA or FASTQ)")
> > +        is_fastq = True
> 
> this is erroneous for `khmer.ReadParser`, which *always* has an accuracy attribute, even for FASTA (it just leaves the field as an empty string)

Yes, I know.  We should fix that one way or the other (personally I prefer
the screed approach, but only at a +0 level).  Neither decision was made
with any forethought.

khmer/utils.py
camillescott: Sure, but plenty of people strip the quality line from their sequence at
some point in their pipeline

On Tue, Feb 10, 2015 at 5:11 PM, C. Titus Brown <notifications@github.com>
wrote:

> In khmer/utils.py
> <https://github.com/ged-lab/khmer/pull/759#discussion_r24455940>:
>
> > +    Checks to see if the two sequence records are left and right pairs
> > +    of each other, respectively.  Returns True or False as appropriate.
> > +
> > +    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
> > +    and 'seq::... 2::...'.
> > +    """
> > +    is_fastq = False
> > +    if hasattr(record1, 'accuracy') or hasattr(record2, 'accuracy'):
> > +        if not (hasattr(record1, 'accuracy') and hasattr(record2, 'accuracy')):
> > +            raise ValueError("both records must be same type (FASTA or FASTQ)")
> > +        is_fastq = True
> > +
> > +    name1 = record1.name
> > +    name2 = record2.name
> > +
> > +    if is_fastq and ' ' in name1:                        # handle '@name 1:rst'
>
> On Tue, Feb 10, 2015 at 02:09:33PM -0800, Camille Scott wrote: > + Checks
> to see if the two sequence records are left and right pairs > + of each
> other, respectively. Returns True or False as appropriate. > + > + Handles
> both Casava formats: seq/1 and seq/2, and 'seq::... 1::...' > + and
> 'seq::... 2::...'. > + """ > + is_fastq = False > + if hasattr(record1,
> 'accuracy') or hasattr(record2, 'accuracy'): > + if not (hasattr(record1,
> 'accuracy') and hasattr(record2, 'accuracy')): > + raise ValueError("both
> records must be same type (FASTA or FASTQ)") > + is_fastq = True > + > +
> name1 = record1.name > + name2 = record2.name > + > + if is_fastq and ' '
> in name1: # handle '@name <https://github.com/name> 1:rst' why do we only
> check casava format if fastq?
> Because you only get CASAVA format in FASTQ files. The specific worry here
> is that people will have overloaded FASTA headers with personal
> identifiers, and we will incorrectly sort. I suppose we could just barf if
> it's FASTA and looks like Casava. I was already on the fence about
> supporting /1 and /2 for FASTA at all, but there are still people out there
> with that format, I bet.
> … <#14b758c6174cf1e8_>
> --titus
>
> —
> <https://github.com/ged-lab/khmer/pull/759/files#r24455940>.
>



-- 
Camille Scott

Department of Computer Science and Engineering
Lab for Genomics, Evolution, and Development
Michigan State University

camille.scott.w@gmail.com

ctb: On Tue, Feb 10, 2015 at 02:15:04PM -0800, Camille Scott wrote:
> > +    Checks to see if the two sequence records are left and right pairs
> > +    of each other, respectively.  Returns True or False as appropriate.
> > +
> > +    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
> > +    and 'seq::... 2::...'.
> > +    """
> > +    is_fastq = False
> > +    if hasattr(record1, 'accuracy') or hasattr(record2, 'accuracy'):
> > +        if not (hasattr(record1, 'accuracy') and hasattr(record2, 'accuracy')):
> > +            raise ValueError("both records must be same type (FASTA or FASTQ)")
> > +        is_fastq = True
> > +
> > +    name1 = record1.name
> > +    name2 = record2.name
> > +
> > +    if is_fastq and ' ' in name1:                        # handle '@name 1:rst'
> 
> Sure, but plenty of people strip the quality line from their sequence at
> some point in their pipeline

I do not believe that to be true; most mappers use it, and all assemblers will
take it even if they don't use it.

But, you know what? let me ask :)

khmer/utils.py
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
...
+            else:                                   # orphan.
+                yield n, False, prev_record, None
+
+        prev_record = record
+
+    # handle the last two records (which might be a pair, or not).
+    if prev_record:
+        if check_is_pair(prev_record, record):
+            yield n, True, prev_record, record  # it's a pair!
+            record = None
```
mr-c: No test coverage for lines 91 + 92

sandbox/trim-low-abund.py
```diff
@@ -36,27 +40,41 @@
 MAX_FALSE_POSITIVE_RATE = 0.8
 
 
-def output_single(read, trim_at=None):
-    name = read.name
-    sequence = read.sequence
-
-    accuracy = None
+def trim_record(read, trim_at):
+    new_read = _screed_record_dict()
```
mr-c: Why not `new_read = copy.deepcopy(reed)` ?

sandbox/trim-low-abund.py
```diff
@@ -36,27 +40,41 @@
...
+    Example::
+
+        trim-low-abund.py -x 5e7 -k 20 -C 2 data/100k-filtered.fa
+    """
+
+    parser = argparse.ArgumentParser(
+        description='Trim low-abundance k-mers using a streaming algorithm.',
+        epilog=textwrap.dedent(epilog))
 
     env_ksize = os.environ.get('KHMER_KSIZE', DEFAULT_K)
```
mr-c:
```
from khmer.khmer_args import build_counting_args
...
parser = build_counting_args(description='Trim low-abundance k-mers using a streaming algorithm.', epilog=textwrap.dedent(epilog))
```

& remove lines 74-91

sandbox/trim-low-abund.py
```diff
@@ -112,7 +136,7 @@ def main():
     tempdir = tempfile.mkdtemp('khmer', 'tmp', args.tempdir)
     print 'created temporary directory %s; use -T to change location' % tempdir
```
mr-c: Output diagnostic message to STDERR

mr-c:
@ctb looking good; just a few nits & cleanups. Almost ready for moving out of the sandbox.

sandbox/trim-low-abund.py
ctb: I like being explicit about what is being transferred here, so -0 on using deepcopy.

That having been said, IMO we need either or both of:

* a public way to create a new screed record;
* a trimming function to do what is used here.

I will add both to issues.

ctb: See #768 and #767.

khmer/utils.py
ctb: See #769

khmer/utils.py
ctb: Turns out I'm wrong - plenty of people think this is a good idea. How weird. willfix.

sandbox/trim-low-abund.py
ctb: fixed.

sandbox/trim-low-abund.py
ctb: fixed 452c36e

ctb:
(Not yet ready for review; one more fix needed.)

khmer/utils.py
ctb: Turns out that second if was unnecessary! Tnx.

mr-c:
Can you make the test data file shorter? It is a bit tedious to scroll through :-)

khmer/utils.py
mr-c: np :-)

ctb:
tests/test-data/test-abund-read-paired.fa? Perhaps, but it's based off of one that worked for filter-low-abund, so would rather not mess with it.

mr-c:
Okay

ctb:
Ready for review.

scripts/trim-low-abund.py
```diff
@@ -13,71 +13,71 @@
```
mr-c: Remove reference to sandbox on line 12

mr-c:
- [ ] add new script to user docs

scripts/trim-low-abund.py
```diff
@@ -13,71 +13,71 @@
...
+    Example::
+
+        trim-low-abund.py -x 5e7 -k 20 -C 2 data/100k-filtered.fa
+    """
+
+    parser = build_counting_args(
+        descr='Trim low-abundance k-mers using a streaming algorithm.',
+        epilog=textwrap.dedent(epilog))
 
     parser.add_argument('--cutoff', '-C', type=int, dest='abund_cutoff',
```
mr-c: I rather like it when the command line option and the argument destination are the same (no `dest='...'`). Makes the code a lot easier to read and think about.

scripts/trim-low-abund.py
```diff
@@ -13,71 +13,71 @@
...
                         help='remove k-mers below this abundance',
                         default=DEFAULT_CUTOFF)
 
     parser.add_argument('--normalize-to', '-Z', type=int, dest='normalize_to',
```
mr-c: This `dest` is redundant

scripts/trim-low-abund.py
```diff
@@ -100,19 +106,20 @@ def main():
     ###
```
mr-c: Instead of quitting we should let the user know that there were duplicate filenames and they were ignored.

scripts/trim-low-abund.py
```diff
@@ -100,19 +106,20 @@ def main():
...
 
     K = args.ksize
-    HT_SIZE = args.min_hashsize
-    N_HT = args.n_hashes
+    HT_SIZE = args.min_tablesize
```
mr-c: The common usage is to just reference `args.whatever` directly.

scripts/trim-low-abund.py
ctb: On Wed, Feb 11, 2015 at 08:06:56AM -0800, Michael R. Crusoe wrote:
> > @@ -100,19 +106,20 @@ def main():
> >      ###
> 
> Instead of quitting we should let the user know that there were duplicate filenames and they were ignored.

I disagree; this could be part of the -f flag, though.

scripts/trim-low-abund.py
```diff
@@ -130,36 +137,77 @@ def main():
```
mr-c: `pass2filename = os.path.join(tempdir, pass2filename)` won't work if the user passed in absolute paths.

mr-c: Nevermind; I missed the call to `os.path.basename`

scripts/trim-low-abund.py
```diff
@@ -130,36 +137,77 @@ def main():
...
 
+        screed_iter = screed.open(filename)
         pass2fp = open(pass2filename, 'w')
         trimfp = open(trimfilename, 'w')
 
         save_pass2 = 0
-        for n, read in enumerate(screed.open(filename)):
+        for n, is_pair, read1, read2 in broken_paired_reader(screed_iter):
             if n % 10000 == 0:
                 print '...', n, filename, save_pass2, read_reads, read_bp, \
```
mr-c: to STDERR, please

scripts/trim-low-abund.py
```diff
@@ -130,36 +137,77 @@ def main():
...
 
-            read_reads += 1
-            read_bp += len(read.sequence)
+            # we want to track paired reads here, to make sure that pairs
+            # are not split between first pass and second pass.
 
-            seq = read.sequence.replace('N', 'A')
-            med, _, _ = ht.get_median_count(seq)
+            if is_pair:
+                read_reads += 2
```
mr-c: s/read_reads/n_reads/g

scripts/trim-low-abund.py
```diff
@@ -130,36 +137,77 @@ def main():
...
+        screed_iter = screed.open(filename)
         pass2fp = open(pass2filename, 'w')
         trimfp = open(trimfilename, 'w')
 
         save_pass2 = 0
-        for n, read in enumerate(screed.open(filename)):
+        for n, is_pair, read1, read2 in broken_paired_reader(screed_iter):
             if n % 10000 == 0:
                 print '...', n, filename, save_pass2, read_reads, read_bp, \
                     wrote_reads, wrote_bp
```
mr-c: s/wrote/written/g

ctb: On Wed, Feb 11, 2015 at 08:20:36AM -0800, Michael R. Crusoe wrote:
> s/wrote/written/g

"a simple past tense of write."
http://dictionary.reference.com/browse/wrote

scripts/trim-low-abund.py
```diff
@@ -100,19 +106,20 @@ def main():
...
-    N_HT = args.n_hashes
+    HT_SIZE = args.min_tablesize
+    N_HT = args.n_tables
 
     CUTOFF = args.abund_cutoff
     NORMALIZE_LIMIT = args.normalize_to
 
-    print 'making hashtable'
+    print >>sys.stderr, 'making counting table'
     ht = khmer.new_counting_hash(K, HT_SIZE, N_HT)
```
mr-c: s/ht/countingtable/g

scripts/trim-low-abund.py
mr-c: You can keep the 'wrote' on line 269; that's the correct place for the verb. For the variable a noun phrase is advised therefore use `written_reads` and `written_bp`

ctb: On Wed, Feb 11, 2015 at 08:30:30AM -0800, Michael R. Crusoe wrote:
> >          pass2fp = open(pass2filename, 'w')
> >          trimfp = open(trimfilename, 'w')
> >  
> >          save_pass2 = 0
> > -        for n, read in enumerate(screed.open(filename)):
> > +        for n, is_pair, read1, read2 in broken_paired_reader(screed_iter):
> >              if n % 10000 == 0:
> >                  print '...', n, filename, save_pass2, read_reads, read_bp, \
> >                      wrote_reads, wrote_bp
> 
> You can keep the 'wrote' on line 269; that's the correct place for the verb. For the variable a noun phrase is advised therefore use `written_reads` and `written_bp`

Uh, this is starting to get perilously close to bikeshedding.  I'll change
it for now, but when spelling is correct and intent is clear, I am -1 on
changing it.

scripts/trim-low-abund.py
ctb: On Wed, Feb 11, 2015 at 08:07:50AM -0800, Michael R. Crusoe wrote:
> > @@ -100,19 +106,20 @@ def main():
> >      ###
> >  
> >      K = args.ksize
> > -    HT_SIZE = args.min_hashsize
> > -    N_HT = args.n_hashes
> > +    HT_SIZE = args.min_tablesize
> 
> The common usage is to just reference `args.whatever` directly.

Keeping K, as it is used elsewhere and it is quite clear :)

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,295 @@
...
+    Example::
+
+        trim-low-abund.py -x 5e7 -k 20 -C 2 data/100k-filtered.fa
+    """
+
+    parser = build_counting_args(
+        descr='Trim low-abundance k-mers using a streaming algorithm.',
+        epilog=textwrap.dedent(epilog))
+
+    parser.add_argument('--cutoff', '-C', type=int, dest='cutoff',
```
mr-c: *sigh* The point is that you can leave the 'dest' off and it'll match the long form of the argument :-)

scripts/extract-paired-reads.py
```diff
@@ -98,45 +71,21 @@ def main():
...
     n_pe = 0
     n_se = 0
 
-    record = None
-    index = 0
-    for index, record in enumerate(screed.open(sys.argv[1])):
+    screed_iter = screed.open(args.infile)
+    for index, is_pair, read1, read2 in broken_paired_reader(screed_iter):
         if index % 100000 == 0 and index > 0:
             print '...', index
```
mr-c: Remaining non-data print statements need redirecting to STDERR

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,295 @@
...
+                        written_reads += 1
+                        written_bp += trim_at
+
+                        if trim_at != len(read1.sequence):
+                            trimmed_reads += 1
+
+        pass2fp.close()
+        trimfp.close()
+
+        print '%s: kept aside %d of %d from first pass, in %s' % \
```
mr-c: to STDERR

standage:
Just a quick question about GED Lab vernacular: it's pretty clear what @ctb means when he says he's +1 or -1 for a suggested change, but what is meant by -0? Not against but negatively inclined?

scripts/trim-low-abund.py
ctb: On Wed, Feb 11, 2015 at 08:46:29AM -0800, Michael R. Crusoe wrote:
> > +                else:                       # trim!!
> > +                    _, trim_at = ct.trim_on_abundance(seq, CUTOFF)
> > +                    if trim_at >= K:
> > +                        new_read = trim_record(read1, trim_at)
> > +                        write_record(new_read, trimfp)
> > +
> > +                        written_reads += 1
> > +                        written_bp += trim_at
> > +
> > +                        if trim_at != len(read1.sequence):
> > +                            trimmed_reads += 1
> > +
> > +        pass2fp.close()
> > +        trimfp.close()
> > +
> > +        print '%s: kept aside %d of %d from first pass, in %s' % \
> 
> to STDERR

No, this is useful information that I want logged/saved; it's not available
anywhere else, and it's not purely diagnostic.

mr-c:
@standage Bingo. Recent archeological digging by myself has revealed that this practice has its roots in the Python developer list.

scripts/trim-low-abund.py
mr-c: Okey dokey

ctb:
test this please

ctb:
ready for review.

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,295 @@
...
+    print >>sys.stderr, \
+        'fp rate estimated to be {fpr:1.3f}'.format(fpr=fp_rate)
+
+    if fp_rate > MAX_FALSE_POSITIVE_RATE:
+        print >> sys.stderr, "**"
+        print >> sys.stderr, ("** ERROR: the k-mer counting table is too small"
+                              " for this data set. Increase tablesize/# "
+                              "tables.")
+        print >> sys.stderr, "**"
+        print >> sys.stderr, "** Do not use these results!!"
```
mr-c: No test coverage for this branch

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,295 @@
...
+                    _, trim_at = ct.trim_on_abundance(seq, CUTOFF)
+                    if trim_at >= K:
+                        new_read = trim_record(read1, trim_at)
+                        write_record(new_read, trimfp)
+
+                        written_reads += 1
+                        written_bp += trim_at
+
+                        if trim_at != len(read1.sequence):
+                            trimmed_reads += 1
```
mr-c: Odd, no test coverage for this line

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,295 @@
...
+                    ct.consume(seq1)
+                    ct.consume(seq2)
+                    write_record_pair(read1, read2, pass2fp)
+                    save_pass2 += 2
+                else:
+                    _, trim_at1 = ct.trim_on_abundance(seq1, CUTOFF)
+                    _, trim_at2 = ct.trim_on_abundance(seq2, CUTOFF)
+
+                    if trim_at1 >= K:
+                        read1 = trim_record(read1, trim_at1)
```
mr-c: No test coverage for this line

mr-c:
Looks good modulo the gaps in test coverage

ctb:
Ready for review. Note I took care of the big test file problem, too :)

khmer/utils.py
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
...
 
+def check_is_pair(record1, record2):
+    """
+    Checks to see if the two sequence records are left and right pairs
+    of each other, respectively.  Returns True or False as appropriate.
+
+    Handles both Casava formats: seq/1 and seq/2, and 'seq::... 1::...'
+    and 'seq::... 2::...'.
+    """
+    is_fastq = False
```
camillescott: now that the casava change has been made, this variable is extraneous

khmer/utils.py
```diff
@@ -18,9 +18,89 @@ def print_error(msg):
...
+    is incremented by 2 for a pair of reads.
+    """
+    record = None
+    prev_record = None
+
+    # handle the majority of the stream.
+    for n, record in enumerate(screed_iter):
+        if prev_record:
+            if check_is_pair(prev_record, record):
+                yield n, True, prev_record, record  # it's a pair!
```
camillescott: Why include the `n` as a return value at all? It's more common to let the user decide if they want to enumerate.

ctb: On Wed, Feb 11, 2015 at 11:03:45AM -0800, Camille Scott wrote:
> > +    Usage::
> > +
> > +       for n, is_pair, read1, read2 in broken_paired_reader(...):
> > +          ...
> > +
> > +    Note that 'n' is the number of records read from the input stream, so
> > +    is incremented by 2 for a pair of reads.
> > +    """
> > +    record = None
> > +    prev_record = None
> > +
> > +    # handle the majority of the stream.
> > +    for n, record in enumerate(screed_iter):
> > +        if prev_record:
> > +            if check_is_pair(prev_record, record):
> > +                yield n, True, prev_record, record  # it's a pair!
> 
> Why include the `n` as a return value at all? It's more common to let the user decide if they want to enumerate.

enumerate will not differentiate between a pair or a singleton.
-- 

camillescott: which is covered by the 2nd value in the tuple already

On Wed, Feb 11, 2015 at 2:04 PM, C. Titus Brown <notifications@github.com>
wrote:

> In khmer/utils.py
> <https://github.com/ged-lab/khmer/pull/759#discussion_r24522692>:
>
> > +    Usage::
> > +
> > +       for n, is_pair, read1, read2 in broken_paired_reader(...):
> > +          ...
> > +
> > +    Note that 'n' is the number of records read from the input stream, so
> > +    is incremented by 2 for a pair of reads.
> > +    """
> > +    record = None
> > +    prev_record = None
> > +
> > +    # handle the majority of the stream.
> > +    for n, record in enumerate(screed_iter):
> > +        if prev_record:
> > +            if check_is_pair(prev_record, record):
> > +                yield n, True, prev_record, record  # it's a pair!
>
> On Wed, Feb 11, 2015 at 11:03:45AM -0800, Camille Scott wrote: > + Usage::
> > + > + for n, is_pair, read1, read2 in broken_paired_reader(...): > + ...
> > + > + Note that 'n' is the number of records read from the input stream,
> so > + is incremented by 2 for a pair of reads. > + """ > + record = None >
> + prev_record = None > + > + # handle the majority of the stream. > + for
> n, record in enumerate(screed_iter): > + if prev_record: > + if
> check_is_pair(prev_record, record): > + yield n, True, prev_record, record
> # it's a pair! Why include the `n` as a return value at all? It's more
> common to let the user decide if they want to enumerate.
> enumerate will not differentiate between a pair or a singleton. -- C.
> Titus Brown, ctbrown@ucdavis.edu
>
> —
> <https://github.com/ged-lab/khmer/pull/759/files#r24522692>.
>



-- 
Camille Scott

Department of Computer Science and Engineering
Lab for Genomics, Evolution, and Development
Michigan State University

camille.scott.w@gmail.com

camillescott: ah. scratch that.

On Wed, Feb 11, 2015 at 2:06 PM, Camille Scott <camille.scott.w@gmail.com>
wrote:

> which is covered by the 2nd value in the tuple already
>
> On Wed, Feb 11, 2015 at 2:04 PM, C. Titus Brown <notifications@github.com>
> wrote:
>
>> In khmer/utils.py
>> <https://github.com/ged-lab/khmer/pull/759#discussion_r24522692>:
>>
>> > +    Usage::
>> > +
>> > +       for n, is_pair, read1, read2 in broken_paired_reader(...):
>> > +          ...
>> > +
>> > +    Note that 'n' is the number of records read from the input stream, so
>> > +    is incremented by 2 for a pair of reads.
>> > +    """
>> > +    record = None
>> > +    prev_record = None
>> > +
>> > +    # handle the majority of the stream.
>> > +    for n, record in enumerate(screed_iter):
>> > +        if prev_record:
>> > +            if check_is_pair(prev_record, record):
>> > +                yield n, True, prev_record, record  # it's a pair!
>>
>> On Wed, Feb 11, 2015 at 11:03:45AM -0800, Camille Scott wrote: > +
>> Usage:: > + > + for n, is_pair, read1, read2 in broken_paired_reader(...):
>> > + ... > + > + Note that 'n' is the number of records read from the input
>> stream, so > + is incremented by 2 for a pair of reads. > + """ > + record
>> = None > + prev_record = None > + > + # handle the majority of the stream.
>> > + for n, record in enumerate(screed_iter): > + if prev_record: > + if
>> check_is_pair(prev_record, record): > + yield n, True, prev_record, record
>> # it's a pair! Why include the `n` as a return value at all? It's more
>> common to let the user decide if they want to enumerate.
>> enumerate will not differentiate between a pair or a singleton. -- C.
>> Titus Brown, ctbrown@ucdavis.edu
>>
>> —
>> Reply to this email directly or view it on GitHub
>> <https://github.com/ged-lab/khmer/pull/759/files#r24522692>.
>>
>
>
>
> --
> Camille Scott
>
> Department of Computer Science and Engineering
> Lab for Genomics, Evolution, and Development
> Michigan State University
>
> camille.scott.w@gmail.com
>



-- 
Camille Scott

Department of Computer Science and Engineering
Lab for Genomics, Evolution, and Development
Michigan State University

camille.scott.w@gmail.com

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,295 @@
...
+    skipped_bp = 0
+    for orig_filename, pass2filename, trimfilename in pass2list:
+        print 'second pass: looking at sequences kept aside in %s' % \
+              pass2filename
+
+        # note that for this second pass, we don't care about paired
+        # reads - they will be output in the same order they're read in,
+        # so pairs will stay together if not orphaned.  This is in contrast
+        # to the first loop.
+
```
camillescott: I'm not sure I understand the logic here

ctb: Disable the pairing if stmt and run tests - see what breaks :). Sugg on clarifying comment welcome.

--

ctb:
test this please

ctb:
test this please, Jenkins

ctb:
please test this, jenkins

standage:
*Just frickin test it Jenkins!!!!!*

Sorry, it's just funny when you get email notifications for thread comments but not the additional commits/pushes between the comments.

ctb:
Jenkins, test this please

ctb:
ok, ready for review. I don't think there are any other obstacles?

ctb:
Hold off on merge, please; identified a bug.

ctb:
Found a bug where trim-low-abund failed on sequences shorter than the k-mer size & found the simplest solution involved updating broken_paired_reader to deal; see 7a28aa4.

ready for review!

camillescott:
Is this really the appropriate place to take care of this bug though? It seems to me that a reader should just read reads, and let any service using it decide on other filtering steps.

ctb:
While I'm not a big fan of the broken_paired_reader function implementation, this gives a cleaner implementation in the actual script than anything I could come up with (the main script is actually unchanged).  The best alternative I've seen is in normalize-by-median:

https://github.com/ged-lab/khmer/blob/master/scripts/normalize-by-median.py#L89

which is using a general for loop to deal with the situation.  (Which also points out that this is a general pattern/problem which could use a general solution.)

It doesn't seem like a particularly bad signature (note also the force_single parameter is pretty useful, and we can add a paired-reads-only option later, too).  A way to generalize it might be to have multiple iterator functions that are composable (e.g. pairs_only(broken_paired_reader(...))), but that's overkill until we have a couple more use cases.

camillescott:
Fair enough. It isn't enough added complexity to be a big deal, so I'll stay -0 on it.

ctb:
- [ ] add khmer citation header to output; add citation info.

khmer/utils.py
```diff
@@ -18,9 +18,101 @@ def print_error(msg):
...
+        if not force_single and record:
+            assert not check_is_pair(prev_record, record)
+
+        yield n, False, prev_record, None
+
+    if record:                     # guaranteed to be orphan
+        # ignore short reads
+        if not min_length or len(record.sequence) >= min_length:
+            n += 1
+            yield n, False, record, None
```
mr-c: No test coverage for 108-110

khmer/utils.py
```diff
@@ -18,9 +18,101 @@ def print_error(msg):
...
+
+        prev_record = record
+        record = None
+
+    # handle the last two records (which cannot be pairs)
+    if prev_record:
+        # the only way into this if statement is if 'prev_record' and
+        # 'record' are both singletons.
+        if not force_single and record:
+            assert not check_is_pair(prev_record, record)
```
mr-c: No test coverage

scripts/extract-paired-reads.py
```diff
@@ -98,45 +71,21 @@ def main():
...
-                n_se += 1
-
-        last_name = name
-        last_record = record
-
-    if last_record:
-        if is_pair(last_name, name):
-            paired_fp.write(output_pair(last_record, record))
-            name, record = None, None
+            print >>sys.stderr, '...', index
```
mr-c: Shucks, now we don't have coverage for this line :-)

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+            "Error: Cannot input the same filename multiple times."
+        sys.exit(1)
+
+    ###
+
+    report_on_config(args)
+    check_valid_file_exists(args.input_filenames)
+    check_space(args.input_filenames, args.force)
+    if args.savetable:
+        check_space_for_hashtable(
```
mr-c: no test coverage

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+            args.n_tables * args.min_tablesize, args.force)
+
+    K = args.ksize
+
+    CUTOFF = args.cutoff
+    NORMALIZE_LIMIT = args.normalize_to
+
+    if args.loadtable:
+        print >>sys.stderr, 'loading k-mer counting table from', args.loadtable
+        ct = khmer.load_counting_hash(args.loadtable)
```
mr-c: lines 127-128 lack test coverage

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+        pass2fp = open(pass2filename, 'w')
+        trimfp = open(trimfilename, 'w')
+
+        save_pass2 = 0
+
+        iter = broken_paired_reader(screed_iter, min_length=K,
+                                    force_single=args.ignore_pairs)
+        for n, is_pair, read1, read2 in iter:
+            if n % 10000 == 0:
+                print >>sys.stderr, '...', n, filename, save_pass2, \
```
mr-c: No test coverage

scripts/trim-low-abund.py
mr-c: This is now covered, thanks

scripts/trim-low-abund.py
mr-c: This is now covered, thanks

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+                              "tables.")
+        print >> sys.stderr, "**"
+        print >> sys.stderr, "** Do not use these results!!"
+        sys.exit(1)
+
+    print 'output in *.abundtrim'
+
+    if args.savetable:
+        print >>sys.stderr, "Saving k-mer counting table to", args.savetable
+        ct.save(args.savetable)
```
mr-c: lines 322-323 lack test coverage

ChangeLog
```diff
@@ -1,3 +1,22 @@
+2015-02-16  Titus Brown  <titus@idyll.org>
+
+   * khmer/utils.py: added 'check_is_pair', 'broken_paired_reader', and
+   'write_record_pair' functions.
+   * khmer/khmer_args.py: added streaming reference for future algorithms
+   citation.
+   * tests/test_functions.py: added unit tests for 'check_is_pair' and
+   'broken_paired_reader'.
+   * scripts/trim-low-abund.py: upgraded to track pairs properly; added
+   proper get_parser information; moved to scripts/.
```
mr-c: This entry should end with 'moved from sandbox/' or begin with the original filename

khmer/utils.py
```diff
@@ -18,9 +18,101 @@ def print_error(msg):
...
+
+        yield n, False, prev_record, None
```
mr-c: n isn't defined in this scope; should be initialized before line 81

khmer/utils.py
```diff
@@ -33,4 +125,15 @@ def write_record(record, fp):
             '>{name}\n{seq}\n'.format(name=record.name,
                                       seq=record.sequence))
 
+
+def write_record_pair(read1, read2, fp):
```
mr-c: Pylint reminds us that 'fp' is both a) too short and b) not self documenting. 'file_object' or 'fileobj' would be clearer here.

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+import sys
+import screed
+import os
+import khmer
+import tempfile
+import shutil
+import textwrap
+
+from screed.screedRecord import _screed_record_dict
+from khmer.khmer_args import (build_counting_args, info, add_loadhash_args,
```
mr-c: `W0611: (unused-import), : Unused import add_loadhash_args`

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+                        help='base cutoff on this median k-mer abundance',
+                        default=DEFAULT_NORMALIZE_LIMIT)
+
+    parser.add_argument('--variable-coverage', '-V', action='store_true',
+                        default=False,
+                        help='Only trim low-abundance k-mers from sequences '
+                        'that have high coverage.')
+
+    parser.add_argument('-s', '--savetable', metavar="tablefile", default='')
+    parser.add_argument('-l', '--loadtable', metavar="tablefile", default='')
```
mr-c: Any particular reason you're not using `add_loadhash_args` for `--loadtable`?

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+
+        pass2list.append((filename, pass2filename, trimfilename))
+
+        screed_iter = screed.open(filename)
+        pass2fp = open(pass2filename, 'w')
+        trimfp = open(trimfilename, 'w')
+
+        save_pass2 = 0
+
+        iter = broken_paired_reader(screed_iter, min_length=K,
```
mr-c: `W0622: (redefined-builtin), main: Redefining built-in 'iter'`

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+                        written_bp += trim_at
+
+                        if trim_at != len(read1.sequence):
+                            trimmed_reads += 1
+
+        pass2fp.close()
+        trimfp.close()
+
+        print '%s: kept aside %d of %d from first pass, in %s' % \
+              (filename, save_pass2, n, filename)
```
mr-c: Is `n` defined in this context?

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+        save_pass2_total += save_pass2
+
+    # ### SECOND PASS. ###
+
+    skipped_n = 0
+    skipped_bp = 0
+    for orig_filename, pass2filename, trimfilename in pass2list:
```
mr-c: W0612: (unused-variable), main: Unused variable 'orig_filename'

scripts/trim-low-abund.py
```diff
@@ -0,0 +1,326 @@
...
+    print 'wrote %d reads, %d bp' % (written_reads, written_bp,)
+    print 'looked at %d reads twice (%.2f passes)' % (save_pass2_total,
+                                                      n_passes)
+    print 'removed %d reads and trimmed %d reads (%.2f%%)' % \
+        (n_reads - written_reads, trimmed_reads, percent_reads_trimmed)
+    print 'trimmed or removed %.2f%% of bases (%d total)' % \
+        ((1 - (written_bp / float(n_bp))) * 100.0, n_bp - written_bp)
+
+    if args.variable_coverage:
+        percent_reads_hicov = 100.0 * float(n_reads - skipped_n) / n_reads
```
mr-c: W0612: (unused-variable), main: Unused variable 'percent_reads_hicov'

scripts/extract-paired-reads.py
ctb: I think we're going to have to live with it; this is prohibitive to test.

mr-c: Shouldn't we be able to stream fake reads in to trigger this?

ctb: Maybe?

But a few thoughts --

this isn't diagnostic output;
if there's one thing that's well tested "in the wild", it's this;
the acceptance tests certainly cover this line.

So I personally think this is simply not worth the extra effort and code.

mr-c: Okay. Once the acceptance tests are running automatically can collect
coverage information from them and fold that into stats.

ctb:
Fixed all of the issues, AFAICT. What should I be doing to look at code coverage?

ctb:
ok ready for review.

mr-c:
`make clean diff-cover`

ctb:
Ready for review; all lacking code coverage has been addressed, and it's now mergeable.

mr-c:
Fweh, finally! :-)

ctb:
heh

Issue #718, opened 2015-01-07
Error needs cleaning up for input file not existing


mr-c:
```
|| This is the script 'extract-paired-reads.py' in khmer.
|| You are running khmer version 1.3-rc3
|| You are also using screed version 0.7
||
|| If you use this script in a publication, please cite EACH of the following:
||
||   * MR Crusoe et al., 2014. http://dx.doi.org/10.6084/m9.figshare.979190
||
|| Please see http://khmer.readthedocs.org/en/latest/citations.html for details.

Traceback (most recent call last):
  File "/usr/local/bin/extract-paired-reads.py", line 5, in <module>
    pkg_resources.run_script('khmer==1.3-rc3', 'extract-paired-reads.py')
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 528, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "build/bdist.linux-x86_64/egg/pkg_resources.py", line 1394, in run_script
    execfile(script_filename, namespace, namespace)
  File "/usr/local/lib/python2.7/dist-packages/khmer-1.3_rc3-py2.7-linux-x86_64.egg/EGG-INFO/scripts/extract-paired-reads.py", line 154, in <module>
    main()
  File "/usr/local/lib/python2.7/dist-packages/khmer-1.3_rc3-py2.7-linux-x86_64.egg/EGG-INFO/scripts/extract-paired-reads.py", line 86, in main
    check_file_status(args.infile, args.force)
  File "/usr/local/lib/python2.7/dist-packages/khmer-1.3_rc3-py2.7-linux-x86_64.egg/khmer/file.py", line 25, in check_file_status
    mode = os.stat(file_path).st_mode
OSError: [Errno 2] No such file or directory: '*.pe.*.abundfilt'
```

mr-c:
@jessicamizzi How is this going?

mr-c:
Fixed by #772

PR #772, opened 2015-02-13
Start of issue 718, file not existing clean up

jessicamizzi:
Fix for #718.

### DIFF
--- a/khmer/kfile.py
+++ b/khmer/kfile.py
@@ -16,7 +16,7 @@ from stat import S_ISBLK, S_ISFIFO
 
 def check_file_status(file_path, force):
     """Check the status of the file; if the file is empty or doesn't exist
-    AND if the file is NOT a fifo/bloc k/named pipe then a warning is printed
+    AND if the file is NOT a fifo/block/named pipe then a warning is printed
     and sys.exit(1) is called
     """
 
--- a/khmer/kfile.py
+++ b/khmer/kfile.py
@@ -16,13 +16,22 @@ from stat import S_ISBLK, S_ISFIFO
 
 def check_file_status(file_path, force):
     """Check the status of the file; if the file is empty or doesn't exist
-    AND if the file is NOT a fifo/block/named pipe then a warning is printed
+    AND if the file is NOT a fifo/bloc k/named pipe then a warning is printed
     and sys.exit(1) is called
     """
 
     if file_path is '-':
         return
-    mode = os.stat(file_path).st_mode
+    try:
+        mode = os.stat(file_path).st_mode
+
+    except OSError as error:
+        print >>sys.stderr, "ERROR: Input file %s does not exist; exiting" % \
+                            file_path
+
+        if not force:
+            sys.exit(1)
+
     # block devices will be nonzero
     if S_ISBLK(mode) or S_ISFIFO(mode):
         return
--- a/khmer/kfile.py
+++ b/khmer/kfile.py
@@ -25,7 +25,7 @@ def check_file_status(file_path, force):
     try:
         mode = os.stat(file_path).st_mode
 
-    except OSError as error:
+    except OSError:
         print >>sys.stderr, "ERROR: Input file %s does not exist; exiting" % \
                             file_path
 
--- a/khmer/kfile.py
+++ b/khmer/kfile.py
@@ -19,17 +19,21 @@ def check_file_status(file_path, force):
     AND if the file is NOT a fifo/block/named pipe then a warning is printed
     and sys.exit(1) is called
     """
+    mode = None    
 
     if file_path is '-':
         return
     try:
         mode = os.stat(file_path).st_mode
     except OSError:
-        print >>sys.stderr, "ERROR: Input file %s does not exist; exiting" % \
+        print >>sys.stderr, "ERROR: Input file %s does not exist" % \
                             file_path
-
+        
         if not force:
+            print >>sys.stderr, "Exiting"
             sys.exit(1)
+        else:
+            return
 
     # block devices will be nonzero
     if S_ISBLK(mode) or S_ISFIFO(mode):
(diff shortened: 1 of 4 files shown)


jessicamizzi:
- [ ] Is it mergable?
- [ ] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

jessicamizzi:
Can 2 people please review this code? @b-wyss @bocajnotnef @brtaylor92

bocajnotnef:
Looks good to me minus the non-mergable status.

b-wyss:
Looks good to me as well, unless you want to change existance to existence? Totally up to you, but the rest looks okay!

khmer/kfile.py
```diff
@@ -22,7 +22,16 @@ def check_file_status(file_path, force):
 
     if file_path is '-':
         return
-    mode = os.stat(file_path).st_mode
+    try:
+        mode = os.stat(file_path).st_mode
+
```
ctb: unnecessary empty line ;)

tests/test_scripts.py
```diff
@@ -1926,3 +1926,15 @@ def test_readstats_empty():
     assert status == 0
 
     assert expected_output in out
+
+
+def test_existance_failure():
```
ctb: You didn't run a spellchecker :)

tests/test_scripts.py
```diff
@@ -1926,3 +1926,15 @@ def test_readstats_empty():
...
+    expected_output = 'ERROR: Input file'
+
+    args = ['thisfiledoesnotexistatall']
```
ctb: Please use:

utils.get_temp_filename('thisfiledoesnotexistatall')

tests/test_scripts.py
```diff
@@ -1926,3 +1926,15 @@ def test_readstats_empty():
...
+
+    status, out, err = utils.runscript(
```
ctb: This test should specifically test the changes you made to kfile.py - if you remove those changes, will this test still succeed? It doesn't look like it to me.

So, I think you should actually write two tests -- one test for a script like extract-paired-reads, that tests the error message and exit status.  And one test in test_script_arguments.py that runs the code in kfile directly, and verifies that behavior.  Note that sys.exit raises SystemExit, so you can look for that in the test.

(Let me know if you'd like me to expand on all of this!)

jessicamizzi: Will the test in test_scripts.py work to as the first test you described?  If I'm understanding you, there only needs to be another test in test_script_arguments.py to test the code in kfile.  I would like some expansion on the test_script_arguments.py test if possible.

Also - how do I a run a spell checker? Thanks!

ctb: for running a spell checker - what editor are you using? it'll depend on that. or you can just copy/paste the code into Word...

yes, I misread the test in test_scripts - it's good now!

the test in test_script_arguments should be of the specific kfile function, as opposed to running a script that runs the kfile function.  so, it should call kfile.check_file_status specifically, and verify that it raises a SystemExit exception when the file doesn't exist; and maybe another test to check that if you specify force, it doesn't exit; etc.  Basically you should test all of the if/else paths in the code.  Does that make more sense?

khmer/kfile.py
```diff
@@ -22,7 +22,16 @@ def check_file_status(file_path, force):
...
+    except OSError as error:
```
mr-c: `W0612: (unused-variable), check_file_status: Unused variable 'error'` -> no need for `as error` here

tests/test_functions.py
```diff
@@ -114,6 +114,21 @@ def test_extract_hashbits_info():
             print >>sys.stderr, '...failed to remove {fn}'.format(fn)
 
 
+def check_file_status_kfile():
+    try:
+        kfile.check_file_status('thisfiledoesnotexistatall')
```
ctb: put an 'assert False' after this, because it *must* raise an OSError to succeed.

tests/test_functions.py
```diff
@@ -114,6 +114,21 @@ def test_extract_hashbits_info():
             print >>sys.stderr, '...failed to remove {fn}'.format(fn)
 
 
+def check_file_status_kfile():
+    try:
```
ctb: generate a guaranteed-not-to-exist filename with utils.get_temp_filename('thisfiledoesnotexist')

tests/test_functions.py
```diff
@@ -114,6 +115,21 @@ def test_extract_hashbits_info():
...
+def test_check_file_status_kfile():
+    try:
+        utils.get_temp_filename('thisfiledoesnotexistatall')
+    except OSError as e:
+        print >>sys.stder, '...failed to remove {fn}'.format(fn)
+
+
+def test_check_file_status_kfile_force():
+    try:
+        khmer.kfile('thisfiledoesnotexistatall', force)
```
mr-c: `khmer.kfile` isn't the name of a function

tests/test_functions.py
```diff
@@ -114,6 +115,24 @@ def test_extract_hashbits_info():
             print >>sys.stderr, '...failed to remove {fn}'.format(fn)
 
 
+def test_check_file_status_kfile():
+    fn = utils.get_temp_filename('thisfiledoesnotexist')
+    try:
+        check_file_status(fn, False)
+        assert True
```
jessicamizzi: @ctb could you remind me of the idea behind the if/if not statement to be put here you mentioned today?

ctb: You _could_ rewrite it as:

```
check_file_status_exited = False
try:
    check_file_status(fn, False)
except SystemExit:
    check_file_status_exited = True

assert check_file_status_exited
```

which makes the logic clear, at least.

jessicamizzi:
@mr-c has the compatibility problem with pep8/autopep8 been fixed?

mr-c:
Yes, sorry. Update to the latest from the master branch and you should be
good to go.

mr-c:
Thanks!

Issue #728, opened 2015-01-16
_khmermodule.cc cleanup


mr-c:
- [x] `tp_name` field of PyTypeObjects needs to include module name; dot separated; Example: `khmer.Read`
- [ ] rename all Python facing mentions of 'hash' to 'table' (and update Python code)
- [x] remove `_common_init_Type` and refactor accordingly
- [x] PyTypeObjects should be name `khmer_${TYPENAME}_Type`; 
- [x] all PyTypeObjects should be implemented as new style Python classes
- [x] Type methods should use their type Object in the method signature instead of `PyObject`. Cast the methods to `PyCFunction`s in the `PyMethodDef`.
- [x] Type deallocators should use their type Object in the method signature instead of `PyObject`. Cast the method in `tp_dealloc` to `(destructor)`

Types to review
- [x] `Read_Type`
- [x] `ReadParser_Type`
- [x] `ReadPairIterator_Type`
- [x] `khmer_KSubsetPartitionType`
- [x] `khmer_KCountingHashType`
- [x] `khmer_KHashbitsType`
- [x] `khmer_KLabelHashType`
- [x] `khmer_ReadAlignerType`

luizirber:
Working on it: #785

ctb:
More or less done (#785, #1044) and the naming scheme stuff is punted to #872.

PR #785, opened 2015-02-19
Some _khmermodule.cc refactoring

### LINKED ISSUE #38: Cleanup C++ Extension Module
While debugging and figuring some things out about creating a Python type's class attributes from C, I realized that much of our extension module code is making life way harder on itself than need be and does not conform to best practices (as given by the CPython API documentation). In some cases, our existing practices result in objects not exposing all of the standard attributes that one would expect from a Python object. In other cases, we are simply writing code for things that various standard CPython API functions, in conjunction with certain type object slots, which have been available since at least Python 2.2, handle automatically.

While I am not advocating a massive housecleaning effort all at once, we should be looking to simplify and otherwise improve the various pieces of code as we touch them. In my humble estimation, we may be able to shave 500 or more lines from the extension module code and actually gain functionality in the process....

(This should be weighed against issue #13. Yes, I still have a SWIG itch I want to scratch.)

luizirber:
Update extension to pave the way for the future :wink:
PyLong instead of PyInt, Type initialization, PyBytes instead of PyString.
Replace common initialization with explicit type structs, and all types conform to the CPython checklist.

This fixes #38 and #728. The only item missing from #728 checklist is changing references from 'hash' to 'table' in all methods. All classes are also conforming to the CPython checklist.

TODO:
  - [x] `ReadParser_Type`: expose constants in tp_dict

Remaining questions:
- Do we define `tp_new` for 'read-only' types, like `ReadPairIterator_Type` and `Read_Type`? They can't be constructed on Python land at the moment, and reflect internal components from C++ structures. (`khmer_KSubsetPartitionType` is also missing `tp_new`, but it is probably appropriate to deal with it in #802.)

- Missing documentation for methods: see #832

- `khmer_KHashbitsType`: Some methods deal with SubsetPartition, but doesn't touch the Hashbits directly. See #833

### DIFF
--- a/khmer/_khmermodule.cc
+++ b/khmer/_khmermodule.cc
@@ -3709,29 +3709,25 @@ typedef struct {
     //PyObject_HEAD
     khmer_KHashbits_Object khashbits;
     LabelHash * labelhash;
-} khmer_KLabelHashObject;
+} khmer_KLabelHash_Object;
 
-static void khmer_labelhash_dealloc(PyObject *);
-static int khmer_labelhash_init(khmer_KLabelHashObject * self, PyObject *args,
+static int khmer_labelhash_init(khmer_KLabelHash_Object * self, PyObject *args,
                                 PyObject *kwds);
 static PyObject * khmer_labelhash_new(PyTypeObject * type, PyObject *args,
                                       PyObject *kwds);
 
-#define is_labelhash_obj(v)  (Py_TYPE(v) == &khmer_KLabelHashType)
+#define is_labelhash_obj(v)  (Py_TYPE(v) == &khmer_KLabelHash_Type)
 
 //
 // khmer_labelhash_dealloc -- clean up a labelhash object.
 //
 
-static void khmer_labelhash_dealloc(PyObject* obj)
+static void khmer_labelhash_dealloc(khmer_KLabelHash_Object * obj)
 {
-    khmer_KLabelHashObject * self = (khmer_KLabelHashObject *) obj;
-
-    delete self->labelhash;
-    self->labelhash = NULL;
+    delete obj->labelhash;
+    obj->labelhash = NULL;
 
     Py_TYPE(obj)->tp_free((PyObject*)obj);
-    //PyObject_Del((PyObject *) obj);
 }
 
 // a little weird; we don't actually want to call Hashbits' new method. Rather, we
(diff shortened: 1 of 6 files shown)


khmer/_khmermodule.cc
```diff
@@ -24,7 +24,31 @@
...
+//
+// Python 2.6 compatibility macros: PyCapsule and PyCObject
+// https://docs.python.org/2/howto/cporting.html#cobject-replaced-with-capsule
+//
+
+#include "capsulethunk.h"
+
+//
+// Python 2/3 compatibility: PyBytes and PyString
+// https://docs.python.org/2/howto/cporting.html#str-unicode-unification
```
mr-c: put these comments in the header files themselves

khmer/_khmermodule.cc
```diff
@@ -24,7 +24,31 @@
...
+//
+
+#include "bytesobject.h"
+
 using namespace khmer;
+using namespace read_parsers;
```
mr-c: this line doesn't do anything

khmer/_khmermodule.cc
```diff
@@ -189,70 +211,59 @@ namespace khmer
...
-    return KHMER_READ_STRING_GETTER( self, annotations );
+    return PyBytes_FromString(obj->read->annotations.c_str()) ;
 }
 
 
 // TODO? Implement setters.
 
 
-static PyGetSetDef _Read_accessors [ ] = {
+static PyGetSetDef khmer_Read_accessors [ ] = {
```
mr-c: Maybe the other way?

luizirber:
jenkins, test this please

luizirber:
# Code review checklist

- [ ] Is it mergable?
- [ ] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

# CPython checklist

  entries
     the `ob_size` field).
     `tp_str` & `tp_repr`
  in `init_khmer()`

luizirber:
Ready for review @mr-c @ctb @camillescott.

luizirber:
jenkins, test this please

mr-c:
Jenkins, test this please

mr-c:
Looks great so far!

mr-c:
Fwoosh!

Issue #615, opened 2014-09-22
loading of counting table with bigcounts needs test coverage


mr-c:
See: https://github.com/ged-lab/khmer/blob/master/lib/counting.cc#L633

mr-c:
Current test coverage: http://ci.ged.msu.edu/job/khmer-master/label=linux/185/cobertura/lib/counting_cc/

Look at lines 693-709
Test this by loading a gzip compressed counting table that was made with bigcount=True

mr-c:
Fixed by #809

PR #809, opened 2015-02-20
loading of counting table with bigcounts (part 2)

mr-c:


SherineAwad:
@CTB  error is not in test_counting but it is related I think. 
The error is:
 Now, there is this error [unpack requires a string argument of length 1] in extractcountinghashinfo n_tables, = unpack('B', countinghash.read(3)) in __init.py >>> and counthash is used in gzfilereader

tests/test_scripts.py
```diff
@@ -2681,6 +2681,18 @@ def test_trim_low_abund_trimtest_savetable():
                 'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
 
 
+<<<<<<< HEAD
```
ctb: these are merge artifacts that should be removed...

ctb:
The problem seems to be that neither extract_hashbits_info nor extract_countinghash_info support gzipped files.   In cases where filename.endswith('.gz'), would need to use gzip.open instead of open.

tests/test_scripts.py
SherineAwad: yes, at the end I will remove everything.

SherineAwad:
@CTB so,  test of compressed counting will fail unless these functions support .gz. I will work on that.

tests/test_scripts.py
ctb: On Mon, Feb 23, 2015 at 04:03:14PM -0800, Sherine wrote:
> > @@ -2681,6 +2681,18 @@ def test_trim_low_abund_trimtest_savetable():
> >                  'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
> >  
> >  
> > +<<<<<<< HEAD
> 
> yes, at the end I will remove everything. 

This will prevent your code from running at all, as it is a Python syntax
error.

--t
-- 

ctb:
On Mon, Feb 23, 2015 at 04:05:50PM -0800, Sherine wrote:
> @CTB so,  test of compressed counting will fail unless these functions support .gz. I will work on that. 

yep! please ask for help as needed!

SherineAwad:
Jenkins test this please.

SherineAwad:
Ready For Review

 - [x] Is it mergable?
- [x] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ChangeLog
```diff
@@ -1,3 +1,6 @@
+2015-02-24  Sherine Awad  <sherine.awad@gmail.com>
+   * tests/test_scripts.py: Test loading of counting table with bigcounts
```
mr-c: blank line between lines 1 & 2 please

mr-c:
Please remove the file `count.ct.gz` from the root of the khmer tree.

SherineAwad:
@mr-c  Points fixed and PR is mergable.

tests/test_scripts.py
```diff
@@ -1,4 +1,5 @@
 #
+
```
mr-c: Please always check your commits for extra random lines like this and remove them.

tests/test_scripts.py
```diff
@@ -2693,6 +2694,17 @@ def test_trim_low_abund_trimtest_savetable():
...
 
+def test_counting():
+    infile = utils.get_temp_filename('count.ct.gz')
+    indir = os.path.dirname(infile)
+    shutil.copyfile(utils.get_test_data('count.ct.gz'), infile)
+    assert os.path.exists(infile), infile
+    try:
+        test_ct = khmer.load_counting_hash(infile)
+    except IOError:
+        raise Exception("Can't load a counting table with bigcount set")
```
mr-c: Exceptions should not be raised in any of the tests; this prevents other tests from running even if `stop = FALSE` is set.

Try `assert 0, "Can't load a counting table with bigcount set"` instead.

SherineAwad:
Not yet.

tests/test_scripts.py
```diff
@@ -2693,6 +2693,17 @@ def test_trim_low_abund_trimtest_savetable():
                 'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
 
 
+def test_counting():
```
ctb: @SherineAwad, a few comments --

first, the test name should be something like 'test_counting_load_bigcount' - test_counting is too generic!

second, please create the 'count.ct.gz' file in this test, from a sequence file that is in test-data, rather than committing the count.ct.gz file to the repository directly for the test to use.  There are a couple reasons for this, but the main one is that if we update the counting table file format, this test will break :).

third, note that there is no guarantee in the test currently that bigcounts are loaded correctly - what you're really testing is that the count.ct.gz file can be loaded by load_counting_hash without a complaint.  That's great to test, but it's not testing anything about bigcounts particularly.  To test bigcount save/load, you should create a saved table from a FQ file that has k-mers with counts over 255; then load that in, and test to see if the k-mer count > 255 is retained.  A particularly easy way to do this would be to create a k-mer table with a ksize of 2 from one of our larger files in test-data, and then find a 2-mer that occurs a lot; some of the *abund* in tests/test-data/ might be a good place to look.

fourth, the advice about the assert function was, I think, incorrect.  The 'load' function raises an IOError if it can't load the file; this is a fine exception to break the test if load fails.  You can eliminate the try/except/assert that's there right now and the test will still fail properly if .load(...) fails.

This might be useful background reading: http://khmer.readthedocs.org/en/latest/dev/a-quick-guide-to-testing.html

mr-c: @ctb If the counting file format changes then shouldn't this test fail?

ctb: On Wed, Feb 25, 2015 at 08:38:25AM -0800, Michael R. Crusoe wrote:
> > @@ -2693,6 +2693,17 @@ def test_trim_low_abund_trimtest_savetable():
> >                  'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
> >  
> >  
> > +def test_counting():
> 
> @ctb If the counting file format changes then shouldn't this test fail?

that's a test with a different purpose, I think.

--titus

mr-c: I thought that was the purpose of this test, to ensure that gzipped
counting tables (still) work, regardless of gzip library or other code
changes.

mr-c: make that "bigcount gzipped counting tables" :-)

ctb:
Note, in a meeting I encouraged @SherineAwad to keep her test of abundance_distribution along with a more pointed test of the bigcount save/load behavior.

SherineAwad:
@ctb ready to take a look please but something wrong with testing with Jenkins??
Jenkins test this please !

mr-c:
@SherineAwad A test failed
```
tests.test_scripts.test_counting_load_gzipped_bigcount ... ERROR

======================================================================
ERROR: tests.test_scripts.test_counting_load_gzipped_bigcount
----------------------------------------------------------------------
  File "/var/lib/jenkins/workspace/khmer-pullrequest/label/linux/.env/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/var/lib/jenkins/workspace/khmer-pullrequest/label/linux/tests/test_scripts.py", line 2791, in test_counting_load_gzipped_bigcount
    newct.load('test_ct.gz')
IOError: Cannot open k-mer count file: test_ct.gz
```

ChangeLog
```diff
@@ -1,3 +1,7 @@
+2015-03-09  Sherine Awad  <sherine.awad@gmail.com>
+
+   * tests/test_scripts.py: Test loading of compressed counting table with bigcounts and test abundance with bigcounts
```
mr-c: FYI: this line is too long. Please wrap the text at 80 characters.

tests/test_scripts.py
```diff
@@ -2777,6 +2775,52 @@ def test_trim_low_abund_trimtest_savetable():
...
+    ct.set_use_bigcount(True)
+    for i in range(500):
+        ct.count('ATATATATAT')
+    ct.save(infile)
+    outfile = utils.get_temp_filename('test_ct.gz')
+    fin = open(infile, 'rb').read()
+    fout = gzip.open(outfile, 'wb')
+    fout.write(fin)
+    newct = khmer.new_counting_hash(10, 1e7, 4)
+    newct.load('test_ct.gz')
```
mr-c: You need to reference the temporary filename stored in the `outfile` variable on line 2786. Otherwise it will try to load `test_ct.gz` from wherever the current directory happens to be.

tests/test_scripts.py
```diff
@@ -2777,6 +2775,52 @@ def test_trim_low_abund_trimtest_savetable():
...
+    flag = False
+    count = newct.get('ATATATATAT')
+    if count == 500:
+        flag = True
+    assert flag
```
mr-c: Lines 2792-96 can be two simple statements.

tests/test_scripts.py
```diff
@@ -2777,6 +2775,52 @@ def test_trim_low_abund_trimtest_savetable():
...
+    infile = utils.get_temp_filename('test_ct')
+    ct = khmer.new_counting_hash(10, 1e7, 4)
+    ct.set_use_bigcount(True)
+    for i in range(500):
+        ct.count('ATATATATAT')
+    ct.save(infile)
+    outfile = utils.get_temp_filename('test_ct.gz')
+    fin = open(infile, 'rb').read()
+    fout = gzip.open(outfile, 'wb')
+    fout.write(fin)
```
mr-c: Why are we making a new gzip-ed counting file when this PR already adds one to the repository as `tests/test-data/count.ct.gz`?

SherineAwad: We will create the gzipped file on the fly as @ctb suggested. So, I will take that file out.

mr-c: I disagree. It needs to be a static file. @ctb I thought we discussed this?

SherineAwad:
jenkins test this please !

mr-c:
Jenkins, test this please

SherineAwad:
@ctb @mr-c Not sure now whether you wanna the counting table static or not. 
Other than that, ready for review !

tests/test_scripts.py
```diff
@@ -2777,6 +2775,49 @@ def test_trim_low_abund_trimtest_savetable():
                 'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
 
 
+def test_counting_load_gzipped_bigcount():
+    import gzip
```
mr-c: Imports go at the top of the file.

mr-c: But I still disagree with this approach.

tests/test_scripts.py
```diff
@@ -2777,6 +2776,47 @@ def test_trim_low_abund_trimtest_savetable():
...
 
+def test_counting_load_gzipped_bigcount():
+    infile = utils.get_temp_filename('test_ct')
+    ct = khmer.new_counting_hash(10, 1e7, 4)
+    ct.set_use_bigcount(True)
+    for i in range(500):
+        ct.count('ATATATATAT')
+    ct.save(infile)
+    outfile = infile + '.gz'
+    with open(infile, 'rb') as f_in:
```
ctb: You shouldn't use writelines for binary data -- use

data = open(infile, 'wb').read()
f_out = gzip.open(outfile, 'wb')
f_out.write(data)

to gzip things.

tests/test_scripts.py
```diff
@@ -2777,6 +2776,47 @@ def test_trim_low_abund_trimtest_savetable():
                 'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
 
 
+def test_counting_load_gzipped_bigcount():
+    infile = utils.get_temp_filename('test_ct')
+    ct = khmer.new_counting_hash(10, 1e7, 4)
```
ctb: you can decrease this a lot, which is probably for the best -- maybe use 1e5 instead of 1e7?

ctb:
I agree with the approach; this isn't about testing versions of files, but about testing loading of gzip files.

tests/test_scripts.py
SherineAwad: When I use these same line, it works with me and fails with Jenkins. I will figure this out.

SherineAwad:
Jenkins test this please !

SherineAwad:
@ctb  Fixed

tests/test_scripts.py
```diff
@@ -2777,6 +2776,49 @@ def test_trim_low_abund_trimtest_savetable():
...
+    ct.set_use_bigcount(True)
+    for i in range(500):
+        ct.count('ATATATATAT')
+    ct.save(infile)
+    outfile = infile + '.gz'
+    data = open(infile, 'rb').read()
+    f_out = gzip.open(outfile, 'wb')
+    f_out.write(data)
+    f_out.close()
+    newct = khmer.new_counting_hash(10, 1e5, 4)
```
ctb: This is unnecessary - please use 'newct = khmer.load_counting_hash(outfile)'

tests/test_scripts.py
```diff
@@ -2777,6 +2776,49 @@ def test_trim_low_abund_trimtest_savetable():
...
+    newct.load(outfile)
+    count = newct.get('ATATATATAT')
+    assert count == 500
+
+
+def test_abundance_distribution_gzipped_bigcount():
```
ctb: I think this test should be changed to look only at bigcount behavior test in counting_hash.abundance_distribution, which is untested elsewhere; your other test function does a fine job of testing bigcount load in gzip files.

So, remove all of the hashtable save/load stuff, and just make it about bigcount testing.

Also, this function should be moved to test_counting_hash.py, because it doesn't test a script.

tests/test_scripts.py
```diff
@@ -1404,15 +1407,11 @@ def test_abundance_dist_nobigcount():
     infile = utils.get_temp_filename('test.fa')
     outfile = utils.get_temp_filename('test.dist')
     in_dir = os.path.dirname(infile)
-
```
ctb: Why are there empty lines removed?

tests/test_scripts.py
```diff
@@ -2777,6 +2776,49 @@ def test_trim_low_abund_trimtest_savetable():
                 'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAGAGACAGCA'
 
 
+def test_counting_load_gzipped_bigcount():
```
ctb: This function should be moved into test_counting_hash.py, as should the above function, test_abundance_dist_nobigcount.

SherineAwad:
Jenkins test this please.
 
@ctb Please take a look and note that  am ignoring some warnings from make diff_pylint_report ? Regardung unnecessary variables sound needed ?!

mr-c:
Jenkins, retest this please

tests/test_counting_hash.py
```diff
@@ -35,6 +38,10 @@ def teardown():
     utils.cleanup()
 
 
+def scriptpath(script):
+    return script
```
mr-c: What is the point of this function?

SherineAwad: Without it,  I get this error:  " global name 'scriptpath' is not defined", for this line:
script = scriptpath('load-into-counting.py')
 your recommendation is welcomed?

SherineAwad: and if I used script = 'load-into-counting.py'  works locally only !

mr-c: Correct. If you move code from one file to another and you reference a
function from the original file then you should continue to reference that
function from the new location instead of writing a "do nothing" stub.

SherineAwad:
@mr-c please check this.

tests/test_counting_hash.py
```diff
@@ -1071,3 +1076,41 @@ def test_find_all_tags_list_error():
         assert False, "a ValueError should be raised for incorrect k-mer size"
     except ValueError:
         pass
+
+
+def test_abundance_distribution_gzipped_bigcount():
+    infile = utils.get_temp_filename('test.fa')
+    in_dir = os.path.dirname(infile)
```
mr-c: `1083: W0612: (unused-variable), test_abundance_distribution_gzipped_bigcount: Unused variable 'in_dir'`

tests/test_counting_hash.py
```diff
@@ -1071,3 +1076,41 @@ def test_find_all_tags_list_error():
...
+    flag = False
+    for _, i in enumerate(abundances):
+        print _, i
+        if _ > 255 and i > 0:
+            flag = True
+            break
+    assert flag
+
+
+def test_counting_load_gzipped_bigcount():
```
mr-c: The title of this test implies that a gzipped counting table will be loaded but it doesn't do so. What is your goal here?

mr-c:
@SherineAwad Otherwise this looks good to me.

tests/test_counting_hash.py
SherineAwad: I changed the name to test_coounting_load_bigcount

mr-c: Thanks!

ctb:
@SherineAwad should we take another look at this?

SherineAwad:
I think now I addressed all the comments. So, yes please.

SherineAwad:
@ctb please wait. Jenkins test failed again.

SherineAwad:
@ctb ready now for review.

ChangeLog
```diff
@@ -1,3 +1,12 @@
+2015-04-04  Sherine Awad  <sherine.awad@gmail.com>
+
+   * tests/test_scripts.py: Test loading of compressed counting table
+   with bigcounts,and test abundance with bigcounts
+
+2015-02-24  Sherine Awad  <sherine.awad@gmail.com>
+
+   * tests/test_scripts.py: Test loading of gzipped counting table with bigcounts and test abundance with bigcounts
```
mr-c: I think you meant to delete the second entry here.

SherineAwad: Yes, sorry. I fixed it.

tests/test_counting_hash.py
```diff
@@ -1071,3 +1076,41 @@ def test_find_all_tags_list_error():
...
         pass
+
+
+def test_abundance_distribution_gzipped_bigcount():
+    infile = utils.get_temp_filename('test.fa')
+    in_dir = os.path.dirname(infile)
+    shutil.copyfile(utils.get_test_data('test-abund-read-2.fa'), infile)
+    outfile = utils.get_temp_filename('test_ct.gz')
+    script = scriptpath('load-into-counting.py')
+    args = ['-x', str(1e7), '-N', str(2), '-k', str(2)]
```
ctb: these next few lines could be simplified to:

args = [ '-x', ..., htfile, infile])

without using the args.extend.

ctb:
Looks good; please address the two things that @mr-c and I brought up, and if you can add some comments in the longer function to break it up I would appreciate it. You'll also need to make it mergable.

We can look it over together today, if you get to it before our meeting.

SherineAwad:
@ctb  @mr-c  args.extend is removed. It is mergable.

SherineAwad:
@ctb Missed @mr-c  point and comments. So wait please.

ctb:
ping

SherineAwad:
Jenkins test this please !
@CTB ready for review pls.

ctb:
@SherineAwad, there is a PEP 8 violation when you run 'make pep8'.

     tests/test_counting_hash.py:1099:72: W291 trailing whitespace

Please fix this by removing the space at the end of that line.

ctb:
Also, diff_pylint_report found the following issues:

    tests/test_counting_hash.py (86.7%):
       21: W0403: (relative-import), : Relative import 'test_scripts', should be 'tests.test_scripts'
       1081: C0103: (invalid-name), test_abundance_distribution_gzipped_bigcount: Invalid function name         
               "test_abundance_distribution_gzipped_bigcount"
       1099: C0303: (trailing-whitespace), : Trailing whitespace
        1109: C0303: (trailing-whitespace), : Trailing whitespace
        1113: C0103: (invalid-name), test_counting_load_bigcount: Invalid variable name "ct"
        1115: W0612: (unused-variable), test_counting_load_bigcount: Unused variable 'i'

Could you please meet with @mr-c this upcoming week to resolve these?

Apart from the formatting issues, this is a good addition and should be merged IMO. thx!

SherineAwad:
Jenkins test this please.

SherineAwad:
@ctb @mr-c  Issues fixed.

ctb:
Hi @SherineAwad this looks good to me; could you update the checklist, please?

Also, I'd like to try rebasing it before merging, to compress down all the commits; mind if I do so, or do you want to do it together?

SherineAwad:
@ctb No, I don't mind at all :)  
- [x] Is it mergable?
- [x] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ctb:
Sorry, I meant, could you check the various boxes in the checklist?? No need to make a new copy, it's already above...

SherineAwad:
@ctb ok. updated.

ctb:
Done! Thanks, @SherineAwad

ctb:
Addresses #615.

Issue #817, opened 2015-02-21
Weird handling of Fastq data


standage:
I grabbed these two interleaved read pairs from one of my data sets to do some screed testing and see if I can begin to wrap my head around #606. I can't remember if these are from a HiSeq or a GAII, but I think it's the former.

```
@HWI-ST412:261:d15khacxx:8:1101:3149:2157 1:N:0:ATCACG
GTACACTATAGTGAGATTCCTTATACCTCGCAATATGATCTCCTGTCGTTACGTCCATTGCTCTTAACAAAGAACACTCATAGAGTAGAATCGAAAAAAC
+
@@CDAFFAHFACC?DBHIAIIICGC?AHFEEHGGIIGG>CEDHBDHDGBGGBB8B<FGHG4@GEFDA@.7CC3?)7@?CDECCAA663;;;(;=??B?>5
@HWI-ST412:261:d15khacxx:8:1101:3149:2157 2:N:0:ATCACG
TCGTTTTCTTGTAGATTAGATTTGACATAAATGTCAATTTTTTTGTATCTACCGTCACTTGAATACTCAGTTGCAATATAAT
+
;+=D7D?DHGHFDCAE@BCCACHHGFBHICC<EHE?BFBDHGGIIADDB*?D8@B@FG>FCE;@D:A>EECEFF3?BD@(.6
@HWI-ST412:261:d15khacxx:8:1101:3426:2112 1:N:0:ATCACG
CTAAACACAGCAAATTTGTCAGAGATTTAATCCGTGAAGTAACTGGTCATGCTCCTTATGAAAAACGTGCTATGGAATTGCTCAAAGTATCACAGGATA
+
1=DDFFAFFFHFIGBHGGDHIH>GC9?FEHGFGIIIIEEBGGBFHGGEIGHICHBFGCHEGIIIIGGCEGEHCHA?@@B?@D@C@>@>CCA>(;5;AA>
@HWI-ST412:261:d15khacxx:8:1101:3426:2112 2:N:0:ATCACG
GCCCAGCCTTCTTTTCAAAAACTTCAATGCACGCTTATCCTTTGATACTTTGAGCAATTCCATAGCACGTTTTTCATAAGGAGCATGACCAGTTACTTCA
+
[...]

PR #789, opened 2015-02-20
Fix/#786 column headers to histogram output of abundance-dist.py

drtamermansour:
Fix #786 : adding column headers to the histogram output of abundance-dist.py **as an option**.
The column headers of this output file are already documented in doc/user/scripts.txt.

Issue #763, opened 2015-02-09
Minor fixes to split-paired-reads.py


ctb:
Once #762 is merged, there are still a few updates to split-paired-reads that could usefully be done.

First, to support https://github.com/ged-lab/khmer/issues/732#issuecomment-71903726 we need to default to stdin.  That can be done by adding ``nargs='?', default='/dev/stdin'`` to line 60, although we also need to require -1 and -2 options in that case.

Second, we should switch to using the broken_paired_reader in khmer.utils that #759 will introduce.

Third, once we can read broken-paired reads, we should allow users to override the requirement that reads be perfectly interleaved. (By default, we should require perfect interleaving, though, because it's the main use case.)

ctb:
Forgot to say - switching to broken_paired_reader is important because we will then support Casava 1.8 pair checking in split-paired-reads (ref #23).

ctb:
Switched to broken_paired_reader & addressed third point in #818.  First point (supporting streaming) still needs to be done.

mr-c:
@ctb This should support streaming as is with `/dev/stdin` being used as the input file name.

mr-c:
@ctb I think this issue is finished; can you confirm?

ctb:
No, not yet --

- [ ] 'infile' should default to /dev/stdin, nargs='?', and in that case -1 and -2 are required

mr-c:
@ctb I don't think we can change that sort of default without bumping the version number?

ctb:
Ok, but most of it can be done wo changing default; does specifying - work?

--
Titus Brown, ctbrown@ucdavis.edu

mr-c:
@ctb '-' will once @anotherthomas is done with #960 (which adds this behaviour to screed)

ctb:
This is now straightforward to do, but I continue to note that when /dev/stdin is specified the default filenames are broken ;).

mr-c:
- [ ] don't break if '/dev/stdin' is the input filename

ctb:
@SherineAwad are you working on this?

SherineAwad:
@ctb just started.

ctb:
ok - please create a new PR, thanks!

ctb:
Fixed in #763 and elsewhere.

mr-c:
@ctb #763 is fixed in #763? :-)

ctb:
;)

PR #818, opened 2015-02-21
Explore a fix for broken casava format handling in a number of scripts

ctb:
See #817, among others.


ctb:
Confronting #817 head-on, we have a few different problems:

- split-paired-reads.py would drop reads that it couldn't interpret (oops)
- screed by default splits up the record name into 'name' and 'description', which breaks Casava 1.8 names
- this splitting leads to silly results -- e.g. see https://github.com/ged-lab/khmer/blob/fix/casava_formats/tests/test_scripts.py#L2316

It looks to me like the right thing to do is make sure that parse_description is always set to false
in screed iterators. This will necessitate changing about a dozen things in scripts/. @mr-c thoughts on this wrt semantic versioning?

cc @standage

ctb:
Addresses #763

standage:
This fixes the splitting/interleaving issues with Casava 1.8 AFAICS.

mr-c:
The changes might not be fun but if the tests pass without modifications then we are violating the semantic versioning.

ctb:
Sorry, clarify: yes, fix, and to heck with semver?

mr-c:
mea culpa; ----> Then we are _not_ violating semantic versioning if the
tests continue to pass unmodified.

mr-c:
To summarize my understanding: khmer has always mangled sequence record names: either by dropping descriptions or appending different pairing notation on the end.

This PR aims to bring khmer in line with expected practice: both old and new style pair naming are accepted everywhere; descriptions will no longer be dropped.

With respect to semantic versioning we have three options:

A) We declare the old behavior a horrible bug and widely publicize the correction and thus we don't increment the major version number.

or 

B) We accept that the old behavior was present for such a long time and likely required workarounds that will break upon its fixing. Therefore we increment the major version to indicate this, thus following our commitment to semantic (and not 'marketing') versioning.

or 

C) we hid the corrected behavior behind a switch and wait for the 2.0 release to make it the default behavior

As Titus privately pointed out to me, seeing what changes will be required to update the protocols (if anything) will inform the wisdom of option B or C.

ctb:
While additional tests are needed, and it's not yet ready for a full CR, it's mostly done, I think? Thoughts welcome.  I'll finish sanding down the corners tomorrow or later today, as time permits.

ctb:
- [ ] add tests for filter-abund
- [ ] add tests for filter-abund-single
- [ ] resolve issues with count-median now allowing spaces in output
- [ ] add tests for do-partition
- [ ] add tests for annotate-partition and extract-partition
- [ ] add tests for filter-stoptags
- [ ] add tests for normalize-by-median
- [ ] add tests for extract-long-sequences
- [ ] add tests for fastq-to-fasta
- [ ] add tests for sample-reads-randomly

OK, so it's a slightly bigger issue than I thought :).

ctb:
- [ ] file issue to update file formats to proper CSV, with headers, for khmer 2.0.

ctb:
- [ ] file an issue regarding --csv for anything that outputs tabular format; also see #789

ctb:
For posterity:

* fixed bug in utils.check_is_pair, introduced last week; https://github.com/ged-lab/khmer/blob/fix/casava_formats/khmer/utils.py#L36
* fixed old bug in split-paired-reads.py where reads with unknown pairing format would just get dropped: https://github.com/ged-lab/khmer/blob/fix/casava_formats/scripts/split-paired-reads.py#L143
* fixed bug in tests for extract-long-reads where output file was improperly named and not actually tested anyway: https://github.com/ged-lab/khmer/blob/fix/casava_formats/tests/test_scripts.py#L1995

So that was awesome... :)

ctb:
For me to do:

- [ ] do diff cover and diff pylint analysis.
- [ ] look at split-paired-reads.py and interleave-reads.py critically, to see if more tests are needed.

@mr-c and others, a code review would be welcome of points not related to the above checklist; basically, any obvious flaws in the approach, or concerns that I've missed? TIA!

mr-c:
Jenkins, test this please

ctb:
- [ ] Is it mergable?
- [ ] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ctb:
Ready for review, w00t.

ctb:
Additional issues/bugs found & fixed:

* removed no longer needed test-data files;
* interleave-reads.py used to happily interleave the same file twice; now it errors out.

scripts/count-median.py
```diff
@@ -34,8 +36,18 @@ def get_parser():
     based on the k-mer counts in the given k-mer counting table.  Can be used
     to estimate expression levels (mRNAseq) or coverage (genomic/metagenomic).
 
-    The output file contains sequence id, median, average, stddev, and seq
-    length.
+    The output file contains sequence id, median, average, stddev, and
+    seq length; fields are separated by spaces.  For khmer 1.x,
+    sequence names are split at the first space, which means that some
```
mr-c: 'For khmer 1.x count-median.py will split sequence names at the first space which means that some ..'

scripts/count-median.py
```diff
@@ -34,8 +36,18 @@ def get_parser():
...
+    sequence formats (e.g. paired FASTQ in Casava 1.8 format) will yield
+    uninformative names.  Use :option:`--csv` to fix this behavior.
+
+    :option:`--csv` will output a comma-separated value file with column
+    headers, and the entire sequence name will be kept.
```
mr-c: 45-46 is a bit redundant; especially when viewed in context: http://athyra.ged.msu.edu/~mcrusoe/kdocs/user/scripts.html#count-median-py

ctb: I think the redundancy is not harmful, in this case, and may be helpful to people who only skim docs.

tests/test_scripts.py
```diff
@@ -258,10 +258,33 @@ def test_filter_abund_3_fq_retained():
...
+
+
+# make sure that FASTQ names are properly parsed, both formats.
+
+
+def test_filter_abund_4_fq_casava_18():
+    infile = utils.get_temp_filename('test.fq')
+    in_dir = os.path.dirname(infile)
+
+    shutil.copyfile(utils.get_test_data('test-abund-read-2.paired2.fq'),
```
mr-c: It would be nifty if Casava 1.8 formatted test sequence files had 'c18' in their name.

mr-c:
LGTM!

ctb:
w00t!

Issue #619, opened 2014-09-25
generate unix manual ("man") pages from the docs


mr-c:

Echelon9:
Claiming this one to work on.

PR #838, opened 2015-02-24
Initial support for man page as a make target

Echelon9:
Fixes Issue #619.

**Work In Progress**
_Not yet ready for review_
- [ ] Is it mergable?
- [x] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,6 +91,15 @@ build/sphinx/latex/khmer.pdf: $(SOURCES) doc/conf.py $(wildcard doc/*.txt)
 	@echo ''
 	@echo '--> pdf in build/sphinx/latex/khmer.pdf'
 
+## man         : render documentation in Linux man
+man: build/sphinx/man/khmer.1
+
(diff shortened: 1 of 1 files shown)


mr-c:
Thank you @Echelon9 for starting this.

The main reason we had an issue to create manual pages is due to Debian's requirement. However, for the khmer Debian package we introduced a `khmer` frontend program to avoid naming conflicts and therefore we only had to write a single (though incomplete) manual page: http://anonscm.debian.org/cgit/debian-med/khmer.git/tree/debian/khmer.1.ronn

Issues:
- [ ] Where / how to ship per script manual page(s) via PyPI
- [ ] How to make those manual pages show up using `man khmer normalize-by-median` and the like with the Debian package

mr-c:
For 'Where / how to ship per script manual page(s) via PyPI' I think we don't ship them via pip/PyPI (See http://stackoverflow.com/questions/20003329/install-python-package-man-pages-with-pip )

standage:
Move to #1788.

PR #843, opened 2015-02-24
Fix the append behavior in normalize-by-median.py (#745)

### LINKED ISSUE #745: normalize-by-median.py --force default should be 'true'
In messing around with `normalize-by-median.py`, I realized that if I killed a run while it was running, then issued a new command using the same output filename, that the product of the new command was *appended* to the results of the first command. __I don't think this is ever when the user wants__ - a set of reads will be duplicated in the normalized output file.. I think that the output should always be overwritten `--force TRUE` with or without a warning that the file is being overwritten.. 

    grep HWI-ST1097:199:D22V1ACXX:1:1101:1154:2129 khmer.normalized.C50.fastq
    >HWI-ST1097:199:D22V1ACXX:1:1101:1154:2129/1
    >HWI-ST1097:199:D22V1ACXX:1:1101:1154:2129/2
    >HWI-ST1097:199:D22V1ACXX:1:1101:1154:2129/1
    >HWI-ST1097:199:D22V1ACXX:1:1101:1154:2129/2

My 2cents

drtamermansour:
Fix #745:  in normalize-by-median.py, change to the default behavior to overwrite the sequences output file. Also add a new argument --append to append new reads to the output file.
I also added a test for the --append option in test_scripts.py


drtamermansour:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

drtamermansour:
jenkins, please test this

drtamermansour:
ready for review

tests/test_scripts.py
```diff
@@ -514,6 +514,21 @@ def test_normalize_by_median():
     assert len(seqs) == 1, seqs
     assert seqs[0].startswith('GGTTGACGGGGCTCAGGGGG'), seqs
 
+    # Testing for the apppend option
```
ctb: please make a separate test function instead -- more granular tests are generally better. So, 

* copy the test function twice
* rename both copies
* remove the outfile/append tests from the original
* have one copy do the append test
* have the other copy do the not-append test

thanks!

scripts/normalize-by-median.py
```diff
@@ -233,13 +236,20 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
     discarded = 0
     input_filename = None
 
-    for index, input_filename in enumerate(args.input_filenames):
-        if args.single_output_filename != '':
-            output_name = args.single_output_filename
+    auto_output = False
```
ctb: The addition of auto_output is a bit confusing and, I think, unnecessary.  What about removing it, and (inside the 'for' loop) putting 'if not args.single_output_filename'? This seems like it would be simpler to me.

drtamermansour:
@ctb please review

ctb:
LGTM on a skim - might ask @mr-c to review/merge since I can't get to it 'til much later today or maybe tomorrow.

scripts/normalize-by-median.py
```diff
@@ -192,6 +192,9 @@ def get_parser():
                         dest='single_output_filename',
                         default='', help='only output a single'
                         ' file with the specified filename')
+    parser.add_argument('--append', default=False, action='store_true',
+                        help='append reads to the outputfile. '
+                        'Only with -o specified')
```
mr-c: I would word the help text this way: "appends reads instead of replacing the output file (only if -o/--out is specified)'

scripts/normalize-by-median.py
```diff
@@ -233,13 +236,17 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
+        if args.append:
             outfp = open(args.single_output_filename, 'a')
         else:
+            outfp = open(args.single_output_filename, 'w+')
+
+    for index, input_filename in enumerate(args.input_filenames):
+        if not args.single_output_filename:
             output_name = os.path.basename(input_filename) + '.keep'
-            outfp = open(output_name, 'w')
+            outfp = open(output_name, 'w+')
```
mr-c: Line 248 overrides line 240
Line 249 overrides line 239 & 244.

Are there tests with multiple input files using `--out` with and without `--append`? There is likely a bug here. Regardless the output_name and outfp should be defined only once.

ctb:
Nice work; thanks!

PR #844, opened 2015-02-24
upgrade write-trimmomatic

mr-c:
Please review @ctb @SherineAwad

### DIFF
--- a/sandbox/write-trimmomatic.py
+++ b/sandbox/write-trimmomatic.py
@@ -8,13 +8,19 @@
 import glob
 import sys
 
-filelist = glob.glob('*R1*.fastq.gz')
+def main():
+    """
+    Outputs commands for running the Debian version of Trimmomatic.
+    Put the path to the adapters to trim off as the first parameter
 
-for r1 in filelist:
-    r2 = r1.replace('R1', 'R2')
-    final_pe = r1[:-9] + '.pe.fq.gz'
-    final_se = r1[:-9] + '.se.fq.gz'
-    print """\
+    """
+    filelist = glob.glob('*R1*.fastq.gz')
+
+    for r1 in filelist:
+        r2 = r1.replace('R1', 'R2')
+        final_pe = r1[:-9] + '.pe.fq.gz'
+        final_se = r1[:-9] + '.se.fq.gz'
+        print """\
 mkdir trim
 cd trim
 TrimmomaticPE ../%s ../%s s1_pe s1_se s2_pe s2_se ILLUMINACLIP:%s:2:30:10
@@ -26,3 +32,6 @@ rm -r ./trim/
 
 chmod u-w %s %s
 """ % (r1, r2, sys.argv[1], final_pe, final_se, final_pe, final_se)
+
+if __name__ == '__main__':
+    main()
--- a/sandbox/write-trimmomatic.py
+++ b/sandbox/write-trimmomatic.py
@@ -8,12 +8,12 @@
 import glob
 import sys
 
-filelist = glob.glob(*R1*.fastq.gz)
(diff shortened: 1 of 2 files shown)


SherineAwad:
--I tested it, and the lines where quotes are removed fail. Does it work with you? 

--Also, my main problem with this function when I used it is that if the user has in the directory files named like this:
R11.fastq.gz R12.fastq.gz  R21.fastq.gz  R22.fastq.gz , or something similar, it won't work or it will miss files up. So either update the code or we can say somewhere explicitly that this function won't accept such names.

mr-c:
@SherineAwad good catch, I don't know what happened to those quote marks.

Do you have a fix for the R1 renaming problem?

SherineAwad:
I did not fix this. I used to consider that this function won't accept such names. But I can work on that if you like.

sandbox/write-trimmomatic.py
```diff
@@ -16,12 +17,12 @@
...
+interleave-reads.py s1_pe s2_pe | gzip -9c > ../%s
 
-cat s1_se s2_se | gzip -9c > ../%s
+cat s1_se s2_se | gzip -9c > ../../trimmed/%s
 cd ..
 rm -r ./trim/
 
 chmod u-w %s %s
-""" % (r1, r2, final_pe, final_se, final_pe, final_se)
+""" % (r1, r2, sys.argv[1], final_pe, final_se, final_pe, final_se)
```
SherineAwad: @mr-c  Why we need this "sys.argv[1]"? I think also function doesn't take arguments?

sandbox/write-trimmomatic.py
```diff
@@ -16,12 +17,12 @@
     print """\
 mkdir trim
 cd trim
-java -jar /usr/local/bin/trimmomatic-0.30.jar PE ../%s ../%s s1_pe s1_se s2_pe s2_se ILLUMINACLIP:/usr/local/share/adapters/TruSeq3-PE.fa:2:30:10
-/usr/local/share/khmer/scripts/interleave-reads.py s1_pe s2_pe | gzip -9c > ../%s
+TrimmomaticPE ../%s ../%s s1_pe s1_se s2_pe s2_se ILLUMINACLIP:%s:2:30:10
+interleave-reads.py s1_pe s2_pe | gzip -9c > ../%s
 
-cat s1_se s2_se | gzip -9c > ../%s
+cat s1_se s2_se | gzip -9c > ../../trimmed/%s
```
SherineAwad: @mr-c  What is trimmed?

sandbox/write-trimmomatic.py
```diff
@@ -16,12 +17,12 @@
     print """\
 mkdir trim
 cd trim
-java -jar /usr/local/bin/trimmomatic-0.30.jar PE ../%s ../%s s1_pe s1_se s2_pe s2_se ILLUMINACLIP:/usr/local/share/adapters/TruSeq3-PE.fa:2:30:10
-/usr/local/share/khmer/scripts/interleave-reads.py s1_pe s2_pe | gzip -9c > ../%s
+TrimmomaticPE ../%s ../%s s1_pe s1_se s2_pe s2_se ILLUMINACLIP:%s:2:30:10
```
SherineAwad: I think I got what you wanna do, but that's not working I think.

SherineAwad: TrimmpmaticPE? Not working I think.

SherineAwad:
=Function now takes arguments, when done, I will update that in the metagenome protocols. Others will need to too. 
=Usually the protocol has a working directory and this ../../trimmed will throw the output outside and user might be blind of this. 
=Other than that LGTM :)

PR #849, opened 2015-02-25
set 'force' to default to False on kfile.check_file_status.

ctb:
minor fix for some sandbox scripts, but I think this is a good default for this function.

#772 should test for this - I don't want to step all over @jessicamizzi's code :)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,6 +6,11 @@
    * tests/test_scripts.py: add a test for the --append option in
    normalize-by-median.py
 
+2015-02-25  Titus Brown  <titus@idyll.org>
+
+   * khmer/kfile.py: set 'force' flag for 'check_file_status' to default to
+   'False'.
+
 2015-02-25  Hussien Alameldin  <hussien@msu.edu>
 
    * khmer/khmer_args.py: add 'hll' citation entry "Irber and Brown,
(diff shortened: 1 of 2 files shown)


ctb:
- [ ] Is it mergable?
- [ ] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ctb:
Ready for review & merge @jessicamizzi @mr-c @luizirber @camillescott

ChangeLog
```diff
@@ -6,6 +6,11 @@
    * tests/test_scripts.py: add a test for the --append option in
    normalize-by-median.py
 
+2015-02-25  Titus Brown  <titus@idyll.org>
```
mr-c: New entries go to the top of the Changelog please

khmer/kfile.py
```diff
@@ -14,7 +14,7 @@
 from stat import S_ISBLK, S_ISFIFO
 
 
-def check_file_status(file_path, force):
+def check_file_status(file_path, force=False):
```
mr-c: I disagree. We left this out on purpose so that the scripts using this function will have to implement the force flag :-)

ctb:
Yeah, I thought that might have been your evil plan.

PR #856, opened 2015-02-26
check for errors while writing to disk

### LINKED ISSUE #443: Update C++ writing code to handle write exceptions properly.
Like #333, but for writing.  Subsumes #95 and #246.  Would address #411.

mr-c:
Will fix #443

### DIFF
--- a/khmer/__init__.py
+++ b/khmer/__init__.py
@@ -146,13 +146,16 @@ def extract_countinghash_info(filename):
     uint_size = len(pack('I', 0))
     ulonglong_size = len(pack('Q', 0))
 
-    with open(filename, 'rb') as countinghash:
-        version, = unpack('B', countinghash.read(1))
-        ht_type, = unpack('B', countinghash.read(1))
-        use_bigcount, = unpack('B', countinghash.read(1))
-        ksize, = unpack('I', countinghash.read(uint_size))
-        n_tables, = unpack('B', countinghash.read(1))
-        table_size, = unpack('Q', countinghash.read(ulonglong_size))
+    try:
+        with open(filename, 'rb') as countinghash:
+            version, = unpack('B', countinghash.read(1))
+            ht_type, = unpack('B', countinghash.read(1))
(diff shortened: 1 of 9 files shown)


mr-c:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

mr-c:
This PR lowers coverage due to extra IO error checking code and how hard it is to test that.

mr-c:
Jenkins, please retest this

mr-c:
Jenkins, please retest this

mr-c:
@luizirber @camillescott @ctb ready for review and merge

khmer/__init__.py
```diff
@@ -146,13 +146,16 @@ def extract_countinghash_info(filename):
...
-        table_size, = unpack('Q', countinghash.read(ulonglong_size))
+    try:
+        with open(filename, 'rb') as countinghash:
+            version, = unpack('B', countinghash.read(1))
+            ht_type, = unpack('B', countinghash.read(1))
+            use_bigcount, = unpack('B', countinghash.read(1))
+            ksize, = unpack('I', countinghash.read(uint_size))
+            n_tables, = unpack('B', countinghash.read(1))
+            table_size, = unpack('Q', countinghash.read(ulonglong_size))
+    except:
```
luizirber: Can this check be used on extract_hashbits_info too?

(at first I thought this could be more strict and catch only [struct.error][1], but there might be IOError too and in the end they are all a kind of corrupted input, so I think that's fine)

[1]: https://docs.python.org/2.7/library/struct.html#struct.error

luizirber:
It would be good to open an issue to add tests for these extra IO error checking code (per https://github.com/ged-lab/khmer/pull/856#issuecomment-87695006 )

mr-c:
Jenkins, retest this please

khmer/__init__.py
mr-c: Done, thanks for the suggestion.

PR #857, opened 2015-02-27
Make the help text of load-into-counting's '-b' flag more verbose

kdmurray91:
Hi all,

I confused myself this morning while trying to work out the effect of -b in load-into-counting.  Admittedly I was having a bit of a slow morning, but I got confused by the `--no-bigcounts (default=True)` help message. Does default=True mean that it's off or on by default?

This PR explains the behavior more verbosely in the help string, hopefully proofing against coffee deprived users :smile:.

Cheers,
K

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-02-27  Kevin Murray  <spam@kdmurray.id.au>
+
+   * scripts/load-into-counting.py: Be verbose in the help text, to clarify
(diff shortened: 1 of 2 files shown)


ctb:
LGTM. Pls make mergable at your convenience, and ping me when you do so :)

kdmurray91:
Rebased, so ping @ctb

Issue #660, opened 2014-11-16
Proposed change to coding guidelines: squashing commits


mr-c:
During the bulk of the development of a pull request I think the commits should fly fast and furious. However I would like to see them squashed down to a handful of commits (perhaps 1) during the refinement phase prior to merging.

Thoughts?

ctb:
I've also now dealt with several merge conflicts caused by rebasing and squashing on branches in PRs. It's a waste of my time and a potential source of bugs when playing around with other people's branches, and I think it needs to be strongly discouraged.

mr-c:
Maybe we all need a rebasing, merging, and cherry picking primer? People
should be able to do what they want with their own branches.

ctb:
On Sat, Jan 31, 2015 at 12:23:21AM -0800, Michael R. Crusoe wrote:
> Maybe we all need a rebasing, merging, and cherry picking primer? People
> should be able to do what they want with their own branches.

As soon as a branch is merged into another branch (as happened with HLL
and some of the SeqAn stuff) squashing and rebasing seems to cause problems.

> On Fri, Jan 30, 2015, 13:11 C. Titus Brown <notifications@github.com> wrote:
> 
> > I've also now dealt with several merge conflicts caused by rebasing and
> > squashing on branches in PRs. It's a waste of my time and a potential
> > source of bugs when playing around with other people's branches, and I
> > think it needs to be strongly discouraged.
> >
> > ???
> > Reply to this email directly or view it on GitHub
> > <https://github.com/ged-lab/khmer/issues/660#issuecomment-72192222>.
> >
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/ged-lab/khmer/issues/660#issuecomment-72309377
-- 
C. Titus Brown, ctbrown@ucdavis.edu

ctb:
http://matthew-brett.github.io/pydagogue/rebase_without_tears.html

wking:
[...]

PR #862, opened 2015-03-08
Implement optimized median filter

camillescott:
Easy optimization to improve normalize-by-median: it's simple to observe that checking if the median of a set is greater than some cutoff is equivalent to checking if more than half the elements of that set are greater than some cutoff. The latter avoids doing a lookup for every kmer every time, and avoids a costly sort. On a small dataset (1m ecoli reads), this was an 18% performance improvement. Not bad for 5 minutes!

TODO: Add a couple tests

### DIFF
--- a/khmer/_khmermodule.cc
+++ b/khmer/_khmermodule.cc
@@ -1151,6 +1151,32 @@ hashtable_get_median_count(khmer_KHashtable_Object * me, PyObject * args)
 
 static
 PyObject *
+hashtable_median_at_least(khmer_KHashtable_Object * me, PyObject * args)
+{
+    Hashtable * hashtable = me->hashtable;
+
+    const char * long_str;
+    unsigned int cutoff;
+
+    if (!PyArg_ParseTuple(args, "sI", &long_str, &cutoff)) {
+        return NULL;
+    }
+
+    if (strlen(long_str) < hashtable->ksize()) {
+        PyErr_SetString(PyExc_ValueError,
+                        "string length must >= the hashtable k-mer size");
+        return NULL;
+    }
+
+    if (hashtable->median_at_least(long_str, cutoff)) {
+        Py_RETURN_TRUE;
+    }
+    Py_RETURN_FALSE;
+
+}
+
+static
+PyObject *
 hashtable_n_tags(khmer_KHashtable_Object * me, PyObject * args)
 {
     Hashtable * hashtable = me->hashtable;
@@ -2302,6 +2328,7 @@ static PyMethodDef khmer_hashtable_methods[] = {
     { "find_all_tags_list", (PyCFunction)hashtable_find_all_tags_list, METH_VARARGS, "Find all tags within range of the given k-mer, return as list" },
     { "consume_fasta_and_tag", (PyCFunction)hashtable_consume_fasta_and_tag, METH_VARARGS, "Count all k-mers in a given file" },
     { "get_median_count", (PyCFunction)hashtable_get_median_count, METH_VARARGS, "Get the median, average, and stddev of the k-mer counts in the string" },
+    { "median_at_least", (PyCFunction)hashtable_median_at_least, METH_VARARGS, "Return true if the median is at least the given cutoff" },
     { "extract_unique_paths", (PyCFunction)hashtable_extract_unique_paths, METH_VARARGS, "" },
     { "load_stop_tags", (PyCFunction)hashtable_load_stop_tags, METH_VARARGS, "" },
     { "save_stop_tags", (PyCFunction)hashtable_save_stop_tags, METH_VARARGS, "" },
--- a/lib/hashtable.cc
+++ b/lib/hashtable.cc
@@ -232,6 +232,26 @@ void Hashtable::get_median_count(const std::string &s,
     median = counts[counts.size() / 2]; // rounds down
 }
 
+//
+// Optimized filter function for normalize-by-median
+//
+bool Hashtable::median_at_least(const std::string &s,
+                                unsigned int cutoff) {
+    KMerIterator kmers(s.c_str(), _ksize);
+    unsigned int min_req = 0.5 + float(s.size() - _ksize + 1) / 2;
+    unsigned int num_cutoff_kmers = 0;
+    while(!kmers.done()) {
+        HashIntoType kmer = kmers.next();
+        if (this->get_count(kmer) >= cutoff) {
+            ++num_cutoff_kmers;
+            if (num_cutoff_kmers >= min_req) {
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
 void Hashtable::save_tagset(std::string outfilename)
 {
     ofstream outfile(outfilename.c_str(), ios::binary);
--- a/lib/hashtable.hh
+++ b/lib/hashtable.hh
@@ -292,6 +292,9 @@ public:
         unsigned long long  &n_consumed
     );
 
+    bool median_at_least(const std::string &s,
+                          unsigned int cutoff);
+
     void get_median_count(const std::string &s,
                           BoundedCounterType &median,
                           float &average,
--- a/tests/test_counting_hash.py
+++ b/tests/test_counting_hash.py
@@ -217,6 +217,137 @@ def test_simple_median():
     assert int(stddev * 100) == 50        # .5
 
 
+def test_median_at_least():
+    hi = khmer.new_counting_hash(6, 1e6, 2)
+
+    hi.consume("AAAAAA")
+    assert hi.median_at_least("AAAAAA", 1)
+    assert hi.median_at_least("AAAAAA", 2) is False
+
+    hi.consume("AAAAAA")
+    assert hi.median_at_least("AAAAAA", 2)
+    assert hi.median_at_least("AAAAAA", 3) is False
+
+    hi.consume("AAAAAA")
+    assert hi.median_at_least("AAAAAA", 3)
+    assert hi.median_at_least("AAAAAA", 4) is False
+
+    hi.consume("AAAAAA")
+    assert hi.median_at_least("AAAAAA", 4)
+    assert hi.median_at_least("AAAAAA", 5) is False
+
+    hi.consume("AAAAAA")
+    assert hi.median_at_least("AAAAAA", 5)
+    assert hi.median_at_least("AAAAAA", 6) is False
+
+
+def test_median_at_least_single_gt():
+    K = 20
+    hi = khmer.new_counting_hash(K, 1e6, 2)
+
+    kmers = ['ATCGATCGATCGATCGATCG',
+             'GTACGTACGTACGTACGTAC',
+             'TTAGTTAGTTAGTTAGTTAG']
+
+    for kmer in kmers:
+        hi.consume(kmer)
+        assert hi.median_at_least(kmer, 1) is True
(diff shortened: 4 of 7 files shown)


mr-c:
Cool!

camillescott:
- [ ] Is it mergeable?
   - [ ] Did it pass the tests?
     fixing as needed.
     changes were made?

camillescott:
ready for review @ctb @mr-c

ctb:
Does this return identical results to old approach?

camillescott:
Yup. All normalize-by-median tests pass with the new function patched in.

camillescott:
jenkins, test this please you useless drunk

camillescott:
jenkins test this please and give it a pass or the floggings will continue

camillescott:
@ctb @mr-c @luizirber okey now pls review

ctb:
Please leave for me to merge - tnx.

camillescott:
@ctb I keep expecting that I've overlooked something really obvious and there's no way a 20% performance increase would be so easy, but *shrug*

khmer/_khmermodule.cc
```diff
@@ -938,6 +938,33 @@ hash_get_max_count(khmer_KCountingHash_Object * me, PyObject * args)
...
+        return NULL;
+    }
+
+    if (strlen(long_str) < counting->ksize()) {
+        PyErr_SetString(PyExc_ValueError,
+                        "string length must >= the hashtable k-mer size");
+        return NULL;
+    }
+
+    if (counting->filter_on_median(long_str, cutoff))
```
ctb: Please add { and } - single-line if statements are dangerous ;).

lib/hashtable.hh
```diff
@@ -291,6 +291,9 @@ public:
         unsigned long long  &n_consumed
     );
 
+    bool filter_on_median(const std::string &s,
+                        unsigned int cutoff);
```
ctb: spacing should be adjusted...

lib/hashtable.cc
```diff
@@ -231,6 +231,29 @@ void Hashtable::get_median_count(const std::string &s,
...
+//
+// Optimized filter function for normalize-by-median:
+// why do all the count lookups and an expensive sort
+// every time when we can just filter if more than half the
+// kmer counts are above the cutoff?
+//
+bool Hashtable::filter_on_median(const std::string &s,
+                                unsigned int cutoff) {
+    KMerIterator kmers(s.c_str(), _ksize);
+    unsigned int min_req = (s.size() - _ksize + 1) / 2;
```
ctb: Please round up - (0.5 + float(s.size() - _ksize + 1) / 2.0);

ctb:
A few comments --

the code didn't return identical results on data/100k-filtered.fa, due to a round-off error.  See rounding comment on hashtable.cc.

Please add tests for single k-mer, med < cutoff; single k-mer, med > cutoff; odd # k-mers, med < cutoff; odd # of k-mers; med > cutoff; and any other edge cases you can think of (that's all I got).  The homopolymer run tests are not sufficient ;).  You should enforce comparison with the old function, too, to make sure they both return the same results.

can we implement the old function without the sort? it seems fairly straightforward to do so although there would still be a performance hit.  This shouldn't be your job to do, but can you create a few issues around (a) improving old function, and (b) replacing old function with this function in places where it would work?

tnx!

camillescott:
@ctb can do on the extra tests and whatnot.

We definitely can't do the old function without the sort -- the key difference here is that the actual median isn't calculated, only that it's greater than some value. There are definitely more efficient median algorithms if we want to replace the naive sort-based one we're using now thoug

ctb:
On Mon, Mar 09, 2015 at 01:15:38PM -0700, Camille Scott wrote:
> @ctb can do on the extra tests and whatnot.

ok, let me know

> We definitely can't do the old function without the sort -- the key difference here is that the actual median isn't calculated, only that it's greater than some value. There are definitely more efficient median algorithms if we want to replace the naive sort-based one we're using now thoug

sure, we should look at the current code and think about where we might
optimize.

camillescott:
Stop being useless Jenkins, and dammit Jenkins, test this please!

camillescott:
Stop being useless Jenkins, and dammit Jenkins, test this please!

camillescott:
Stop being useless Jenkins, and dammit Jenkins, test this please!

camillescott:
@ctb, ready for review

khmer/_khmermodule.cc
ctb: Ping.

tests/test_counting_hash.py
```diff
@@ -212,6 +212,135 @@ def test_simple_median():
...
+            'GTACGTACGTACGTACGTACCCC',
+            'TTAGTTAGTTAGTTAGTTAGCCC']
+
+    for seq in seqs:
+        hi.consume(seq)
+        hi.consume(seq)
+        hi.consume(seq)
+
+        med, _, _ = hi.get_median_count(seq)
+        assert hi.filter_on_median(seq, 4) is (med >= C)
```
ctb: replace 4 with C!

tests/test_counting_hash.py
```diff
@@ -212,6 +212,135 @@ def test_simple_median():
...
+    seqs = ['ATCGATCGATCGATCGATCGCCC',
+            'GTACGTACGTACGTACGTACCCC',
+            'TTAGTTAGTTAGTTAGTTAGCCC']
+
+    for seq in seqs:
+        hi.consume(seq)
+        assert hi.filter_on_median(seq, 1) is True
+
+
+def test_filter_on_median_even_lt():
```
ctb: please annotate the difference from 'test_filter_on_median_odd_lt' (seqs are different lengths)

ctb:
Two questions for consideration cc @mr-c --

* should we store the results of normalize-by-median on (say) 100k-filtered.fa, as a test for consistency going forward?

* filter_on_median is an unsatisfying function name, since it doesn't actually filter anything. Would 'median_greater_than' or something like that be clearer? Suggestions welcome.

mr-c:
+1 for storing known-good output. How big, when compressed? Worse case we keep it elsewhere and let Jenkins deal with it.

camillescott:
Very well -- I'll add a comparison test to known output of the mentioned file and change the function name.

camillescott:
@ctb @mr-c okay I added the new test -- it adds ~5mb the repo. Is that manageable?

mr-c:
Hrmm.. that would nearly double the distribution size..

camillescott:
@ctb @mr-c further thoughts on this??

mr-c:
We need a separate repo to store larger pieces of test data and have tests that run only on Jenkins by default. @ctb What do you think?

bocajnotnef:
Would something like git's LFS work?

https://github.com/blog/1986-announcing-git-large-file-storage-lfs

mr-c:
@bocajnotnef In this case the file is 5Megabytes so GitHub LFS wouldn't be needed. I don't want to make the git clone of the main repository larger than necessary. Another option is to add the 5M file but exclude it from packaging and adjust the test to skip if the file isn't present.

ctb:
On Thu, Apr 09, 2015 at 08:55:01AM -0700, Michael R. Crusoe wrote:
> @bocajnotnef In this case the file is 5Megabytes so GitHub LFS wouldn't be needed. I don't want to make the git clone of the main repository larger than necessary. Another option is to add the 5M file but exclude it from packaging and adjust the test to skip if the file isn't present.

What about a 'khmer-data' repository?  Although that seems kind of silly.

ctb:
On Thu, Apr 09, 2015 at 08:19:19AM -0700, Michael R. Crusoe wrote:
> We need a separate repo to store larger pieces of test data and have tests that run only on Jenkins by default. @ctb What do you think?

Since the data is mostly going to be read-only, it seems silly to have a full
repo.  Maybe their new data extension *is* the way to go for that?  And have a
khmer-data repo that stores the pointers?

camillescott:
How about we get the optimization merged for now, then open an issue for supporting github LFS or some other variety of data hosting to be addressed in the future?

mr-c:
+1

ctb:
- [ ] create an issue
- [ ] make mergable

ctb:
@camillescott - @mr-c is going to address data size problem for test by putting that elsewhere, but other than that, all you need to do is make it mergable, I believe.

mr-c:
It can go into https://github.com/dib-lab/khmer-testdata ; just commit directly

ctb:
ping @camillescott

camillescott:
@mr-c need access to khmer-testdata repo

camillescott:
jenkins, test this please

Blahah:
There is so much winning in this PR, @camillescott.

ctb:
Suggest squashing as in #660.

camillescott:
Jenkins, test this please

camillescott:
@ctb ready for final review / merge

camillescott:
also pleeeeeeeeeease merge this before merging any other PR's in, it's all squashed and pretty and I've had to update and tweak little things on this branch way too many times :'(

ctb:
LGTM; nice work.

camillescott:
Yay, thanks!

lib/hashtable.cc
```diff
@@ -232,6 +232,26 @@ void Hashtable::get_median_count(const std::string &s,
...
+bool Hashtable::median_at_least(const std::string &s,
+                                unsigned int cutoff) {
+    KMerIterator kmers(s.c_str(), _ksize);
+    unsigned int min_req = 0.5 + float(s.size() - _ksize + 1) / 2;
+    unsigned int num_cutoff_kmers = 0;
+    while(!kmers.done()) {
+        HashIntoType kmer = kmers.next();
+        if (this->get_count(kmer) >= cutoff) {
+            ++num_cutoff_kmers;
+            if (num_cutoff_kmers >= min_req) {
```
ctb: Random thought @camillescott - what if you did a for or a while loop up until min_req, and only then checked the 'if'? Obviously if num_kmers < min_req there's no point in checking.

camillescott: :cries: that's probably fair, though I imagine the compiler + cpu does a fair job with it's branch prediction in this scenario

ctb: On Tue, Jun 09, 2015 at 10:20:17AM -0700, Camille Scott wrote:
> > @@ -232,6 +232,26 @@ void Hashtable::get_median_count(const std::string &s,
> >      median = counts[counts.size() / 2]; // rounds down
> >  }
> >  
> > +//
> > +// Optimized filter function for normalize-by-median
> > +//
> > +bool Hashtable::median_at_least(const std::string &s,
> > +                                unsigned int cutoff) {
> > +    KMerIterator kmers(s.c_str(), _ksize);
> > +    unsigned int min_req = 0.5 + float(s.size() - _ksize + 1) / 2;
> > +    unsigned int num_cutoff_kmers = 0;
> > +    while(!kmers.done()) {
> > +        HashIntoType kmer = kmers.next();
> > +        if (this->get_count(kmer) >= cutoff) {
> > +            ++num_cutoff_kmers;
> > +            if (num_cutoff_kmers >= min_req) {
> 
> :cries: that's probably fair, though I imagine the compiler + cpu does a fair job with it's branch prediction in this scenario

:)

Issue #868, opened 2015-03-11
test_get_raw_tables_view fails on Python 2.7.2


ctb:
```
  File "/Users/t/dev/khmer/tests/test_counting_hash.py", line 117, in test_get_raw_tables_view
    memv = memoryview(tab)
TypeError: cannot make memory view because object does not have the buffer interface
```
c.f. #671

kdmurray91:
Oh balls. Will investigate.

I think the issue may be old/new style buffers in CPython???

I was examining how to efficiently pass these to Cython as memory views and hit upon a similar issue (to do with read only buffers).

kdmurray91:
Although that's rather intriguing, as `type(tab) is buffer`. And you'd hope that a buffer exposed the buffer interface :grinning:.

luizirber:
This also breaks Python 2.6, but I guess we're not aiming for 2.6 anymore.
``` python
Traceback (most recent call last):
  File "khmer/.tox/py26/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "khmer/tests/test_counting_hash.py", line 117, in test_get_raw_tables_view
    memv = memoryview(tab)
NameError: global name 'memoryview' is not defined
```

kdmurray91:
Mmmh, not good. I'll add a fix for @luizirber's observation.

kdmurray91:
I haven't been able to find any python bugs/changes that indicate this is a python problem, but that may be my poor google skills.

kdmurray91:
@luizirber any chance you could post your tox.ini? or even check it in, if that's ok.

luizirber:
I didn't make a PR because it needs more polishing, for now I'm using something that works but is also wrong (the ideal would be moving parts from the Makefile here, and then calling tox on the Makefile)

``` ini
[tox]
envlist=py26,py27

[testenv]
commands=
    make install-dependencies
    make clean
    make test
```

ctb:
Note, further investigation suggests that it is a problem in Python 2.7.2 but not Python 2.7.6.

kdmurray91:
Interesingly, while testing on another machine (supercomputer using ICC), I got this:

```
khmer/_khmermodule.cc: In function 'PyObject* hash_get_raw_tables(khmer_KCountingHash_Object*, PyObject*)':
khmer/_khmermodule.cc:691: error: reference to 'Byte' is ambiguous
[...]

PR #869, opened 2015-03-12
Skip the get_raw_tables test if python doesn't  have memoryview

kdmurray91:
This addresses @luizirber's comment in #868 that python2.6 chokes on this test.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,8 +1,12 @@
+2015-03-15  Kevin Murray  <spam@kdmurray.id.au>  &  Titus Brown  <titus@idyll.org>
+
+   * tests/test_counting_hash.py: Skip get_raw_tables test if python doesn't
+   have the memoryview type/function.
 
 2015-03-11  Erich Schwarz  <ems394@cornell.edu>
 
-   * Added URLs and brief descriptions for khmer-relevant documentation in 
-   doc/introduction.txt, pointing to http://khmer-protocols.readthedocs.org and 
+   * Added URLs and brief descriptions for khmer-relevant documentation in
+   doc/introduction.txt, pointing to http://khmer-protocols.readthedocs.org and
    khmer-recipes.readthedocs.org, with brief descriptions of their content.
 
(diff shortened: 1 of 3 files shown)


kdmurray91:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

kdmurray91:
Ready for review

mr-c:
Thanks @kdmurray91. I, and others, don't automatically get notified unless we are already subscribed to an issue so an @ mention will get our attention :-)

ctb:
ping @luizirber can you check this on 2.6?

Also, @mr-c, what do we want to do about 2.7.2 vs 2.7.6? The error in #868 is seen in 2.7.2.  I think we should put in the same SkipTest if we get a type error on the memoryview call, but alternative suggestions welcome.

mr-c:
The skiptest is fine; it is an experimental & unsupported API either way :-)

ctb:
I've set up a PR #871 against this branch that makes the tests pass on Python 2.6.2.

kdmurray91:
Rebased on master, should be mergeable.

ChangeLog
```diff
@@ -1,8 +1,12 @@
+2015-03-15  Kevin Murray  <spam@kdmurray.id.au>  &  Titus Brown  <titus@idyll.org>
```
kdmurray91: @mr-c  @ctb  is this the right way to credit 2 people in a CL entry?

ctb: On Sat, Mar 14, 2015 at 03:57:28PM -0700, Kevin Murray wrote:
> > @@ -1,8 +1,12 @@
> > +2015-03-15  Kevin Murray  <spam@kdmurray.id.au>  &  Titus Brown  <titus@idyll.org>
> 
> @mr-c  @ctb  is this the right way to credit 2 people in a CL entry?

Oh, no need to credit... seriously... :)

kdmurray91:
Can confirm this works w/ 2.7.9 and 2.7.6, which is all I have access to.

mr-c:
LGTM

PR #876, opened 2015-03-18
Thread-safety for hashtable

camillescott:
The multithreading PR has gotten unwieldy, so this is part of the process to break it down into more manageable chunks. This set of commits adds an array of locks to Hashtable with are indexed by the hashvalue, allowing a thread to lock only a small block when writing. It also introduces a new function meant specifically for thread-safe counting, to avoid overhead under single-threaded cases and maintain backwards compatibility.

### DIFF
--- a/lib/counting.hh
+++ b/lib/counting.hh
@@ -94,6 +94,17 @@ public:
         return _counts;
     }
 
+    virtual void init_threadstuff(unsigned int block_size=TABLE_BLOCK_SIZE) {
+        HashIntoType _max_size = _tablesizes.back();
+        _n_table_blocks = (_max_size < block_size) ? 1 : (_max_size / block_size);
+        //std::cout << "Table lock blocks: " << _n_table_blocks << std::endl;
+        _table_spinlocks = new uint32_t[_n_table_blocks];
+        for (unsigned int i=0; i<_n_table_blocks; ++i) {
+            _table_spinlocks[i] = 0;
+        }
+        _threadsafe = true;
+    }
+
     virtual BoundedCounterType test_and_set_bits(const char * kmer)
     {
         BoundedCounterType x = get_count(kmer); // @CTB just hash it, yo.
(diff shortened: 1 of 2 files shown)


camillescott:
- [ ] Is it mergeable?
   - [ ] Did it pass the tests?
     fixing as needed.
     changes were made?

camillescott:
Jenkins, test this please, or we'll be having another stern talk.

camillescott:
Dammit Jenkis, RESPECT MY AUTHORITEH (and test this please)

camillescott:
You ready to test this please, Jenkins?

camillescott:
@mr-c @ctb @luizirber thoughts? is the testing robust enough? does a separate function (`count_ts`) make sense?

camillescott:
Jenkins, have you been tippin' the bottle again? Test this please!

camillescott:
Have you been tippin' the bottle again? Jenkins, test this please!

khmer/_khmermodule.cc
```diff
@@ -1541,10 +1618,14 @@ static PyMethodDef khmer_counting_methods[] = {
     { "hashsizes", (PyCFunction)hash_get_hashsizes, METH_VARARGS, "" },
     { "set_use_bigcount", (PyCFunction)hash_set_use_bigcount, METH_VARARGS, "" },
     { "get_use_bigcount", (PyCFunction)hash_get_use_bigcount, METH_VARARGS, "" },
+    { "init_threadsafe", (PyCFunction)hash_init_threadsafe, METH_VARARGS, "Initialize the table lock array" },
+    { "n_lock_blocks", (PyCFunction)hash_n_lock_blocks, METH_VARARGS, "Number of blocks in the lock array" },
```
mr-c: Shouldn't this be `_` prefixed to indicate that it is a private/testing function?

camillescott: The number of lock blocks is passed when calling `init_threadsafe`. It has a default value that is hardcoded, but the user can override it; there isn't necessary anything private or testing-specific about it. I will make it const const though.

mr-c:
The current `count()` function https://github.com/ged-lab/khmer/blob/feature/thread_safety/lib/counting.hh#L175 looks to be half thread-safe.

Is there really that much overhead when single-threaded?

I'd rather see the extra thread safety added to `count()`. If there is significant slowdown when single threaded then we can put all the spinlocks (old and new) inside `if` statements.

camillescott:
@mr-c: I'm not sure on the overhead. Some recent reading I did suggests that CAS operations have significant overhead when called repeatedly (for example: http://joeduffyblog.com/2009/01/08/some-performance-implications-of-cas-operations/). However, I'll need to run some benchmarks to figure out the effects of this particular implementation.

mr-c:
(Also, thanks!)

lib/counting.hh
```diff
@@ -94,6 +94,17 @@ public:
         return _counts;
     }
 
+    virtual void init_threadstuff(unsigned int n_table_locks=DEFAULT_TABLE_LOCKS) {
+        if(!_threadsafe) {
+            _n_table_locks = n_table_locks < 4 ? 4 : n_table_locks;
```
ctb: why '4'?

lib/counting.hh
```diff
@@ -202,6 +213,51 @@ public:
...
+        count_ts(hash);
+    }
+
+    // Thread-safe counting function
+    virtual void count_ts(HashIntoType khash)
+    {
+        bool is_new_kmer = true;
+        unsigned int  n_full      = 0;
+        uint32_t lock_index = khash % _n_table_locks;
+        while(!__sync_bool_compare_and_swap( &_table_spinlocks[lock_index], 0, 1));
```
ctb: for readability, maybe put a comment (or make a macro) that says "take lock", and "release lock" (below).

ctb:
+1 on @mr-c's comment regarding adding thread-safety to count instead of building a new function.  Other than that, and a few questions, looks good and straightforward.  The testing is OK although I doubt it's going to hit any problematic cases even w/o thread safety (might be interesting to test this by disabling thread safety and running that test a few hundred times... does it ever break?)

I'm assuming that the real purpose of this addition is not to use the Python interface, right? The Python interface is not going to give you a massive speed improvement here, I don't think!

camillescott:
(Putting some of our Friday discussion here just to have it written down somewhere)

> The testing is OK although I doubt it's going to hit any problematic cases even w/o thread safety (might be interesting to test this by disabling thread safety and running that test a few hundred times... does it ever break?)

It seems that it takes longer for the Python interpreter to load in the next bit of bytecode and execute it than it does for the `count_ts` function to be called -- I couldn't get this to trip from Python land.

> I'm assuming that the real purpose of this addition is not to use the Python interface, right? The Python interface is not going to give you a massive speed improvement here, I don't think!

Correct -- I was trying to figure out a way to test it with our current suite. At this point, it looks like more thorough testing will have to wait for the full multithreading PR.

Issue #676, opened 2014-12-04
Scripts should output metadata file about what they did


mr-c:
While most of the file formats we work with don't have support for metadata we aren't off the hook for recording such information and preserving it.

Based upon the discussion captured here: https://groups.google.com/d/msg/common-workflow-language/wx8G2zvDUV4/lzZPUPtQEHwJ we should store such information in a JSON file (like @kdmurray91's addition to load-into-counting.py)

ctb:
A few more thoughts --

salmon does this now!

also see #189, generating PROV-compliant output.

I always get hung up on what to do when we are running many things in a row. Do we just have a file that we append to? or perhaps a directory that contains multiple little files with UUIDs containing the provenance output?

is there any workflow-supported standard emerging? is PROV it?

standage:
Brainstorming some considerations.

- probably want to use a filename format similar to `2017-04-03T06:55:21-khmer.json` (maybe `s/khmer/oxli/`) 
    - fixed length
    - sorts correctly every time without any tricks
- store a distinct file for each script invocation in some directory?
    - *in situ* with something like `./.oxli/` or `./oxli-meta/`?
    - global with something like `~/.oxli`?
- ...or use a user-specified file to keep a running log of JSON entries?
    - if nothing is specified, default to something like `./oxli-meta.json`
- wait until the end of the script (to confirm successful exit) to print?
    - otherwise could end up with a lot of unhelpful/misleading files/entries
- report relative paths (for portability) or absolute paths (for clearest provenance)?

standage:
> salmon does this now!

The most recent stable release, or current master? I just compiled the latest release from source and ran a few commands but didn't see any JSON metadata or any way to invoke it.

ctb:
On Mon, Apr 03, 2017 at 09:13:24AM -0700, Daniel Standage wrote:
> > salmon does this now!
> 
> The most recent stable release, or current master? I just compiled the latest release from source and ran a few commands but didn't see any JSON metadata or any way to invoke it.

it's in the quant directory that contains quant.sf.

standage:
👍  Got it. Also several .json files in the index directory.

ctb:
On Mon, Apr 03, 2017 at 06:15:59AM -0700, Daniel Standage wrote:
> - report relative paths (for portability) or absolute paths (for clearest provenance)?

both!

we could also use a sqlite database to track entries.

standage:
[...]

PR #877, opened 2015-03-19
Initial implementation of read counting in C++  read parser

kdmurray91:
Hi all,

I've been thinking about #676, and decided to jump right in with one pre-requisite of metadata reporting, counting reads.

This is an implementation of read counting for the C++ parser(s), and an accessor for python-land objects.

It appears threadsafe, from my hacky testing (increments are done within spinlocked code blocks). I'll test further, and write tests etc as time permits :)

Cheers,
K

### DIFF
--- a/khmer/_khmermodule.cc
+++ b/khmer/_khmermodule.cc
@@ -466,6 +466,20 @@ ReadParser_iter_reads(PyObject * self, PyObject * args )
     return PyObject_SelfIter( self );
 }
 
+static
+PyObject *
+ReadParser_get_num_reads(khmer_ReadParser_Object * me, PyObject * args)
+{
+    read_parsers:: IParser *  parser = me->parser;
+
+    if (!PyArg_ParseTuple(args, "")) {
+        return NULL;
+    }
+
+    unsigned long n_reads = parser->get_num_reads();
+
+    return PyLong_FromLong(n_reads);
+}
 
 static
 PyObject *
@@ -504,6 +518,10 @@ static PyMethodDef _ReadParser_methods [ ] = {
     {
         "iter_read_pairs",  (PyCFunction)ReadParser_iter_read_pairs,
         METH_VARARGS,       "Iterates over paired reads as pairs."
+    },
+    {
+        "get_num_reads",  (PyCFunction)ReadParser_get_num_reads,
+        METH_VARARGS,       "Returns the current number of reads processed."
     },
 
     { NULL, NULL, 0, NULL } // sentinel
--- a/khmer/_khmermodule.cc
+++ b/khmer/_khmermodule.cc
@@ -468,17 +468,9 @@ ReadParser_iter_reads(PyObject * self, PyObject * args )
 
 static
 PyObject *
-ReadParser_get_num_reads(khmer_ReadParser_Object * me, PyObject * args)
+ReadParser_get_num_reads(khmer_ReadParser_Object * me)
 {
-    read_parsers:: IParser *  parser = me->parser;
-
-    if (!PyArg_ParseTuple(args, "")) {
-        return NULL;
-    }
-
-    unsigned long n_reads = parser->get_num_reads();
-
-    return PyLong_FromLong(n_reads);
+    return PyLong_FromLong(me->parser->get_num_reads());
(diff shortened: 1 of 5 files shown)


kdmurray91:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

kdmurray91:
  entries
     the `ob_size` field).
     `tp_str` & `tp_repr`
  in `init_khmer()`

kdmurray91:
ping @mr-c, since #676 is your issue :)

kdmurray91:
Also, any thoughts about `_num_reads` vs `_read_count` as the method's name?

kdmurray91:
retest this please

khmer/_khmermodule.cc
```diff
@@ -466,6 +466,20 @@ ReadParser_iter_reads(PyObject * self, PyObject * args )
     return PyObject_SelfIter( self );
 }
 
+static
+PyObject *
+ReadParser_get_num_reads(khmer_ReadParser_Object * me, PyObject * args)
+{
```
mr-c: `readparser.get_num_reads()` is a bit Java-y. You can implement this value as a simple attribute (accessed via `readparser.num_reads` via `tp_members` (https://docs.python.org/2/extending/newtypes.html#generic-attribute-management)

kdmurray91: Yup, my thoughts exactly. A pox on all your `.class`es.

Was unaware of `tp_members`, my `Python.h` knowledge is limited :)

Will investigate this.

kdmurray91: Hang on, how about using `getsetters`, *a la* the `HLLCounter`'s `alpha` and `error_rate`? From what I can see, it would be difficult to expose a protected member of the IParser class via the khmer_ReadParser_Object struct.

However, would love to stand corrected

kdmurray91: Implemented accessor-based access in  526fe19

kdmurray91:
Ready for review, modulo a possible name change 'num_reads' => 'read_count'

kdmurray91:
Rebased on master, conflicts resolved.

lib/read_parsers.cc
```diff
@@ -57,6 +58,7 @@ void SeqAnParser::imprint_next_read(Read &the_read)
     if (!atEnd) {
         ret = seqan::readRecord(the_read.name, the_read.sequence,
                                 the_read.quality, _private->stream);
+        _num_reads++;
```
mr-c: this should only be incremented if `ret == 0`

lib/read_parsers.cc
```diff
@@ -41,6 +41,7 @@ SeqAnParser::SeqAnParser( char const * filename ) : IParser( )
     }
     __asm__ __volatile__ ("" ::: "memory");
     _private->seqan_spin_lock = 0;
+    _num_reads = 0;
```
mr-c: Isn't this redundant with line 120? https://github.com/ged-lab/khmer/pull/877/files#diff-be310f56bb83cef45967cfbe13f0adeeR120

kdmurray91: Will the constructor be inherited like that? I never got my head around ctor inheritance in C++ to be honest. My understanding was that it doesn't automatically, but I'm probably wrong. In which case yes, it's redundant.

lib/read_parsers.cc
kdmurray91: Indeed, fixed

lib/read_parsers.cc
mr-c: Since `_num_reads` is initialized in the no argument constructor of the superclass there is no need to re-initialize it. 

http://stackoverflow.com/a/120943

kdmurray91: Ah ha! Thanks for that S/O post, i finally get it :)

I'll make the change now.

mr-c:
LGTM, make it mergeable and I'll do so.

kdmurray91:
That should be good to go @mr-c

mr-c:
Thanks @kdmurray91 !

PR #882, opened 2015-03-26
Option for printing diginorm output to stdout. Closes #881.

### LINKED ISSUE #881: Option to send normalize-by-median.py output to stdout
Is this currently possible? If not, implementing the option would facilitate piping output to other commands.

standage:
Closes #881 

- [x] Is it mergable?
- [x] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?


mr-c:
@standage does passing in `-` or `/dev/stdout` not work for you?

standage:
Using `-` does not work on my MacBook Pro (Yosemite), it interprets the character as a filename. I had forgotten about the `/dev/stdout` trick though. That seemed to work just fine. I'll change my PR so that it supports `-` instead of `stdout`, although I'm okay if this is deemed unnecessary.

standage:
Huh...this time the CI only ran for 24 seconds before failing. :-/

mr-c:
It should work with `-`; if not we should add that to screed directly.

mr-c:
Jenkins, retest this please

standage:
Maybe I'm missing something, but AFAICS it doesn't look like the script is using screed for output. It is calling the `write_record` function, which is part of the ``khmer::utils`` module and doesn't use any screed code.

mr-c:
Yes, quite right :-) Lets solve this using http://stackoverflow.com/a/11721741 for this script

Once we have tests showing that this approach works we can generalize it for the other scripts (or wait until oxli does the same).

standage:
Ok, sounds good. I'll fix my PR to use the built-in argparse support to do this.

mr-c:
Thanks!

standage:
Hmm...On second look, I think the current code might be cleaner than the suggested alternative given the current interface. In order to support the `--append` option, we would have to get very creative with the `argparse` module to determine whether to use `type=argparse.FileType('w')` or `type=argparse.FileType('a')`. I'm open to suggestions, but right now I can't think of any way of doing this that would be any cleaner than what we have now.

mr-c:
Appending it a type of writing; does it do the right thing for you when the mode is `a`?

mr-c:
Good catch; looks like they forgot to code for mode `a`. I would get around that like this:

``` diff
diff --git a/scripts/normalize-by-median.py b/scripts/normalize-by-median.py
index 050aaad..4dded1f 100755
--- a/scripts/normalize-by-median.py
+++ b/scripts/normalize-by-median.py
@@ -238,7 +238,7 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
 
     if args.single_output_filename:
         output_name = args.single_output_filename
-        if args.append:
+        if args.append and args.single_output_filename is not '-':
             outfp = open(args.single_output_filename, 'a')
         else:
             outfp = open(args.single_output_filename, 'w')
```

standage:
When the mode is `a`, `-` is an invalid value--`argparse` throws an error. Also, say the user did not use the `--append` option and was therefore expecting an existing output file to be overwritten. Using `a` would append in every case, even when overwriting is the correct behavior.

standage:
With your suggested fix, a file named `-` is created.

standage:
Here's my assessment.
- Python's vanilla `open` machinery doesn't treat the filename `-` specially. Therefore if we want to support the convention that it indicates stdout, then we need to code that (my latest commit does just that).
- The `argparse` module does have support for the `-` = `stdout` convention, but because the script also supports an append option (when not printing to stdout), the output file mode cannot be determined at argument parsing time.

The latest commit on this PR is still the cleanest, IMO.

mr-c:
Ah yes, I was intending for my diff to be combined with the `type=argparse.FileType` approach but I see now how that doesn't work out.

LGTM so far; lets add some tests

standage:
Soooo...I added a new test in `tests/test_scripts.py`. But it looks like using shell file redirection (the `>` character) with the normal test scaffolding is problematic. Any pointers would be appreciated.

camillescott:
Why not just use `argparse.FileType` and if the user wants to append, they can write to stdout and use `>>` instead of `>`?

standage:
@camillescott  That would work for stdout, but would break functionality for writing directly to a file. You have to specify a mode, right? Which mode are you going to specify if you want to write directly to a file rather than stdout?

camillescott:
Direct to file:
`diginorm.py <stuff> -o <filename>` or `diginorm.py <stuff> > <filename>`

Append to file:
`diginorm.py <stuff> >> <filename>`

standage:
@camillescott Doesn't that break the ``--append`` mode, which can be called whether or not the ``-o`` option is used?

camillescott:
I guess what I'm saying is that with proper output to `stdout` there's no
need for `--append` mode

On Thu, Mar 26, 2015 at 4:21 PM, Daniel Standage <notifications@github.com>
wrote:

> @camillescott <https://github.com/camillescott> Doesn't that break the
> --append mode, which can be called whether or not the -o option is used?
>
> —
> <https://github.com/ged-lab/khmer/pull/882#issuecomment-86698631>.
>



-- 
Camille Scott

Department of Computer Science and Engineering
Lab for Genomics, Evolution, and Development
Michigan State University

camille.scott.w@gmail.com

standage:
@camillescott I'm guessing the append mode is around to make fault tolerance and recovery possible. If you have 5 input files and want 5 output files, then specifying mode `w` will overwrite each output file even if you wanted to append instead.

standage:
@camillescott I dunno, maybe I'm missing something really simple. But I have a suspicion that the ``--append`` mode, the fault tolerance & recovery options, and support for multiple input/output files (with `.keep` extension) are the reasons the ``argparse.FileType`` wasn't being used for the `-o` option in the first place.

mr-c:
See also #745 & #843. The good news is that this hasn't been released yet so we can change the behavior if we decide to.

standage:
With a closer reading of the script's interface, and the fact that the `--append` option hasn't yet been released, I think @camillescott is right. Appending never was supported for multiple output files, and is easily achieved for a single output file without having a dedicated option. This last commit removed the `--append` option and implemented the `-o` option with `argparse.FileType('w')`. Overall I think everything looks a lot better.

That still leaves the problem of the functional test--how to redirect stdout to a file. Placing the file redirect character in the array of arguments doesn't play nice with the testing framework.

ctb:
Try shell=True

standage:
I'm familiar with `shell=True` for the subprocess module. Where would I put that in `test_scripts.py`? It doesn't seem to work with `khmer_tst_utils::runscript`.

mr-c:
Look at https://github.com/ged-lab/khmer/blob/master/tests/test_scripts.py#L2290 to see how we tested streaming _into_ `normalize-by-median.py`

standage:
@mr-c Thanks, I'll take a look!

mr-c:
Any chance you could wrap this up in the next 24 hours? Otherwise I'll finish it up for the next khmer release.

scripts/normalize-by-median.py
```diff
@@ -189,12 +189,10 @@ def get_parser():
...
-                        dest='single_output_filename',
-                        default='', help='only output a single'
-                        ' file with the specified filename')
-    parser.add_argument('--append', default=False, action='store_true',
-                        help='append reads to the outputfile. '
-                        'Only with -o specified')
+                        dest='single_output_file',
+                        type=argparse.FileType('w'),
+                        default='', help='only output a single file with the '
+                        'specified filename; use - to print to the terminal')
```
mr-c: I would say: "use a single dash '-' to specify that output should go to STDOUT (the terminal)."

And then in the epilog we need to explain how to append the output to an existing file with an example.

standage:
I'll give the testing another shot, that was really all that was left. I'll obviously also have to merge in the other commits as well before this is mergeable. Hopefully the only conflicts will be in the change log.

standage:
Retest this please Jenkins

standage:
Checklist complete, CI passing, and mergeable. Ready for code review @mr-c/@luizirber.

ChangeLog
```diff
@@ -107,6 +115,7 @@
    * lib/.gitignore,third-party/.gitignore: Add more compiled outputs
    * lib/.check_openmp.cc: add source that checks compiler for openmp support.
    * lib/khmer.pc.in: add pkg-config file for khmer
+>>>>>>> 6ea6e65d9eb013712579953f26d966412427d555
```
luizirber: oops! remove this please =]

scripts/normalize-by-median.py
```diff
@@ -151,6 +151,10 @@ def get_parser():
     table up to that point will be dumped, and processing will continue on the
     next file.
 
+    To append reads to an output file (rather than overwriting it), send output
+    to STDOUT with `--out -` and use UNIX file redirection syntax (`>>`) to
+    append to the file.
```
mr-c: Please add an example of this below

standage:
Stop being a jerk and retest this Jenkins

standage:
Retest this please Jenkins

standage:
Anything else?

standage:
Anything else?

standage:
Are you referring to the dash? That is indeed a valid path, and a convention used by many Unix tools. It was one of the motivations for this PR.

standage:
Oh, I see.

standage:
Jenkins retest this please.

standage:
Please merge this now! ;)

scripts/normalize-by-median.py
```diff
@@ -162,6 +166,11 @@ def get_parser():
 
     Example::
 
+""" "        normalize-by-median.py -p -k 17 -o - tests/test-data/test-fastq-reads.fq >> appended-output.fq"  # noqa
```
mr-c: This command fails

``` bash
(env)mcrusoe@athyra:~/khmer/gl-master$ normalize-by-median.py -p -k 17 -o - tests/test-data/test-fastq-reads.fq >> appended-output.fq

|| This is the script 'normalize-by-median.py' in khmer.
|| You are running khmer version 1.4-rc1-20-gb589b17-dirty
|| You are also using screed version 0.8
||
|| If you use this script in a publication, please cite EACH of the following:
||
||   * MR Crusoe et al., 2014. http://dx.doi.org/10.6084/m9.figshare.979190
||   * CT Brown et al., arXiv:1203.4802 [q-bio.GN]
||
|| Please see http://khmer.readthedocs.org/en/latest/citations.html for details.


PARAMETERS:
 - kmer size =    17            (-k)
 - n tables =     4             (-N)
 - min tablesize = 1e+06        (-x)

Estimated memory usage is 4e+06 bytes (n_tables x min_tablesize)
--------
making k-mer counting table
** ERROR: Error: Improperly interleaved pairs                     895:1:1:1246:14654 1:N:0:NNNNN 895:1:1:1248:9583 1:N:0:NNNNN
** Failed on tests/test-data/test-fastq-reads.fq: 
** ...dumping k-mer counting table to test-fastq-reads.fq.ct.failed
** ERROR: problem removing corrupt filtered file
** Exiting!
```

standage:
Ready to merge...at least for the next few minutes :)

Issue #884, opened 2015-03-29
Behaviour of SeqAnParser with a truncated fastq


kdmurray91:
This is a bit of a strange question, but here goes.

While making a test for #877, I made the following FASTQ file:

```
@1:1:1:1:1 1:N:0:NNNNN
CAGGCGCCCA
+
][aaX__aa[
@1:1:1:1:2 1:N:0:NNNNN
CAGGCGCCCA
```

I.e., a valid record, followed by a record truncated after the sequence.

From my understanding of the format, and certainly from a pragmatic point, this is an invalid file. But the seqan parser silently accepts it, leaving `read.quality := ""`.

And there's no indication of an invalid record (i.e., `seqan::readRecord` returns 0, no exceptions are thrown).

I'm fairly certain that it's me who is wrong and this is some corner case of the format, but it smells strange to me somehow.

Cheers,
Kevin

kdmurray91:
Further testing indicates the same happens with

```
@1:1:1:1:1 1:N:0:NNNNN
CAGGCGCCCA
+
][aaX__aa[
@1:1:1:1:2 1:N:0:NNNNN
```

which *has* to be invalid, surely :)

kdmurray91:
And even this:

```
@1:1:1:1:1 1:N:0:NNNNN
CAGGCGCCCA
+
][aaX__aa[
@
```

kdmurray91:
So I gather this is a bug @ctb? If so, would you like me to dig further and/or contact seqan?

mr-c:
Further digging is welcome! Looks like a bug with SeqAn 1.4.x. It'll be interesting to see if it also exists in 2.x.

mr-c:
@ctb Shall we add a check for FASTQ files that have missing quality records and throw our own exception?

- [ ] Add to known issues list

ctb:
On Mon, Mar 30, 2015 at 03:03:41AM -0700, Michael R. Crusoe wrote:
> @ctb Shall we add a check for FASTQ files that have missing quality records and throw our own exception?
> 
> - [ ] Add to known issues list

Sure.

kdmurray91:
OK, I'm happy to a) investiage w/ SeqAn 2, and b), raise a khmer exception when the quality is "" if it's a fastq file.

mr-c:
Thanks!

kdmurray91:
[...]

PR #897, opened 2015-04-02
Throw exception on weird seqan parsing of truncated file

kdmurray91:
Half workaround, half bugfix

In #884, I noticed that truncated fastq files cause an empty read to be emitted from the parser. This throws an exception to work around this. Seqan fixes this in 2.0.

- [x] Is it mergeable?
- [x] Did it pass the tests?
  fixing as needed.
  changes were made?

### DIFF
--- a/lib/read_parsers.cc
+++ b/lib/read_parsers.cc
@@ -58,7 +58,19 @@ void SeqAnParser::imprint_next_read(Read &the_read)
         ret = seqan::readRecord(the_read.name, the_read.sequence,
                                 the_read.quality, _private->stream);
         if (ret == 0) {
+            if (_num_reads == 0 && the_read.quality.length() != 0) {
+                _have_qualities = true;
+            }
             _num_reads++;
+        }
+        if (the_read.sequence.length() == 0) {
+            throw InvalidRead("Sequence is empty");
+        }
+        if (_have_qualities) {
+            if (the_read.sequence.length() != the_read.quality.length()) {
+                throw InvalidRead("Sequence and quality lengths differ");
+            }
+
         }
     }
     __asm__ __volatile__ ("" ::: "memory");
@@ -119,6 +131,7 @@ IParser(
         throw khmer_exception();
     }
     _num_reads = 0;
+    _have_qualities = false;
 }
 
 IParser::
--- a/lib/read_parsers.cc
+++ b/lib/read_parsers.cc
@@ -58,23 +58,22 @@ void SeqAnParser::imprint_next_read(Read &the_read)
         ret = seqan::readRecord(the_read.name, the_read.sequence,
                                 the_read.quality, _private->stream);
         if (ret == 0) {
-            if (_num_reads == 0 && the_read.quality.length() != 0) {
-                _have_qualities = true;
-            }
             _num_reads++;
-        }
-        if (the_read.sequence.length() == 0) {
-            throw InvalidRead("Sequence is empty");
-        }
-        if (_have_qualities) {
-            if (the_read.sequence.length() != the_read.quality.length()) {
-                throw InvalidRead("Sequence and quality lengths differ");
-            }
-
         }
     }
     __asm__ __volatile__ ("" ::: "memory");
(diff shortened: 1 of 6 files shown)


lib/read_parsers.cc
```diff
@@ -58,8 +58,20 @@ void SeqAnParser::imprint_next_read(Read &the_read)
         ret = seqan::readRecord(the_read.name, the_read.sequence,
                                 the_read.quality, _private->stream);
         if (ret == 0) {
+            if (_num_reads == 0 && the_read.quality.length() != 0) {
+                _have_qualities = true;
+            }
             _num_reads++;
         }
+        if (the_read.sequence.length() == 0) {
```
kdmurray91: @mr-c  should this be within the `ret == 0` block? Was unsure.

mr-c: Everything should be outside the spinlocks as much as possible. Especially throwing an exception. Looks like the `_num_reads` increment could also be moved after the spin lock is released.

lib/read_parsers.hh
```diff
@@ -22,12 +22,31 @@ namespace read_parsers
 {
 
 struct NoMoreReadsAvailable : public  khmer_exception {
+    explicit NoMoreReadsAvailable(const char *msg) :
+        khmer_exception(msg) {}
+    explicit NoMoreReadsAvailable() :
```
mr-c: "the use of explicit with a default constructor has no effect unless it has default arguments that would allow it to be used in copy-initialization context" http://en.cppreference.com/w/cpp/language/explicit

lib/read_parsers.hh
```diff
@@ -22,12 +22,31 @@ namespace read_parsers
...
+        khmer_exception("No More Reads Available In This File") {}
```
mr-c: Why The Title Caps?

lib/read_parsers.cc
kdmurray91: Cool, will do.

WRT `_num_reads`, wouldn't you then need to use `__sync_fetch_and_add`?

lib/read_parsers.hh
kdmurray91: 11:30 pm??? :smile: 

WIll fix

lib/read_parsers.hh
kdmurray91: Doh.

lib/read_parsers.cc
mr-c: Ah, there was a reason we kept that there. The rest of it should be outside the spinlock then.

kdmurray91:
jenkins retest this please you oaf

lib/read_parsers.cc
```diff
@@ -63,10 +63,29 @@ void SeqAnParser::imprint_next_read(Read &the_read)
...
     _private->seqan_spin_lock = 0;
-    if (atEnd) {
+    if (!atEnd) {
+        if (_num_reads > 0 && the_read.quality.length() != 0) {
+            _have_qualities = true;
+        }
+        if (the_read.sequence.length() == 0) {
+            // We decrement the number of reads we've got, as the last one is
+            // invlaid
+            __sync_fetch_and_sub(&_num_reads, 1);
```
mr-c: Hrmm... I'd rather we did all thread-aware operations inside the spinlocks instead of combining them with atomic ops. The check for seq and qual length should therefore go inside the spinlocks. No worse than what will happen when we switch to Seqan 2.0

kdmurray91: Sure, will do

kdmurray91:
that should be better, now everything's inside the spinlocks except the `throw` calls.

mr-c:
Jenkins, retest this please

kdmurray91:
retest this please

kdmurray91:
jenkins retest this please

kdmurray91:
jenkins, please sort yourself out and then retest this please

mr-c:
Top notch PR, thanks @kdmurray91!

PR #922, opened 2015-04-12
Test/py3

mr-c:
Raw table access has been disabled for now


mr-c:
For the new PrePartitionInfo CPython type:

  entries
     the `ob_size` field).
     `tp_str` & `tp_repr`
  in `init_khmer()`

mr-c:
the new PrePartitionInfo CPython type is in the spirit of the KSubSetPartition CPython type: just a handle to the underlying C++ object. Not creatable in Python land.

- [ ] Update CPython Type checklist for types that are just handles for C++ pointers.

luizirber:
Current status:
  - failing tests are annotated with 'failing_python2' and 'failing_python3' attributes.
  - mostly they involve unicode/bytes issues and division.

mr-c:
@luizirber Can you summarize the status of the py3 compatibility? Checkboxes for the remaining tests would be super.

I also remember you saying that there were issues found in Screed via the khmer tests and that we need to port some tests back to screed for future

luizirber:
See #978 for updated discussion (and squashed branch)

PR #933, opened 2015-04-13
Add (broken interleaved) paired-end support to sample-reads-randomly.py script. Closes #899.

### LINKED ISSUE #899: Add paired-end support to sampling script
Unless I'm mistaken, the `sample-reads-randomly.py` script does not support paired end reads. There is a workaround, in which you: 1) decouple your interleaved reads; 2) apply the script to both paired files using the same seed; and 3) re-interleave the sampled data.

Rather than forcing users to split and then recouple paired end data, it would be great if this script supported interleaved paired-end data out-of-the-box.

standage:
- [x] Is it mergeable?
- [x] Did it pass the tests?
  fixing as needed.
  changes were made?


standage:
Note: I've used the broken paired reader for paired-end support. Pairs are treated as a single unit and will either both be output, or neither. In a file containing only paired-end reads, requesting 10 reads will give you 20 records (10 record pairs). In a file containing only single-end reads, requesting 10 reads will give you 10 records. In a broken interleaved file, what you get will depend on the sampling.

standage:
Ready for code review.

standage:
Closes #899.

standage:
Poke @mr-c

mr-c:
Punting to @ctb

scripts/sample-reads-randomly.py
```diff
@@ -158,7 +162,9 @@ def main():
             output_file = open(output_filename, 'w')
 
         for record in reads[0]:
-            write_record(record, output_file)
+            write_record(record[2], output_file)
```
ctb: not critical, but for clarity might suggest

    for (record1, record2) in reads[0]: ...

scripts/sample-reads-randomly.py
```diff
@@ -127,7 +130,8 @@ def main():
     # read through all the sequences and load/resample the reservoir
     for filename in args.filenames:
         print >>sys.stderr, 'opening', filename, 'for reading'
-        for record in screed.open(filename, parse_description=False):
+        screed_iter = screed.open(filename, parse_description=False)
+        for record in broken_paired_reader(screed_iter):
```
ctb: I personally dislike treating 'record' as opaque here -- and throughout the code base, 'record' is used as a single screed record. so would suggest breaking this out into _, is_paired, record1, record2.

ctb: (or trying out a new convention that doesn't use 'record')

scripts/sample-reads-randomly.py
```diff
@@ -63,6 +63,9 @@ def get_parser():
     parser.add_argument('-S', '--samples', type=int, dest='num_samples',
                         default=1)
     parser.add_argument('-R', '--random-seed', type=int, dest='random_seed')
+    parser.add_argument('-p', '--paired', default=False, dest='paired',
```
ctb: This option isn't used at all, is it?

standage:
Correct, I have removed the option.

standage:
@ctb Agree with your comments regarding "records." I like to think of a record as a single entry in a fastq file, and this code abused that terminology. I was leaning toward unpacking the return values of **broken_paired_reader** as soon as I committed the code, but alas, travel. :(

standage:
Ready for code review @ctb.

ChangeLog
```diff
@@ -9,6 +9,13 @@
 
 2015-04-13  Daniel Standage  <daniel.standage@gmail.com>
 
+   * scripts/sample-reads-randomly.py: use broken paired reader to provide paired-
```
ctb: Please move this to the top now ;)

ctb: (and update the date)

ctb: Also, make sure lines are wrapped before 80 characters, please; go ahead and fix Elmar Bucher's 4/13 entry, too, if you can.

ctb:
LGTM outside of the changelog fixes.  One last question - what about providing a command line option to force single-ended behavior? (see force_single argument to broken_paired_reader). I don't see a reason to force paired-ends (there are other ways to do that) but the user may well want to excplitly ignore pairs when sampling, which there is no way to do currently in the code base or in this code.

standage:
I agree that forcing paired wouldn't make much sense, but ignoring pair information is a valid use case. Since it's already implemented as an option to broken_paired_reader, it should be easy to add an option to activate this mode. I'll fix the change log, add the option, and then (when I have time) add a test for the new option.

ctb:
thanks much appreciated :)

standage:
Huh. For some reason the script is reporting N+1 reads / read pairs instead of N. Checking whether I introduced a fencepost error...

standage:
Jenkins retest this please.

standage:
- [x] New `--force_single` option added, with corresponding test
- [x] Fixed changelog
- [x] Fixed pep8 errors introduced by recently merged commits

Ready for final review, @ctb.

ctb:
I think we use - in options, so --force-single. ;)

--

standage:
There is some inconsistency in this script: `--num_reads`, `--max_reads`, and `--random-seed`. I can fix this new option, but fixing the existing options would require a major version bump, would it not?

ctb:
Ok leave as is - will review tomorrow morning!

ctb:
LGTM; thanks, @standage !

PR #961, opened 2015-04-15
[Oxli] Initial oxli build 

bocajnotnef:
Initial oxli build using @ctb's API prototype found here: https://github.com/ctb/2015-experimental-graphalign/blob/445c200e662719c2bbf2f172a994ab6806d01edd/khmer_api.py

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2015-05-19  Jacob Fenton  <bocajnotnef@gmail.com>
+
+   * oxli/{__init__,khmer_api,common}.py,scripts/build-graph.py,
+   tests/test_scripts.py: added oxli module, oxlified load_graph script, tests
+   * scripts/load-graph.py: replaced with oxlified version
+   * setup.py: added oxli module and entry point
+
 2015-05-18  Sherine Awad  <sherine.awad@gmail.com>
 
    * tests/test_scripts.py: Test loading of compressed counting table
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -24,11 +24,11 @@ class ComboFormatter(argparse.ArgumentDefaultsHelpFormatter,
     pass
 
 
-def build_hash_args(descr=None, epilog=None):
+def build_hash_args(descr=None, epilog=None, parser=None):
     """Build an ArgumentParser with args for bloom filter based scripts."""
-    parser = argparse.ArgumentParser(
-        description=descr, epilog=epilog,
-        formatter_class=ComboFormatter)
+    if parser is None:
+        parser = argparse.ArgumentParser(description=descr, epilog=epilog,
+                                         formatter_class=ComboFormatter)
 
     env_ksize = os.environ.get('KHMER_KSIZE', DEFAULT_K)
     env_n_tables = os.environ.get('KHMER_N_TABLES', DEFAULT_N_TABLES)
@@ -60,9 +60,10 @@ def build_counting_args(descr=None, epilog=None):
     return parser
 
 
-def build_hashbits_args(descr=None, epilog=None):
+def build_hashbits_args(descr=None, epilog=None, parser=None):
     """Build an ArgumentParser with args for hashbits based scripts."""
-    parser = build_hash_args(descr=descr, epilog=epilog)
+
+    parser = build_hash_args(descr=descr, epilog=epilog, parser=parser)
     parser.hashtype = 'hashbits'
 
     return parser
--- a/oxli/__init__.py
+++ b/oxli/__init__.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python
+#
+# This file is a part of khmer, http://github.com/ged-lab/khmer/, and is
+# Copyright (C) Michigan State University, 2015. It is licensed under the
+# three-clause BSD license; see doc/LICENSE.txt.
+# Contact: khmer-project@idyll.org
+#
+
+"""
+Single entry point script for khmer
+"""
+
+import argparse
+import textwrap
+from khmer import khmer_args
+from oxli import build_graph
+
+
+def get_parser():
+    """
+    returns the parser object for the oxli subcommand handler
+    """
+
+    parser = argparse.ArgumentParser(
+        description='Single entry point script for khmer',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    subparsers = parser.add_subparsers()
+
+    # build-graph (formerly load-graph) parsers here
+    parser_build_graph = \
+        subparsers.add_parser('build-graph',
+                              help="Load sequences into the compressible graph"
+                              "format plus optional tagset",
+                              description="Load sequences into the "
+                              "compressible graph format plus optional tagset")
+
+    khmer_args.build_hashbits_args("Load sequences into the compressible"
+                                   "graph format plus optional tagset.",
+                                   None, parser=parser_build_graph)
+    build_graph.build_parser(parser_build_graph)
+    parser_build_graph.set_defaults(func=build_graph.main)
+
+    return parser
+
+
+def main():
+    """
+    main function; does the parsing and kicks off the subcommand
+    """
+    args = get_parser().parse_args()
+    args.func(args)
+
+if __name__ == '__main__':
+    main()
--- a/oxli/build_graph.py
+++ b/oxli/build_graph.py
@@ -1,0 +1,98 @@
+#! /usr/bin/env python2
+#
+# This file is part of khmer, http://github.com/ged-lab/khmer/, and is
+# Copyright (C) Michigan State University, 2009-2015. It is licensed under
+# the three-clause BSD license; see doc/LICENSE.txt.
+# Contact: khmer-project@idyll.org
+#
+# pylint: disable=invalid-name,missing-docstring
+"""
+Build a graph from the given sequences, save in <ptname>.
+
+% python scripts/load-graph.py <ptname> <data1> [ <data2> <...> ]
+
+Use '-h' for parameter help.
+"""
+
+import sys
+
+import khmer
+from khmer.khmer_args import (report_on_config, info, add_threading_args)
+from khmer.kfile import check_input_files, check_space
+from khmer.kfile import check_space_for_hashtable
+from oxli import functions
+
+
+def build_parser(parser):
+    add_threading_args(parser)
+    parser.add_argument('--no-build-tagset', '-n', default=False,
+                        action='store_true', dest='no_build_tagset',
+                        help='Do NOT construct tagset while loading sequences')
+    parser.add_argument('output_filename',
+                        metavar='output_presence_table_filename', help='output'
+                        ' k-mer presence table filename.')
+    parser.add_argument('input_filenames', metavar='input_sequence_filename',
+                        nargs='+', help='input FAST[AQ] sequence filename')
+    parser.add_argument('-f', '--force', default=False, action='store_true',
+                        help='Overwrite output file if it exists')
+    return parser
+
+
+def main(args):
+    info('build-graph.py', ['graph', 'SeqAn'])
+
+    report_on_config(args, hashtype='hashbits')
+    base = args.output_filename
+    filenames = args.input_filenames
+
+    for fname in args.input_filenames:
+        check_input_files(fname, args.force)
+
+    check_space(args.input_filenames, args.force)
+    check_space_for_hashtable(
+        (float(args.n_tables * args.min_tablesize) / 8.), args.force)
+
+    print >>sys.stderr, 'Saving k-mer presence table to %s' % base
+    print >>sys.stderr, 'Loading kmers from sequences in %s' % repr(filenames)
+    if args.no_build_tagset:
+        print >>sys.stderr, 'We WILL NOT build the tagset.'
+    else:
+        print >>sys.stderr, 'We WILL build the tagset', \
+                            ' (for partitioning/traversal).'
+
+    print >>sys.stderr, 'making k-mer presence table'
+    htable = khmer.new_hashbits(args.ksize, args.min_tablesize, args.n_tables)
+
+    functions.build_graph(filenames, htable, args.threads,
+                          not args.no_build_tagset)
+
+    print >> sys.stderr, 'Total number of unique k-mers: {0}'.format(
+        htable.n_unique_kmers())
+
+    print >>sys.stderr, 'saving k-mer presence table in', base + '.pt'
+    htable.save(base + '.pt')
+
+    if not args.no_build_tagset:
+        print >>sys.stderr, 'saving tagset in', base + '.tagset'
+        htable.save_tagset(base + '.tagset')
+
+    info_fp = open(base + '.info', 'w')
+    info_fp.write('%d unique k-mers' % htable.n_unique_kmers())
+
+    fp_rate = \
+        khmer.calc_expected_collisions(htable, args.force, max_false_pos=.15)
+    # 0.18 is ACTUAL MAX. Do not change.
+
+    print >>sys.stderr, 'false positive rate estimated to be %1.3f' % fp_rate
+    print >>info_fp, '\nfalse positive rate estimated to be %1.3f' % fp_rate
+
+    print >> sys.stderr, 'wrote to', base + '.info and', base + '.pt'
+    if not args.no_build_tagset:
+        print >> sys.stderr, 'and ' + base + '.tagset'
+
+    sys.exit(0)
+
+if __name__ == '__main__':
+    main(None)
+
+# vim: set ft=python ts=4 sts=4 sw=4 et tw=79:
--- a/scripts/load-graph.py
+++ b/scripts/load-graph.py
@@ -22,101 +22,20 @@ from khmer.khmer_args import build_hashbits_args
 from khmer.khmer_args import (report_on_config, info, add_threading_args)
 from khmer.kfile import check_input_files, check_space
 from khmer.kfile import check_space_for_hashtable
+from oxli import build_graph
 
 
 def get_parser():
     parser = build_hashbits_args(descr="Load sequences into the compressible "
-                                 "graph format plus optional tagset.")
-    add_threading_args(parser)
-    parser.add_argument('--no-build-tagset', '-n', default=False,
-                        action='store_true', dest='no_build_tagset',
-                        help='Do NOT construct tagset while loading sequences')
-    parser.add_argument('output_filename',
-                        metavar='output_presence_table_filename', help='output'
-                        ' k-mer presence table filename.')
-    parser.add_argument('input_filenames', metavar='input_sequence_filename',
-                        nargs='+', help='input FAST[AQ] sequence filename')
-    parser.add_argument('--report-total-kmers', '-t', action='store_true',
-                        help="Prints the total number of k-mers to stderr")
-    parser.add_argument('--write-fp-rate', '-w', action='store_true',
-                        help="Write false positive rate into .info file")
-    parser.add_argument('-f', '--force', default=False, action='store_true',
-                        help='Overwrite output file if it exists')
+                                       "graph format plus optional tagset.")
+
+    parser = build_graph.build_parser(parser)
     return parser
 
-
-def main():
-    info('load-graph.py', ['graph', 'SeqAn'])
-    args = get_parser().parse_args()
-    report_on_config(args, hashtype='hashbits')
-
-    base = args.output_filename
-    filenames = args.input_filenames
-
-    for _ in args.input_filenames:
-        check_input_files(_, args.force)
-
-    check_space(args.input_filenames, args.force)
-    check_space_for_hashtable(
-        (float(args.n_tables * args.min_tablesize) / 8.), args.force)
-
-    print >>sys.stderr, 'Saving k-mer presence table to %s' % base
-    print >>sys.stderr, 'Loading kmers from sequences in %s' % repr(filenames)
-    if args.no_build_tagset:
-        print >>sys.stderr, 'We WILL NOT build the tagset.'
-    else:
-        print >>sys.stderr, 'We WILL build the tagset', \
-                            ' (for partitioning/traversal).'
-
-    print >>sys.stderr, 'making k-mer presence table'
-    htable = khmer.new_hashbits(args.ksize, args.min_tablesize, args.n_tables)
-
-    if args.no_build_tagset:
-        target_method = htable.consume_fasta_with_reads_parser
-    else:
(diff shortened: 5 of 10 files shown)


bocajnotnef:
Working off of https://image.slidesharecdn.com/2015-vancouver-vanbug-150312150935-conversion-gate01/95/2015-vancouvervanbug-41-638.jpg?cb=1426209146 as a guide(ish)

bocajnotnef:
retest this please

bocajnotnef:
Working in Normalize-by-median, found this:

```
 88         passed_filter = False                                                   
 89         passed_length = True                                                    
 90         for record in batch:                                                    
 91             if len(record.sequence) < ksize:                                    
 92                 passed_length = False                                           
 93                 continue                                                        
 94                                                                                 
 95             seq = record.sequence.replace('N', 'A')                             
 96             med, _, _ = htable.get_median_count(seq)                            
 97                                                                                 
 98             if med < desired_coverage:                                          
 99                 htable.consume(seq)                                             
100                 passed_filter = True                                            
101                                                                                 
102         # Emit records if any passed                                            
103         if passed_length and passed_filter:                                     
104             for record in batch:                                                
105                 write_record(record, outfp)                                     
106         else:                                                                   
107             discarded += batch_size 
```

I'm confused by the comment on 102--seems to indicate if *any* of the records pass the filters then *all* the records in the batch are shoved into the outfp, but reading through the code seems to imply it's actually "if ANY pass the coverage filter and if ALL pass the length filter"

Am I reading this right?

Also, supposing we're doing paired reads, do we output each 'batch' seperately?

@mr-c @ctb

mr-c:
Here batch refers to either a pair or single read (See https://github.com/ged-lab/khmer/blob/2012-paper-diginorm/scripts/normalize-by-median.py#L48)

I agree that reading this block of code without that context is quite confusing which speaks to how messy this function has gotten :-/

tests/test_scripts.py
```diff
@@ -2207,15 +2237,15 @@ def test_sample_reads_randomly():
 
     seqs = set([r.name for r in screed.open(outfile)])
     print seqs
-    assert seqs == set(['850:2:1:1849:7272/1',  '850:2:1:1849:7272/2',
-                        '850:2:1:1574:8683/1',  '850:2:1:1574:8683/2',
+    assert seqs == set(['850:2:1:1849:7272/1', '850:2:1:1849:7272/2',
```
bocajnotnef: pep8 changes, here

bocajnotnef:
restest this, please

bocajnotnef:
@mr-c CR?

bocajnotnef:
@camillescott @luizirber CR, please?

bocajnotnef:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

bocajnotnef:
@camillescott @luizirber CR, please?

ChangeLog
```diff
@@ -18,6 +23,7 @@
    sample-reads-randomly.py.
    * tests/test_scripts.py: updated tests to correspond with correct
    behavior of sample-reads-randomly.py.
+>>>>>>> 97b897f3ce4998da7c97111a7f82ea9f60c3c3b4
```
luizirber: Oops.

oxli/__init__.py
```diff
@@ -0,0 +1,6 @@
+#                                                                               
+# This file is part of khmer, http://github.com/ged-lab/khmer/, and is          
+# Copyright (C) Michigan State University, 2010-2015. It is licensed under
```
luizirber: Just "2015" for the copyright?

oxli/khmer_api.py
```diff
@@ -0,0 +1,196 @@
...
+    for n, is_pair, read1, read2 in input_stream:
+        if is_pair:
+            write_record_pair(read1, read2, out_fp)
+        else:
+            write_record(read1, out_fp)
+
+
+def build_graph(ifilenames, graph, num_threads = 1, tags = False):
+    
+    if(tags):
```
luizirber: No need for parenthesis.

ChangeLog
bocajnotnef: Gaaah. Every time!

oxli/khmer_api.py
```diff
@@ -0,0 +1,196 @@
...
+        eat = graph.consume_fasta_and_tag_with_reads_parser
+    else:
+        eat = graph.consume_fasta_with_reads_parser
+    
+    for _, ifile in enumerate(ifilenames):
+        rparser = khmer.ReadParser(ifile)
+        threads = []
+    
+        for num in xrange(num_threads):
+            cur_thread = threading.Thread( target=eat, args=(rparser,))
```
luizirber: Remove whitespace before "target"

oxli/khmer_api.py
```diff
@@ -0,0 +1,196 @@
...
+            threads.append(cur_thread)
+            cur_thread.start()
+
+        for thread in threads:
+            thread.join()
+        
+
+def diginorm(input_stream, ct, coverage, ksize=0, ofp = False):
```
luizirber: Remove whitespaces around "="

oxli/khmer_api.py
```diff
@@ -0,0 +1,196 @@
...
+    n = 0
+    discard = 0
+    passed = False
+    for _, is_pair, read1, read2 in input_stream:
+        if is_pair:
+            med1, _, _ = ct.get_median_count(read1.sequence)
+            med2, _, _ = ct.get_median_count(read2.sequence)
+
+            if (med1 < coverage and len(read1.sequence) > ksize) \
+                    or (med2 < coverage and len(read2.sequence) > ksize):
```
luizirber: Line continuations can be avoided:
```python
if ((med1 < coverage and len(read1.sequence) > ksize) 
    or (med2 < coverage and len(read2.sequence) > ksize)):
```

oxli/khmer_api.py
```diff
@@ -0,0 +1,196 @@
...
+        else:
+            med, _, _ = ct.get_median_count(read1.sequence)
+            if med >= normalize_coverage:
+                read1 = _trim_record(read1, ct, trusted_coverage)
+            yield n, False, read1, None
+            n += 1
+
+
+def streamtrim(input_stream, ct, normalize_coverage, trusted_coverage):
+    import khmer.utils
```
luizirber: Why not import `broken_paired_reader` on line https://github.com/ged-lab/khmer/pull/961/files#diff-0fe57386cad3c94ff54bb7daece9b655R12 ?

bocajnotnef: to replace the `khmer.utils` import?

luizirber: and also https://github.com/ged-lab/khmer/pull/961/files#diff-0fe57386cad3c94ff54bb7daece9b655R174 and https://github.com/ged-lab/khmer/pull/961/files#diff-0fe57386cad3c94ff54bb7daece9b655R175.

Putting all imports in the top of the file is a good practice, also easier to see what are the dependencies.

bocajnotnef:
@luizirber Updated

ctb:
@mr-c and I talked about this today, and came to the following tentative conclusions:

* oxli package itself can/should be created for 2.0;
* you should create the 'oxli' entry point command 'oxli build-graph' to run this code;
* you can also replace 'load-graph.py' with this code, but please don't create a new entry point;
* the unused bits khmer_api.py should be removed;

@bocajnotnef any thoughts?

bocajnotnef:
For the oxli entry point,  I assume you mean being able to run build-graph from the command line?

Not sure what other entry point I would be making (i.e., not sure what I'm avoiding)

ctb:
On Mon, May 18, 2015 at 01:06:12PM -0700, Jake Fenton wrote:
> For the oxli entry point,  I assume you mean being able to run build-graph from the command line?
> 
> Not sure what other entry point I would be making (i.e., not sure what I'm avoiding)

% oxli build-graph

should be how you call build-graph - the subcommand format.

bocajnotnef:
retest this please

bocajnotnef:
retest this please

bocajnotnef:
retest this please

bocajnotnef:
Note: code coverage tests don't include the `oxli/` directory.

mr-c:
You should fix that :-)

bocajnotnef:
@camillescott @luizirber @mr-c CR, please?

khmer/_version.py
```diff
@@ -114,7 +114,7 @@ def git_versions_from_keywords(keywords, tag_prefix, verbose=False):
         # "stabilization", as well as "HEAD" and "master".
         tags = set([r for r in refs if re.search(r'\d', r)])
         if verbose:
-            print("discarding '%s', no digits" % ",".join(refs-tags))
+            print("discarding '%s', no digits" % ",".join(refs - tags))
```
ctb: I think these modifications can be discarded. Frankly, you should stop running autopep :)

oxli/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+    parser = argparse.ArgumentParser(
+        description='Single entry point script for khmer',
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter)
+
+    subparsers = parser.add_subparsers()
+
+    # build-graph (formerly load-graph) parsers here
+    parser_build_graph = \
+        subparsers.add_parser('build_graph',
```
ctb: let's make this build-graph (- instead of _)

oxli/common.py
```diff
@@ -0,0 +1,172 @@
+#
```
ctb: Why not just import these from khmer for now, instead of duplicating the code?

scripts/build-graph.py
```diff
@@ -0,0 +1,39 @@
+#! /usr/bin/env python2
```
ctb: Please rename this file to load-graph.py

scripts/build-graph.py
```diff
@@ -0,0 +1,39 @@
...
+import screed
+from khmer.khmer_args import build_hashbits_args
+from khmer.khmer_args import (report_on_config, info, add_threading_args)
+from khmer.kfile import check_input_files, check_space
+from khmer.kfile import check_space_for_hashtable
+from oxli import khmer_api, build_graph
+
+
+def get_parser():
+    parser = build_hashbits_args(descr="Load-Graph called.")
```
ctb: descr?

scripts/normalize-by-median.py
```diff
@@ -30,6 +30,7 @@
 from khmer.kfile import (check_space, check_space_for_hashtable,
                          check_valid_file_exists)
 from khmer.utils import write_record, check_is_pair
+
```
ctb: please don't introduce unneeded /unnecessary changes.

tests/test_scripts.py
```diff
@@ -1016,6 +1016,66 @@ def test_load_graph():
     assert x == (1, 0), x
 
 
+def test_build_graph():
```
ctb: most of these tests are copies of the tests that are being used on load-graph, yes? in which case they can be removed.

ctb:
Done. good stuff, but still things to be cleaned up! let me know if you want rationales/reasons/explanations/more info on anything.

oxli/common.py
bocajnotnef: I had to change a few of the `build_parser` funcs to add to a parser instead of generate one (because of how subparsers work).

I could modify the khmer ones to be along the line of
```
build_hashbits(parser=None):
   if parser == None:

```

tests/test_scripts.py
bocajnotnef: Indeed, especially if I'm removing `scripts/build-graph.py`. Or moving it back into `load-graph`, which amounts to the same thing.

bocajnotnef:
retest this please

bocajnotnef:
@mr-c I'm getting a strange test failure here--the linux host is dying on the sphinx docgen step but I can't reproduce the problem on Athyra.

mr-c:
You need the unreleased version of sphinx-contrib/autoprogram. Look at the
Jenkins build script and doc/requirements.txt

bocajnotnef:
I'm having trouble determining what's causing the issue--Everything was working fine until the last couple of commits but I don't see what I changed that could cause this issue. From looking at the error it looks like sphinx is dying on the `setup()` line in setup.py but I didn't edit that in these commits.

bocajnotnef:
Figured out it's because the versioneer cmdclass returns a list when a buffer is expected, but still not getting why this is just now a problem.

Also not sure which cmd class to give it.

bocajnotnef:
Not the versioneer cmdclass--No idea what it is now. Looks to be failing on `**SETUP_METADATA` but the only changes made to that were done by Luiz and were merged in by 3bb86af--which passed.

scripts/load-graph.py
```diff
@@ -18,105 +18,22 @@
...
-                        ' k-mer presence table filename.')
-    parser.add_argument('input_filenames', metavar='input_sequence_filename',
-                        nargs='+', help='input FAST[AQ] sequence filename')
-    parser.add_argument('--report-total-kmers', '-t', action='store_true',
-                        help="Prints the total number of k-mers to stderr")
-    parser.add_argument('--write-fp-rate', '-w', action='store_true',
-                        help="Write false positive rate into .info file")
-    parser.add_argument('-f', '--force', default=False, action='store_true',
-                        help='Overwrite output file if it exists')
+    parser = build_hashbits_args()
```
mr-c: Why are you no longer passing in the `descr`?

bocajnotnef:
retest this please

bocajnotnef:
@mr-c @ctb @camillescott @luizirber Updated; CR, please?

khmer/khmer_args.py
```diff
@@ -60,9 +60,12 @@ def build_counting_args(descr=None, epilog=None):
...
 
 
-def build_hashbits_args(descr=None, epilog=None):
+def build_hashbits_args(descr=None, epilog=None, parser=None):
     """Build an ArgumentParser with args for hashbits based scripts."""
-    parser = build_hash_args(descr=descr, epilog=epilog)
+    if parser is None:
+        parser = build_hash_args(descr=descr, epilog=epilog)
+    else:
+        parser = build_hash_args(descr=descr, epilog=epilog, parser=parser)
```
luizirber: since `build_hash_args` has the same default for `parser`, you can just pass it and avoid checking if it is None.

oxli/build_graph.py
```diff
@@ -0,0 +1,103 @@
...
+# This file is part of khmer, http://github.com/ged-lab/khmer/, and is
+# Copyright (C) Michigan State University, 2009-2015. It is licensed under
+# the three-clause BSD license; see doc/LICENSE.txt.
+# Contact: khmer-project@idyll.org
+#
+# pylint: disable=invalid-name,missing-docstring
+"""
+Build a graph from the given sequences, save in <ptname>.
+
+% python scripts/load-graph.py <ptname> <data1> [ <data2> <...> ]
```
luizirber: This isn't true anymore

oxli/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    khmer_args.build_hashbits_args("Load sequences into the compressible"
+                                   "graph format plus optional tagset.",
+                                   None, parser=parser_build_graph)
+    build_graph.build_parser(parser_build_graph)
+    parser_build_graph.set_defaults(func=execute_build_graph)
+
+    return parser
+
+
+def execute_build_graph(args):
```
mr-c: Why does this function exist?

oxli/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+    build_graph.main(args)
+
+
+def main():
+    """
+    main function; does the parsing and kicks off the subcommand
+    """
+    args = get_parser().parse_args()
+    args.func(args)
+    return 0
```
ctb: is the return value ever used?

oxli/build_graph.py
bocajnotnef: No, that still works--@ctb had me remove `build-graph.py` so that will work.
Granted, the oxli command also works but I don't think we're officially moving to that yet.

oxli/khmer_api.py
```diff
@@ -0,0 +1,31 @@
+#
```
ctb: Suggest renaming this from 'khmer_api' to something else. Not sure what - maybe just 'api' or 'functions'?

scripts/load-graph.py
```diff
@@ -2,7 +2,7 @@
 #
 # This file is part of khmer, https://github.com/dib-lab/khmer/, and is
 # Copyright (C) Michigan State University, 2009-2015. It is licensed under
-# the three-clause BSD license; see LICENSE.
+# the three-clause BSD license; see doc/LICENSE.txt.
```
ctb: FIXME

oxli/khmer_api.py
```diff
@@ -0,0 +1,31 @@
...
+# Contact: khmer-project@idyll.org
+#
+# pylint: disable=missing-docstring
+
+
+import threading
+import khmer.utils
+
+
+def build_graph(ifilenames, graph, num_threads=1, tags=False):
```
mr-c: What is the point of this?

ctb:
LGTM in concept!

oxli/khmer_api.py
mr-c: Nevermind, I found the call in `oxli/build_graph.py`

oxli/__init__.py
bocajnotnef: Actually yes. There's a strange issue @luizirber helped me troubleshoot where if we don't explicitly return something the testing framework grabs `None` as the status code and dies.

oxli/khmer_api.py
mr-c: naming inputs with a 'i' prefix is not a convention we use. Also needs docstrings.

bocajnotnef: The point of what? That is build_graph, now.

oxli/__init__.py
luizirber: I suggested this, because the way we run scripts for tests. Full discussion on the `ged-lab/dev-flood` gitter channel, but:

What I think it’s happening here is that we don’t have a ‘pure’ shell running the script, so when we run from the shell the None returned is converted to 0
but in our runscript() it’s actually inside Python
and the return code just stays None

luizirber: Super old post by Guido: https://www.artima.com/weblogs/viewpost.jsp?thread=4829
Recent one, good tips: http://www.jeffknupp.com/blog/2014/02/04/starting-a-python-project-the-right-way/

We can remove the `return 0`, but we might want to replace the `main()` call for `sys.exit(main())`

ctb: > We can remove the `return 0`, but we might want to replace the `main()` call for `sys.exit(main())`

+1

bocajnotnef:
@luizirber @ctb Cleaned up--Ready for merge?

oxli/build_graph.py
ctb: correct.

oxli/build_graph.py
```diff
@@ -0,0 +1,103 @@
...
+                        action='store_true', dest='no_build_tagset',
+                        help='Do NOT construct tagset while loading sequences')
+    parser.add_argument('output_filename',
+                        metavar='output_presence_table_filename', help='output'
+                        ' k-mer presence table filename.')
+    parser.add_argument('input_filenames', metavar='input_sequence_filename',
+                        nargs='+', help='input FAST[AQ] sequence filename')
+    parser.add_argument('--report-total-kmers', '-t', action='store_true',
+                        help="Prints the total number of k-mers to stderr")
+    parser.add_argument('--write-fp-rate', '-w', action='store_true',
```
ctb: turn on by default ref #1011

oxli/build_graph.py
```diff
@@ -0,0 +1,103 @@
...
+    add_threading_args(parser)
+    parser.add_argument('--no-build-tagset', '-n', default=False,
+                        action='store_true', dest='no_build_tagset',
+                        help='Do NOT construct tagset while loading sequences')
+    parser.add_argument('output_filename',
+                        metavar='output_presence_table_filename', help='output'
+                        ' k-mer presence table filename.')
+    parser.add_argument('input_filenames', metavar='input_sequence_filename',
+                        nargs='+', help='input FAST[AQ] sequence filename')
+    parser.add_argument('--report-total-kmers', '-t', action='store_true',
```
ctb: turn on by default ref #1011

ctb: (by which I mean "remove as option" because I don't imagine anyone will not want to know this)

oxli/build_graph.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def main(args):
+    info('build-graph.py', ['graph', 'SeqAn'])
+
+    report_on_config(args, hashtype='hashbits')
+    base = args.output_filename
+    filenames = args.input_filenames
+
+    for _ in args.input_filenames:
```
ctb: replace '_' with a named variable.

scripts/load-graph.py
```diff
@@ -22,101 +22,19 @@
...
-    if not args.no_build_tagset:
-        print >> sys.stderr, 'and ' + base + '.tagset'
+    parser = build_graph.build_parser(parser)
+    return parser
 
 if __name__ == '__main__':
-    main()
+    parser = get_parser()
+    args = parser.parse_args()
+    build_graph.main(args)
```
ctb: sys.exit(...)

ctb:
Before merge, this is a good candidate for squashing, ref #1013.

oxli/build_graph.py
bocajnotnef: turn on by default as in always on with no way to disable, or flip flop the arguments so the arg disables reporting?

Can't imagine why you'd want to turn that off, though...

mr-c: No way to disable

oxli/__init__.py
bocajnotnef: So, am I removing the return? If so, what am I adding in?

luizirber: `>>> sys.exit(None)`
will have a return code 0 for the shell, so you can just remove it.

oxli/__init__.py
```diff
@@ -0,0 +1,56 @@
...
+
+if __name__ == '__main__':
+    main()
```
luizirber: here you do
```sys.exit(main())```

bocajnotnef:
retest this please

bocajnotnef:
retest this please

bocajnotnef:
@ctb @luizirber Updated; CR?

ctb:
LGTM; thanks!

ctb:
@bocajnotnef or others - please merge if the build succeeds (which it should - I just fixed the ChangeLog and updated to latest master).

PR #971, opened 2015-04-15
fixing docs to remind folks to make sure tests fail for the right reasons

susinmotion:
in docstrings for khmer_tst_utils.py and in a-quick-guide-to-testing


mr-c:
Jenkins, ok to test

PR #978, opened 2015-04-21
Python 3/2 compatibility

luizirber:
Squashed version of #922

Currently all tests are passing (also some previously known_failing ones).


Makefile
```diff
@@ -36,7 +36,8 @@ help: Makefile
 install-dep: install-dependencies
 
 install-dependencies:
-	pip2 install --upgrade $(DEVPKGS) || pip install --upgrade $(DEVPKGS)
+	pip install --upgrade $(DEVPKGS)
+	pip install git+https://github.com/ged-lab/screed.git@py3
```
luizirber: This will go away when the py3 branch is merged into screed master and released.

khmer/_khmer.cc
```diff
@@ -145,7 +158,7 @@ static
 PyObject *
 Read_get_name(khmer_Read_Object * obj, void * closure )
 {
-    return PyBytes_FromString(obj->read->name.c_str()) ;
+    return PyUnicode_FromString(obj->read->name.c_str()) ;
```
luizirber: This is a design decision: All strings passed to Python will be Unicode, but we accept either Unicode or Bytes as parameters for all exposed C++ methods.

khmer/_khmer.cc
```diff
@@ -703,8 +742,13 @@ hash_get_raw_tables(khmer_KCountingHash_Object * self, PyObject * args)
...
     for (unsigned int i=0; i<sizes.size(); ++i) {
-        PyObject * buf = PyBuffer_FromMemory(table_ptrs[i], sizes[i]);
-        if(!PyBuffer_Check(buf)) {
+        Py_buffer buffer;
+        int res = PyBuffer_FillInfo(&buffer, NULL, table_ptrs[i], sizes[i], 0, PyBUF_FULL_RO);
+        if (res == -1) {
+            return NULL;
+        }
+        PyObject * buf = PyMemoryView_FromBuffer(&buffer);
+        if(!PyMemoryView_Check(buf)) {
```
luizirber: This got more complicated, but now we pass a MemoryView directly instead of a Buffer. Since the use cases involved converting it into a MemoryView anyway it shouldn't be a problem (and numpy supports it too using the np.asarray method).

khmer/kfile.py
```diff
@@ -22,16 +24,16 @@ def check_input_files(file_path, force):
     """
     mode = None
 
-    if file_path is '-':
+    if file_path == '-':
```
luizirber: wow, super subtle bug. A description of why this is wrong: http://stackoverflow.com/questions/2988017/string-comparison-in-python-is-vs
(I had to pdb and step to find it).

khmer/khmer_args.py
```diff
@@ -19,6 +22,18 @@
 DEFAULT_N_THREADS = 1
 
 
+class _VersionStdErrAction(_VersionAction):
```
luizirber: Sigh. Python 3.3 changed the default action to print version to stdout instead of stderr, since we keep the old behavior this class is needed.

khmer/khmer_args.py
```diff
@@ -196,7 +211,11 @@ def info(scriptname, algorithm_list=None):
 
     for alg in algorithm_list:
         sys.stderr.write("||   * ")
-        sys.stderr.write(_algorithms[alg])
+        algstr = _algorithms[alg].encode('utf-8', 'surrogateescape').decode('utf-8', 'replace')
+        try:
+            sys.stderr.write(algstr)
+        except UnicodeEncodeError:
+            sys.stderr.write(algstr.encode(sys.getfilesystemencoding(), 'replace'))
```
luizirber: Weird, but check http://lucumr.pocoo.org/2014/5/12/everything-about-unicode/ to see why this is needed...

setup.py
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python2
+#! /usr/bin/env python
```
luizirber: We need to set all shebangs back to `#! /usr/bin/env python`
Almost all scripts (either on sandbox/ or scripts/) need to be changed.

setup.py
```diff
@@ -210,9 +210,8 @@ def run(self):
                                          SETUP_METADATA["packages"])
 
         if "z" not in self.libraries:
-            zcmd = ['bash', '-c', 'cd ' + ZLIBDIR + ' && ( test Makefile -nt'
-                    ' configure || bash ./configure --static ) && make -f '
-                    'Makefile.pic PIC']
+            zcmd = ['bash', '-c', 'cd ' + ZLIBDIR + ' &&  bash ./configure '
+                    '--static && make -f Makefile.pic PIC']
```
luizirber: This was giving me problems, probably should be removed. Or is there any problem in regenerating the Makefile?

tests/test_counting_hash.py
```diff
@@ -107,27 +109,18 @@ def test_get_raw_tables():
...
-            memv = memoryview(tab)
-        except TypeError:
-            raise nose.SkipTest("This test needs a higher version of Python.")
-        assert sum(memv.tolist()) == 0
+        assert sum(tab.tolist()) == 0
     ht.consume('AAAATTTTCCCCGGGGAAAA')
     for tab in tables:
-        memv = memoryview(tab)
-        assert sum(memv.tolist()) == 1
+        assert sum(tab.tolist()) == 1
```
luizirber: See comment https://github.com/ged-lab/khmer/pull/978/files#r28821307

tests/test_scripts.py
```diff
@@ -2175,12 +2176,21 @@ def test_sample_reads_randomly():
...
+                      '850:2:1:2399:20086/2', '850:2:1:2503:4494/2',
+                      '850:2:1:2084:17145/1', '850:2:1:2273:13309/1',
+                      '850:2:1:2263:11143/2', '850:2:1:1984:7162/2',
+                      '850:2:1:2065:16816/1', '850:2:1:1792:15774/2'])
+    else:
+        answer = set(['850:2:1:1946:20852/1', '850:2:1:1251:16575/1',
+                      '850:2:1:1625:9364/2', '850:2:1:3123:15968/1',
+                      '850:2:1:1601:18498/2', '850:2:1:1267:6790/1',
+                      '850:2:1:2562:16360/2', '850:2:1:1199:4197/1',
+                      '850:2:1:1832:14607/1', '850:2:1:2401:4896/1'])
```
luizirber: This is tricky. Python 3 changed the implementation of random.randrange (which underlies the random.randint implementation), so they will give different results using the same seed on Python 2 and 3. If we want to remain consistent we need to grab the old implementation and leave it somewhere in our source.

tests/test_scripts.py
```diff
@@ -2369,7 +2412,10 @@ def test_count_overlap_invalid_datafile():
     args = ['--ksize', '20', '--n_tables', '2', '--min-tablesize', '10000000',
             htfile + '.pt', htfile + '.pt', outfile]
     (status, out, err) = utils.runscript(script, args, in_dir, fail_ok=True)
-    assert "IOError" in err
+    if sys.version_info.major == 2:
+        assert "IOError" in err
+    else:
+        assert "OSError" in err
```
luizirber: `PyExc_IOError` will be converted to an `OSError` on Python 3.3+. (Different exceptions were merged into OSError, see https://docs.python.org/3/library/exceptions.html#OSError)

khmer/khmer_args.py
mr-c: Sadness, thank you for fixing this.

khmer/khmer_args.py
mr-c: Wat. Thanks for tracking this down.

setup.py
mr-c: Oi

setup.py
mr-c: configure should be regenerating the Makefile; what problems are you getting?

tests/test_subset_graph.py
```diff
@@ -205,8 +207,7 @@ def test_save_load_merge(self):
         assert total_reads == 3, total_reads
 
         divvy = ht.divide_tags_into_subsets(1)
-        print divvy
-        assert len(divvy) is 3
```
mr-c: What's wrong with this assert? It prevents the test from erroring on line 211

mr-c:
@luizirber The tests pass for me on Athyra in Py3.4

mr-c:
Jenkins, test this please

mr-c:
And in py2.7

mr-c:
Weird, it is like Jenkins is one commit behind on git.

mr-c:
Jenkins, test this please

mr-c:
Hey hey hey, the tests pass!

- [ ] Get screed Py3+2 PR reviewed and merged: also test that PR with regular khmer
- [ ] Run acceptance test and BaTLab with this PR and new screed with all available Python versions

khmer/_khmer.cc
```diff
@@ -3131,16 +3130,6 @@ static PyObject* khmer_hashbits_new(PyTypeObject * type, PyObject * args,
...
-// there are no attributes that we need at this time, so we'll just return 0
-static int khmer_hashbits_init(khmer_KHashbits_Object * self, PyObject * args,
-                               PyObject * kwds)
-{
-    if (khmer_KHashtable_Type.tp_init((PyObject *)self, args, kwds) < 0) {
-        return -1;
-    }
-    return 0;
-}
-
```
luizirber: This 'empty' __init__ was a problem for Python 3, removed until we need it.

luizirber:
## Checklist

- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

## CPython checklist

  entries
     the `ob_size` field).
     `tp_str` & `tp_repr`
  in `init_khmer()`

luizirber:
Ready for review; still unsure how to write ChangeLog entry.

Steps for merge:
 - [ ] screed release (0.9, with Python 3 support)
 - [ ] merge #1045 
 - [ ] run acceptance test for this branch
 - [ ] merge this branch

mr-c:
`*.py: summary of techniques used` ?

luizirber:
ready for review and merge @mr-c @ctb

ctb:
Note that master is no longer passing on my laptop, while this branch does :). So, let's merge this sooner rather than later!

ctb:
In particular, I see no reason to run the acceptance tests.

mr-c:
Tremendous work, congratulations @luizirber

Issue #724, opened 2015-01-12
Feature request: compatibility wrapper for Jellyfish


mr-c:
Since we have a different workflow this doesn't have to recreate the same command line options but should produce compatible output

`jellyfish count [-> jellyfish merge] -> jellyfish histo -> jellyfish dump`

```
Usage: jellyfish histo [options] db:path

Create an histogram of k-mer occurrences

Create an histogram with the number of k-mers having a given
count. In bucket 'i' are tallied the k-mers which have a count 'c'
satisfying 'low+i*inc <= c < low+(i+1)*inc'. Buckets in the output are
labeled by the low end point (low+i*inc).

The last bucket in the output behaves as a catchall: it tallies all
k-mers with a count greater or equal to the low end point of this
bucket.

Options (default value in (), *required):
 -l, --low=uint64                         Low count value of histogram (1)
 -h, --high=uint64                        High count value of histogram (10000)
 -i, --increment=uint64                   Increment value for buckets (1)
 -t, --threads=uint32                     Number of threads (1)
 -f, --full                               Full histo. Don't skip count 0. (false)
 -o, --output=string                      Output file
 -v, --verbose                            Output information (false)
 -U, --usage                              Usage
     --help                               This message
     --full-help                          Detailed help
 -V, --version                            Version
```
Jellyfish histogram files are space delimited.

```
Usage: jellyfish dump [options] db:path

Dump k-mer counts

By default, dump in a fasta format where the header is the count and
the sequence is the sequence of the k-mer. The column format is a 2
column output: k-mer count.

Options (default value in (), *required):
 -c, --column                             Column format (false)
 -t, --tab                                Tab separator (false)
 -L, --lower-count=uint64                 Don't output k-mer with count < lower-count
 -U, --upper-count=uint64                 Don't output k-mer with count > upper-count
 -o, --output=string                      Output file
     --usage                              Usage
 -h, --help                               This message
 -V, --version                            Version
```

User story: PBcR-MHAP uses Jellyfish configured to request 1 TB of memory. Luiz wants to use less memory.

http://sourceforge.net/p/wgs-assembler/svn/HEAD/tree/trunk/src/AS_PBR/PBcR.pl#l1621

[...]

PR #983, opened 2015-05-07
add scripts to output k-mers and k-mer counts

ctb:
An initial implementation of a simple script to output k-mer counts.

Per request by @drtamermansour.

### DIFF
--- a/sandbox/count-kmers-single.py
+++ b/sandbox/count-kmers-single.py
@@ -40,8 +40,6 @@ def get_parser():
                         dest='output_file',
                         type=argparse.FileType('w'),
                         default=None, help='output counts to this file')
-    parser.add_argument('--report-total-kmers', '-t', action='store_true',
-                        help="Prints the total number of k-mers to stderr")
 
     return parser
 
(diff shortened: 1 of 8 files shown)


mr-c:
Sweet. This is a great foundation for implementing #724

ctb:
- [ ] write a recipe for both scripts
- [ ] create an issue to move the final for loop(s) into a C++ function as with abundance-dist.py

ctb:
@mr-c @camillescott @luizirber please review at your leisure.

sandbox/count-kmers-single.py
```diff
@@ -0,0 +1,106 @@
...
+
+    parser.add_argument('input_sequence_filename', help='The input'
+                        ' FAST[AQ] sequence file.')
+
+    parser.add_argument('-o', '--out', metavar="output_file",
+                        dest='output_file',
+                        type=argparse.FileType('w'),
+                        default=None, help='output counts to this file')
+    parser.add_argument('--report-total-kmers', '-t', action='store_true',
+                        help="Prints the total number of k-mers to stderr")
```
mr-c: This should just happen without a flag. Flagged version was only done so as to not break the CLI API.

mr-c:
Otherwise, LGTM

sandbox/count-kmers-single.py
ctb: So, how do you want to handle this in general? Let's touch base on it today and then write up an issue.

ctb: (I take your point that this can be default for new scripts :)

mr-c:
Otherwise, LGTM

ctb:
Points in https://github.com/dib-lab/khmer/pull/983#issuecomment-100564626 punted to issues; ready for review and merge @mr-c @luizirber @camillescott

mr-c:
LGTM, done

PR #987, opened 2015-05-08
Fix behavior of filter-abund and filter-abund-single with respect to Ns.

### LINKED ISSUE #985: filter-abund.py and filter-abund-single.py discard reads with 'N's in them
Pointed out by @drtamermansour. See:

https://github.com/ged-lab/khmer/blob/master/scripts/filter-abund.py#L98
https://github.com/ged-lab/khmer/blob/master/scripts/filter-abund-single.py#L109

The appropriate behavior is to mimic normalize-by-median.py and replace 'N' with 'A', I think.

While we're at it, I think these scripts should be simplified and the threading removed.

ctb:
filter-abund.py and filter-abund-single.py no longer discard sequences with Ns in them, and instead treat Ns as As.  Fixes #985.

Also unified the rest of the codebase to treat Ns as As.


ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

ctb:
Note to @mr-c: this changes the output behavior, and may require a version increase to 2.0 upon merge.

ctb:
@mr-c @luizirber @camillescott ready for review.

@drtamermansour your comments on the approach solicited!

drtamermansour:
Two comments:
1. You keep the N's in the output only if the reads are less than the
"normalize_to" cutoff. I think you mean it and I can see why.
2. Should we change the behavior of normalize-by-median to keep the 'N's in
the output?

ctb:
I think we always keep the Ns, as long as they're not trimmed off.

I believe normalize-by-median already keeps Ns in the output.

drtamermansour:
You assume that trimming will always remove the newly changed 'A's which
should not be always true, right?

You are right. I am sorry. You use the transformed read for counting kmers
but you output the original record

ctb:
On Fri, May 08, 2015 at 09:57:41AM -0700, Tamer Mansour wrote:
> You assume that trimming will always remove the newly changed 'A's which
> should not be always true, right?

I didn't intend to assume that... if the trimming doesn't remove the
new "A" base, an N is correctly output for the filtered sequence.

drtamermansour:
Oooh ya, I see it now. This is very clever

ctb:
On Fri, May 08, 2015 at 10:09:47AM -0700, Tamer Mansour wrote:
> Oooh ya, I see it now. This is very clever

:)

ctb:
@luizirber @camillescott ready for review.  Note that this should not be merged yet, as it will bump the version number significantly @mr-c.

drtamermansour:
Can not we have a beta version for Khmer 2 were we merge such updates?

Tamer

ctb:
We'll figure something out; as @mr-c and I discussed yesterday, the plan is to
release more quickly over the upcoming months.

ctb:
@bocajnotnef @luizirber @camillescott CR please? Now that it's after 1.4 release, we could merge.

tests/test_scripts.py
```diff
@@ -435,6 +435,72 @@ def test_filter_abund_6_trim_high_abund_Z():
...
+
+    # check to see if that 'N' was properly changed to an 'A'
+    seqs = set([r.sequence for r in screed.open(outfile)])
+    assert 'GGTTGACGGGGCTCAGGGGGCGGCTGACTCCGAG' not in seqs, seqs
+
+    # ...and that an 'N' remains in the output sequences
+    found_N = False
+    for s in seqs:
+        if 'N' in s:
+            found_N = True
```
bocajnotnef: Perhaps `found_N = 'N' in s`

admittedly nitpicky

bocajnotnef:
Only one nitpick and it isn't mergable. otherwise LGTM.

tests/test_scripts.py
ctb: > Perhaps `found_N = 'N' in s`

I think this is more confusing to read than the 'if' statement, so -0 on
changing.

ctb:
Ready for review and merge @luizirber @camillescott

luizirber:
LGTM

PR #995, opened 2015-05-11
fixed boundary error in sample-reads-randomly

ctb:
This fixes a bug introduced in #933.

@standage, can you do a quick check?

@mr-c, this should probably be fixed before 1.4, if you can fit it in; or else 1.4.1 needs to be released quickly :).


standage:
This works as expected on real data, and the (minimal) changes to the code look good to me. I'm curious though as to the boundary case(s) that caused the previous code to fail.

ctb:
On Mon, May 11, 2015 at 05:11:06PM -0700, Daniel Standage wrote:
> This works as expected on real data, and the (minimal) changes to the code look good to me. I'm curious though as to the boundary case(s) that caused the previous code to fail.

The new assert is the key point - the count returned by the
broken_paired_reader can be higher than the number of tuples returned,
so the script was using an index that was sometimes out of bounds for
the number of reads actually returned.  Making the count be based on
the number of _pairs_ fixes this.

It's a failure mode that depends on the random seed used as well as
having a reasonable large number of reads, so I didn't think an explicit
test was worthwhile.  Instead, the assert!

ctb:
@luizirber @camillescott ready for review & merge.

ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

luizirber:
LGTM.

Food for thought: Testing the random outputs using exact matches is good for backwards compatibility, but how do we keep up with changes like the one in this PR or [incompatibilities between Python versions][1]?

[1]: https://github.com/ged-lab/khmer/pull/978/files#r28824170

ctb:
On Tue, May 12, 2015 at 08:41:59AM -0700, Luiz Irber wrote:
> LGTM.

k - can you merge, then?

> Food for thought: Testing the random outputs using exact matches is good for backwards compatibility, but how do we keep up with changes like the one in this PR or [incompatibilities between Python versions][1]?
> 
> [1]: https://github.com/ged-lab/khmer/pull/978/files#r28824170

no idea...

standage:
If the same seed produces different behavior in Python 2 vs 3, automated tests could simply check whether the test output matches one of two possible correct answers.

PR #1001, opened 2015-05-15
Diginorm skip kmers

### LINKED ISSUE #1000: normalize-by-median skip consuming kmers of PE reads
The normalize-by-median script in the PE mode keeps both paired ends if only one of them is below the coverage cutoff. However then script only consume the kmers from one end. This is why if we run abound-dist.py on the output k-mer counting table, 1st line will show some no of kmers with 0 frequency.

drtamermansour:
This should fix #1000 solving 2 main problems:
1) fix normalize-by-median.py to count kmers from both PE reads when either one of them is below the coverage cutoff.
2) Change the function hash_consume in khmer/_khmermodule.cc to checks for non-ACGT characters


drtamermansour:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

drtamermansour:
Ready for revision @ctb

ctb:
This looks like a good set of fixes to me, although I have to double-check on the side effects.

More critically, we need some tests for both behavior.  Happy to chat more about what those tests would look like if you want help.

drtamermansour:
done @ctb

drtamermansour:
I did a test with real data (PE and SE reads of a ~ 7GB with considerable number of reads containing 'N's). I compared the number of unique kmers calculated in the output of normalize-by-median.py versus the numbers produced after loading the *.keep files into a new hash table using load-into-counting.sh. They are exactly the same.

ctb:
In the future, could you please put your branches on gedlab? thanks :)

ChangeLog
```diff
@@ -1,3 +1,11 @@
+2015-5-15 Tamer Mansour <drtamermansour@gmail.com>
+
+   * normalize-by-median changed to count kmers from both PE reads when either
+   one of them is below the coverage cutoff
+   * The function hash_consume in khmer/_khmermodule.cc changed to checks for
```
ctb: 'check', not 'checks'

khmer/_khmermodule.cc
```diff
@@ -912,9 +912,10 @@ hash_consume(khmer_KCountingHash_Object * me, PyObject * args)
                         "string length must >= the hashtable k-mer size");
         return NULL;
     }
-
+    std::string read(long_str);
+    bool is_valid;
     unsigned int n_consumed;
-    n_consumed = counting->consume_string(long_str);
+    n_consumed = counting->check_and_process_read(read, is_valid);
```
ctb: reverting this change does not break any tests, which means it's untested. please write a test targeting the new behavior - thanks :)

ctb:
Otherwise looks good to me.

khmer/_khmermodule.cc
drtamermansour: It should not break anything. I did not fix consume_string(). I used another function instead which do the check then call consume_string. I am testing the new functionally by trying data file with 'N's characters.

ctb: On Mon, May 18, 2015 at 08:08:32AM -0700, Tamer Mansour wrote:
> >      unsigned int n_consumed;
> > -    n_consumed = counting->consume_string(long_str);
> > +    n_consumed = counting->check_and_process_read(read, is_valid);
> 
> It should not break anything. I did not fix consume_string(). I used another function instead which do the check then call consume_string. I am testing the new functionally by trying data file with 'N's characters.

Yes, but I want a test that demonstrates that something has changed about
the functionality.  This test should break when consume_string is used and
succeed when check_and_process_read is used.

drtamermansour:
@ctb  
Now it should fail if you revert back to consume_string. But I think I discovered a really bigger problem. As you can see, I did not add any new tests. I just changed the sequence of the test data a little bit. I will file a new issue describing the new problem

scripts/normalize-by-median.py
```diff
@@ -94,12 +94,12 @@ def normalize_by_median(input_filename, outfp, htable, paired, cutoff,
             med, _, _ = htable.get_median_count(seq)
 
             if med < desired_coverage:
-                htable.consume(seq)
                 passed_filter = True
 
         # Emit records if any passed
         if passed_length and passed_filter:
             for record in batch:
+                htable.consume(record.sequence)
```
ctb: OK, sorry for the delay, took some time to wrap my head around this.  This is effectively changing the behavior of diginorm to no longer convert Ns to As but instead ignore k-mers in reads with Ns in them.  Do we want to do that? I am skeptical.

What is wrong with simply replacing Ns with As as was done previously?

drtamermansour:
This pull request is changing 2 functionalities: 

The 1st is moving the consume function outside the loop to make sure that both reads of a PE will be correctly consumed if any of them did not path the cutoff. Before fixing that we would keep some reads without counting their kmers. This is why if you run abund-dist.py on the *.keep files using the diginorm counting table, we used to get some kmers with count 0 which does not make sense. 

The script used to convert 'N's to 'A's before consuming them. Now by moving the consume function outside the loop, we will change this behavior. It will pass the read with 'N's to the consume function. We discussed the behavior of this function toward reads with 'N's here #1014 . As you said it change 'N's to 'G's. The second change in this pull request is changing the behavior of the consume function itself in _khmermodule.cc to skip counting the reads with 'N's. I have many reasons which favor this change:
1) "Most importantly" for code consistency and reproducibility: Loading the the diginorm output using load-into-counting.py (which ignore reads with 'N's) will product a different counting table and different unique kmer count. The same principle apply to other scripts creating their own counting table e.g. abundance-dist-single.py and filter-abund-single.py.
2) On switching 'N's to 'A's (or 'G's), we are adding mostly erroneous kmers to our counting table which is likely to affect the downstream applications. For example, I think pumping the count of erroneous kmers is not helping abund-filter specially if the user is not using the -V mode.   
3) For modularity: If we are working on error correction module, we should reserve these decisions to it.
4) what are the drawbacks of my suggested change: a little bit less efficient but a little bit more accurate diginorm. We will retain reads with 'N's  as long as they are below the normalization cutoff (less efficient) but diginorm now is checking coverage with more trusted reads so less risk of data loss (more accurate).

ctb:
OK, I understand now.  Note that the change with respect to 'N' behavior wasn't in the original description ;)

Your fix to consume both pairs is uncontroversial and fixes a bug; I'm happy to merge it.

However, the alteration to the N behavior should be discussed in a separate issue.  I'm not necessarily against this new behavior but I want to lay out concerns.  Can you remove that change from this PR and file a new issue for that?

ctb:
@bocajnotnef could you take this over, please, unless @drtamermansour wants to follow through? I suggest closing this PR and redoing it to fix the immediate bug, since the larger behavior change is not going to be merged until we think it through (see #1036).

bocajnotnef:
@ctb Vulturing.

drtamermansour:
sigh of relief. Thank you

bocajnotnef:
@ctb To recap, I'm fixing the issue in #1000 by moving a consume outside of a loop (as @drtamermansour did) but in such a way that the script still changes N's to A's prior to consumption, yes?

ctb:
On Mon, Jun 01, 2015 at 08:54:18AM -0700, Jake Fenton wrote:
> @ctb To recap, I'm fixing the issue in #1000 by moving a consume outside of a loop (as @drtamermansour did) but in such a way that the script still changes N's to A's prior to consumption, yes?

yep.

and writing a test so it NEVER HAPPENS AGAIN.

drtamermansour:
When we move the consume function outside the loop, we are not converting 'N's to 'A's anymore. Instead we pass the reads as it is to the consume function. The current implementation of the consume function changes 'N's to 'G's.

ctb:
Closing; being handled in #1057.

Issue #1002, opened 2015-05-15
request for discussion: screed / reading from location in the middle of a file


ctb:
I would like to solicit thoughts on how, in screed, to implement reading from arbitrary locations in a file. 

Specifically, I need to be able to point a screed iterator at a given byte location in a FASTQ file, and find the first valid record after that.  My first thought is to:

1. seek to that position in the file
2. find the first likely valid record (look for '\n@'?)
3. seek to that position
4. hand file handle to fastq_iter

seem reasonable? Am I missing something?

Important caveats -
- I am OK with this being FASTQ-only for now;
- it doesn't need to be super efficient;
- A more generic solution is fine but is not time sensitive; I need something ~nowish.
- another solution we need to implement would be to read *efficiently* from multiple, randomly generated locations within the file. I would probably hack that in with the above solution by reading 1000 sequences at a time, at N different locations chosen within N equally sized segments across the file. thoughts on that also welcome. sample-reads-randomly is too inefficient (although yes, nicely random).

@mr-c @bocajnotnef @luizirber @camillescott @standage any thoughts? This is somewhat urgent :)

drj11:
Your approach sounds fine in principle.

How random do you need it? I suspect a slight bias towards smaller records.

ctb:
peanut gallery at https://twitter.com/ctitusbrown/status/599220862311469056 - @pmelsted suggests something neat that I might try.

ctb:
@drj11 good point, we can assume records are all the same size and compress ~equally (in case of gz)

lindenb:
point 2:  warning quality lines can start with '@'

jessehoff:
isn't an index helpful for this and relatively cheap?

ctb:
Very helpful clarifications requested/made on twitter:

* I want a well-bounded (~fixed) number of reads, not related to size of FASTQ as long as it's big;
* we can assume, for my purposes, that FASTQ records are all the same size and compress equally;
* reading the entire file once is too time-expensive, you all!!

ctb:
A good point that overengineering is the root of much evil - I will probably opt for something based on a fixed subset of the entire read collection for my immediate need.

ctb:
I'm not proud of it, but it should suffice for publication. And it's certainly KISS.

https://github.com/dib-lab/khmer/commit/69ebed553002d3931ec2653cdcd83c39b71db106

ctb:
[...]

PR #1003, opened 2015-05-15
introduce 1m read skip in calc-error-profile

ctb:
See #1002 ; for github.com/dib-lab/2014-streaming/.


standage:
Ready for review and merge? @ctb

Issue #1006, opened 2015-05-16
Refactor normalize-by-median with extreme prejudice


ctb:
There is mounting dissatisfaction amongst the natives with the agglomerative mess that normalize-by-median has become.

In casual conversation, @camillescott has pointed out that much of the mess comes from fail-safe attempts, and that context managers (see e.g. http://eigenhombre.com/2013/04/20/introduction-to-context-managers/) could potentially simplify this. 

I think this could also be part of addressing #988.

I'd like to see this done in stages by refactoring, with code review at each stage. @bocajnotnef you up for it? cc @drtamermansour

bocajnotnef:
Sounds fun. Normalize-by-median was on my list of things to oxlize
(oxlify?).

I would like to see https://github.com/dib-lab/khmer/pull/961 merged first,
as it contains some API stuff that could be useful.

ctb:
On Sat, May 16, 2015 at 07:41:31AM -0700, Jake Fenton wrote:
> Sounds fun. Normalize-by-median was on my list of things to oxlize
> (oxlify?).
> 
> I would like to see https://github.com/dib-lab/khmer/pull/961 merged first,
> as it contains some API stuff that could be useful.

"Refactoring" here means taking the existing code and modifying progressively,
rather than going for a clean rewrite.  The latter is more likely to lead to
unexpected and mass breakage, especially in the absence of comprehensive
tests.  I need to talk with @mr-c about #961, too; the prototype code
in khmer_api is not necessarily ready for the khmer repo, and I am not
sure we want to start putting oxli code into the codebase just yet, especially
given http://ivory.idyll.org/blog/2015-modifying-our-dev-process.html.

So -1 on predicating this on #961, +0 on moving towards code that looks
as simple as the khmer_api code.  Will revisit after talking with @mr-c.

drtamermansour:
+1 for working on code simplification while addressing #988
<https://github.com/dib-lab/khmer/issues/988>.

bocajnotnef:
So, specific pain points:

 - [ ] Lots of error checking/failsafing that could and should be done with context managers
 - [ ] Should accept and deal with broken-pairs
 - [ ] Better default arguments for enhanced understanding / less intimidating CLI appearance (in the vein of UX)

What'll the stages be?

bocajnotnef:
as per discussion with Dr Tamer:

Stage 0:
- [ ] a) proper dealing with broken-paired input
[...]

PR #957, opened 2015-04-14
Fix Issuer #58: normalize-by-median: accept paired & unpaired data at the same time

susinmotion:
fix: Fixes issue #58 Allows unpaired data to be included with paired data.
refactor: Pulls error handling for normalize into separate wrapper function.

- [x] Is it mergeable?
- [x] Did it pass the tests?
- [x] If it introduces new functionality in scripts/ is it tested?
  Check for code coverage with `make clean diff-cover`
- [x] Is it well formatted? Look at `make pep8`, `make diff_pylint_report`,
  `make cppcheck`, and `make doc` output. Use `make format` and manual
  fixing as needed.
- [x] Did it change the command-line interface? Only additions are allowed
  without a major version increment. Changing file formats also requires a
  major version number increment.
- [x] Is it documented in the ChangeLog?
  http://en.wikipedia.org/wiki/Changelog#Format
- [x] Was a spellchecker run on the source code and documentation after
  changes were made?
- [x] Is the Copyright year up to date?

PR #1010, opened 2015-05-18
Normalize-by-median refactor: Stage 0

bocajnotnef:
Initial implementation of context managers of move some boilerplate out of the way in normalize-by-median; addresses parts of #1006.

Much more to come....


scripts/normalize-by-median.py
```diff
@@ -170,7 +181,7 @@ def normalize_by_median_and_check(input_filename, htable, single_output_file,
                 .format(inp=input_filename, kept=total - discarded,
                         total=total, perc=int(100. - discarded /
                                               float(total) * 100.))
-            print >> sys.stderr, 'output in', output_name
+            print >> sys.stderr, 'output in', outfp
```
bocajnotnef: this will be reverted

bocajnotnef:
@camillescott @ctb quick once-over to make sure I'm on the right track?

My concern is that so far this is just moving code around.

ctb:
On Mon, May 18, 2015 at 08:26:31AM -0700, Jake Fenton wrote:
> @camillescott @ctb quick once-over to make sure I'm on the right track?
> 
> My concern is that so far this is just moving code around.

didn't look, but please proceed with confidence.  "just moving code around" is fine.

bocajnotnef:
Mkay. Proceeding with prejudice.

bocajnotnef:
@luizirber @camillescott @ctb CR, please?

ctb:
Looks good on a quick skim.  Please only ask for a CR after the Jenkins tests pass, though.

ctb:
p.s. "just moving code around without breaking anything" is refactoring done properly. This looks like it's already a big improvement!

bocajnotnef:
Huh. I only changed the changelog, since the last commit--Jenkins shouldn't have broken.

Looks like code coverage dropped; denominator issue. I'll go write a test. (Batchwise stuff isn't tested at all, it seems.)

ctb:
On Fri, May 22, 2015 at 08:06:16AM -0700, Jake Fenton wrote:
> Huh. I only changed the changelog, since the last commit--Jenkins shouldn't have broken.

It was a general point, not a specific one :)

scripts/normalize-by-median.py
```diff
@@ -126,6 +128,22 @@ def handle_error(error, output_name, input_name, fail_save, htable):
         print >> sys.stderr, '** ERROR: problem removing corrupt filtered file'
 
 
+@contextmanager
+def FailSafe(ifile, ofile, save_on_fail, ht, corrupted, total, dicarded, jedi):
```
luizirber: nice pun with jedi/force, but it is harder to read

bocajnotnef: It was to fit in 80 chars. I can change it, but I'll have to throw in a line break.

luizirber: That's  fine, because there is a typo in 'discarded' too

bocajnotnef:
@ctb Updated; CR?

ctb:
Not gonna get to it today.  Keep on working :)

scripts/normalize-by-median.py
```diff
@@ -270,6 +282,24 @@ def get_parser():
...
+
+        if (freq > 0 and index > 0 and index % freq == 0):
+            print 'Backup: Saving k-mer counting file through', \
+                input_filenames[index]
+            if savename:
+                hashname = savename
+                print '...saving to', hashname
+            else:
+                hashname = 'backup.ct'
+                print 'Nothing given for savetable, saving to', hashname
```
bocajnotnef: Should this be to stderr?

ctb: yep.

bocajnotnef:
I think I've refactored much of the obvious.

scripts/normalize-by-median.py
```diff
@@ -43,23 +45,13 @@ def batchwise(coll, size):
...
-    if paired:
-        batch_size = 2
-
-    index = -1
-    total = 0
-    discarded = 0
-    for index, batch in enumerate(batchwise(screed.open(
-            input_filename, parse_description=False), batch_size)):
         if index > 0 and index % 100000 == 0:
             print >>sys.stderr, '... kept {kept} of {total} or'\
```
ctb: I think this 'print' could usefully be encapsulated in a function; just a thought.

scripts/normalize-by-median.py
```diff
@@ -81,6 +73,31 @@ def normalize_by_median(input_filename, outfp, htable, paired, cutoff,
                 raise IOError('Error: Improperly interleaved pairs \
                     {b0} {b1}'.format(b0=batch[0].name, b1=batch[1].name))
 
+        yield batch
+
+        index = index + 1
```
ctb: Could this be done with an enumerate(...) around batchwise(...)?

scripts/normalize-by-median.py
```diff
@@ -126,9 +143,40 @@ def handle_error(error, output_name, input_name, fail_save, htable):
...
+
+            sys.exit(1)
+        else:
+            print >> sys.stderr, '*** Skipping error file, moving on...'
+            corrupt_files.append(ifile)
+
+    if total_acc == 0 and discarded_acc == 0:
+        print >> sys.stderr, 'SKIPPED empty file', ifile
+    elif total_acc is not None and discarded_acc is not None:
+        # will be none if the context manager caught an error
```
ctb: replace "the context manager" with which one you're talking about here.

scripts/normalize-by-median.py
```diff
@@ -126,9 +143,40 @@ def handle_error(error, output_name, input_name, fail_save, htable):
         print >> sys.stderr, '** ERROR: problem removing corrupt filtered file'
 
 
+@contextmanager
+def FailSafe(ifile, ofile, save_on_fail, ht, force):
+    global corrupt_files, total, discarded, total_acc, discarded_acc
+    try:
+        yield
```
ctb: huh. what does this do?

scripts/normalize-by-median.py
luizirber: Talking about `print`, it would be nice to move to the print function syntax:
- add `from __future__ import print_function` to the top of the script
- move `sys.stderr` to the `file` parameter:
```
print('... kept {kept} of {total} or {perc:2}%'
      .format(kept=total - discarded,
              total=total,
              perc=int(100. - discarded / float(total) * 100.)),
      file=sys.stderr)
```
This will help for the Python 3 compatibility.

scripts/normalize-by-median.py
```diff
@@ -308,24 +354,16 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
-            if args.savetable:
-                hashname = args.savetable
-                print '...saving to', hashname
-            else:
-                hashname = 'backup.ct'
-                print 'Nothing given for savetable, saving to', hashname
-            htable.save(hashname)
+                report_fp)
 
     if args.paired and args.unpaired_reads:
```
ctb: it is unclear to me what is going on here - comment?

ctb:
I really like the context managers; nice work.

Requests --

* can you rename FailSafe to something like "CatchInputErrors"?
* rename SavingTable to "CheckpointCountingTable"
* document all the context managers with docstrings

No code review needed after this; I'd just forge ahead with broken_paired_reader!

scripts/normalize-by-median.py
bocajnotnef: the `@contextmanager` decorator turns a generator into a context manager. When the block executes the interpreter replaces the `yield` with whatever's within the relevant `with` block. In this case from here: https://github.com/dib-lab/khmer/pull/1010/files#diff-6b3b058584d44911b3cf3fe21d9c9af4R197; the lines 199 & 200 are shoved in where the `yield` is and then it's executed as a 'normal' block.

If you meant the context manager it just abstracts out the IOError catching and accounting.

scripts/normalize-by-median.py
bocajnotnef: Dunno. Not my doing. Looks like it's from the broken paired/`-p` implementation, from https://github.com/dib-lab/khmer/pull/957/files#diff-6b3b058584d44911b3cf3fe21d9c9af4R325

If you meant the mass deletion of the dump frequency code, I moved that into a generator.

ctb: On Wed, May 27, 2015 at 07:55:31AM -0700, Jake Fenton wrote:
> >                  args.fail_save, args.paired, args.cutoff, args.force,
> > -                corrupt_files, report_fp)
> > -
> > -        if (args.dump_frequency > 0 and
> > -                index > 0 and index % args.dump_frequency == 0):
> > -            print 'Backup: Saving k-mer counting file through', input_filename
> > -            if args.savetable:
> > -                hashname = args.savetable
> > -                print '...saving to', hashname
> > -            else:
> > -                hashname = 'backup.ct'
> > -                print 'Nothing given for savetable, saving to', hashname
> > -            htable.save(hashname)
> > +                report_fp)
> >  
> >      if args.paired and args.unpaired_reads:
> 
> Dunno. Not my doing. Looks like it's from the broken paired/`-p` implementation, from https://github.com/dib-lab/khmer/pull/957/files#diff-6b3b058584d44911b3cf3fe21d9c9af4R325

Great, so refactor it to be clearer or comment as is :)

bocajnotnef:
retest this please.

bocajnotnef:
@ctb By 'no code review needed' does that mean go ahead and merge?

ctb:
Well, you haven't proposed it for merge, so no - and definitely would need a code review before that.  What I was suggesting was that you go ahead with this PR and add the next step.  If you want us to merge first & start a new PR for the next step, then go ahead and fill out the checklist...

bocajnotnef:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

@ctb @luizirber @camillescott CR & merge, please?

ctb:
Note that @camillescott is en route to Davis, @mr-c is out for next two weeks, and I'm not expert on context managers. So I will punt this to @luizirber who might be able to look at it Friday...

bocajnotnef:
Alright. I'm approaching my cap for the week anywho. I'll start plotting how to do broken_paired stuff and making Davis preparations.

scripts/normalize-by-median.py
```diff
@@ -43,33 +47,31 @@ def batchwise(coll, size):
...
+def WithDiagnostics(ifile, batch_size, fp, paired):
+    """
+    Generator/context manager to do boilerplate output of statistics while
+    normalizing data. Also checks for properly paired data.
+    """
 
-    desired_coverage = cutoff
-    ksize = htable.ksize()
+    index = 0
+    global total, discarded
```
ctb: I dislike these globals.  Globals in general are bad.  Among other things they make the code non-reusable elsewhere.

Could we make WithDiagnostics an object that had 'total' and 'discarded' attributes, instead? Happy to sketch that out in more detail if you like.

scripts/normalize-by-median.py
```diff
@@ -105,30 +130,67 @@ def normalize_by_median(input_filename, outfp, htable, paired, cutoff,
...
+    """
+    Context manager to do boilerplate excepting of IOErrors; also does
+    upkeep on some statistics and diagnostic output.
+    """
+
+    global corrupt_files, total, discarded, total_acc, discarded_acc
+    try:
+        yield
+    except IOError as err:
+        handle_error(err, ofile, ifile, save_on_fail,
```
ctb: unnecessary line wrap :)

scripts/normalize-by-median.py
```diff
@@ -105,30 +130,67 @@ def normalize_by_median(input_filename, outfp, htable, paired, cutoff,
...
+
+
+@contextmanager
+def CatchIOErrors(ifile, ofile, save_on_fail, ht, force):
+    """
+    Context manager to do boilerplate excepting of IOErrors; also does
+    upkeep on some statistics and diagnostic output.
+    """
+
+    global corrupt_files, total, discarded, total_acc, discarded_acc
```
ctb: Same comment as above - can we avoid making these global?

scripts/normalize-by-median.py
```diff
@@ -105,30 +130,67 @@ def normalize_by_median(input_filename, outfp, htable, paired, cutoff,
...
+
+            sys.exit(1)
+        else:
+            print('*** Skipping error file, moving on...', file=sys.stderr)
+            corrupt_files.append(ifile)
+
+    if total_acc == 0 and discarded_acc == 0:
+        print('SKIPPED empty file ' + ifile, file=sys.stderr)
+    elif total_acc is not None and discarded_acc is not None:
+        # will be none if the context manager caught an error
```
ctb: can errors be passed other ways? especially if we are making the context manager into an object? this is a bit non-intuitive/complex.

scripts/normalize-by-median.py
```diff
@@ -270,6 +311,26 @@ def get_parser():
...
 
 
+def CheckpointCountingTable(input_filenames, freq, ht, savename):
+    """
+    Generator/context manager to progressively save counting tables
+    """
+
+    for index, ifile in enumerate(input_filenames):
+        yield ifile
+        if (freq > 0 and index > 0 and index % freq == 0):
```
ctb: can remove parantheses, and also just write:

     if freq and index and index % freq == 0:

scripts/normalize-by-median.py
bocajnotnef: I'm not sure how to make it an object with generator properties--a sketch would be useful.

scripts/normalize-by-median.py
bocajnotnef: The difficulty I have there is that the point of the context manager is to handle the error behind the scenes, but the accounting stuff relies on knowing if there was an error or not. 

Essentially all the solutions I can think of are flags--either something like:

```
err = none
[code]
elif error is not none:
```

or with a boolean.

scripts/normalize-by-median.py
bocajnotnef: Though I assume a line like `with WithDiagnostics as obj` would help, but that has to do more with context managers than generators

ctb: See #1039 and take what inspiration you can from it :)

bocajnotnef:
retest this please

bocajnotnef:
retest this please

bocajnotnef:
@ctb @luizirber Updated; CR when possible, please.

scripts/normalize-by-median.py
```diff
@@ -43,92 +47,163 @@ def batchwise(coll, size):
...
+def WithDiagnostics(ifile, batch_size, fp, paired, norm):
+    """
+    Generator/context manager to do boilerplate output of statistics while
+    normalizing data. Also checks for properly paired data.
+    """
 
-    desired_coverage = cutoff
-    ksize = htable.ksize()
+    index = 0
+    # global total, discarded
```
bocajnotnef: Whoops. I'll eliminate this in the next commit. (Assuming there's more stuff.)

ctb:
I have several minor comments, but my major comment is really: why not replace normalize_by_median(...) with a call to the Normalizer object, which would be created in main() and passed into normalize_by_median_and_check?

bocajnotnef:
Hrm. That does make the most sense--and is in keeping with the spirit of refactoring.

I never really understood the difference between normalize_..._and_check and regular normalize anywho.

I'll distill down the functionality over the weekend--gonna be busy from this afternoon through tomorrow.

The accounting for this will be so much fun.

bocajnotnef:
@ctb Updated

ctb:
Looking much neater, but my major comment remains: why not replace normalize_by_median(...) with a call to the Normalizer object, which would be created in main() and passed into normalize_by_median_and_check?

bocajnotnef:
I did that. `normalize_by_median(...)` no longer exists and all the processing that was done there has been moved to `normalize_by_median_and_check(...)`, which gets a `Normalizer` object from `main`. 

Which reminds me, there's some stuff I can move.

scripts/normalize-by-median.py
```diff
@@ -300,65 +361,59 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
-            if args.savetable:
-                hashname = args.savetable
-                print '...saving to', hashname
-            else:
-                hashname = 'backup.ct'
-                print 'Nothing given for savetable, saving to', hashname
-            htable.save(hashname)
+                f, htable, args.single_output_file,
+                args.fail_save, args.paired, args.force, norm, report_fp)
+        corrupt_files += corrupt
```
bocajnotnef: could just pull from `norm.corrupt_files` instead.

ctb:
I'm kind of allergic to the use of 'norm' in WithDiagnostics, but will accept it as a short-term evil :).  (If you didn't have 'norm' in there, this could be a more generally useful class; as it is, it's only usable in this script.)  I also think the code will become much cleaner with broken_paired_iterator.

So I'm -0 on merging this right now but will do so if you fix #1000 in this PR, and then promise to move quickly on the next stage of the refactor!

bocajnotnef:
retest this please

bocajnotnef:
@ctb I agree re norm in WithDiagnostics--I didn't really know of an easier way to get the information into WithDiagnostics that we needed (that being the rolling total/discarded counts). 

I've implemented @drtamermansour's fix for the PE countings in such a way that we maintain the current behavior AFAICT. 

I've got to get rid of `tests/test-data/paired_withN.fa` as I wiped out that test (or perhaps adapt it to check for the current behavior)

Next stage of the refactor would be implementing broken_paired_reader, which looks fun. I'll get to it as soon as this is CR'd/merged.

ctb:
OK, looks good to me. Let's get on that 2nd round cleanup tho :)

PR #1015, opened 2015-05-19
Fix: empty zipped files cause a C++ exception

kdmurray91:
Hi all,

The test in `kfile.check_input_file` only asserts that file size is `!= 0`. However, "empty" zipped files have non-zero file size due to the format's header. The zipped file is assumed to have reads, and gets passed into C++-land things which expect non-empty files. This throws an uncaught StreamReadError, which leaves an empty output file and isn't an overly clean way of exiting.

So, and this is a bit of a hack, I've checked if reading a byte from zipped files. If that returns an empty string, the file is empty.

Suggestions about how to check this portably without reading are welcome. I thought the gzip header size was 20 bytes, and it is on the cluster I have access to, but is 32 on my laptop. I assume it's due to different format versions.

Cheers,
Kevin


kdmurray91:
@camillescott @mr-c @luizirber This is "ready" for initial review/comments.

khmer/kfile.py
```diff
@@ -46,7 +48,21 @@ def check_input_files(file_path, force):
...
             sys.exit(1)
     else:
-        if os.stat(file_path).st_size == 0:
+        is_empty = False
+        file_size = os.stat(file_path).st_size
+
+        if file_size == 0:
+            is_empty = True
+        elif file_path.endswith('gz'):
+            fh = gzip.open(file_path, 'rb')
```
mr-c: I suspect that this will be bad for streaming support. We only want a single `open` in a given code path

kdmurray91: That's my fear. Any ideas about how to do it properly, and portably?

khmer/kfile.py
```diff
@@ -46,7 +48,21 @@ def check_input_files(file_path, force):
         if not force:
             sys.exit(1)
     else:
-        if os.stat(file_path).st_size == 0:
+        is_empty = False
+        file_size = os.stat(file_path).st_size
+
+        if file_size == 0:
+            is_empty = True
+        elif file_path.endswith('gz'):
```
mr-c: And this overlaps with the file type sniffing in screed. Perhaps this pre-check funcio should be replaced with a context manager?

kdmurray91:
Also, after reading up on the format, the extra size is due to the encoding of the original file name during creation of the test `.gz` file. So we can't rely on any specific size of the gzip file. A size of 20 is what is created by zlib, where there is no original file name to encode.

khmer/kfile.py
mr-c: See https://github.com/dib-lab/screed/blob/master/screed/openscreed.py#L40 for what screed does. Perhaps this functionality could move there via an optional argument.

khmer/kfile.py
kdmurray91: The entire `check_input_file`?

ctb:
Any conclusions here?

kdmurray91:
Not thus far :) Other than this is probably something that needs to go into screed.

mr-c:
On the screed side we return an empty list in the event of an empty file. So just a C++ issue, yep.

PR #1019, opened 2015-05-19
Add graphalign parameter getter/setters to CTB's graphalign branch

mr-c:


mr-c:
@ctb should be ready for you. Jenkins is failing due to test coverage drop on your PR vs master.

setup.py
```diff
@@ -182,7 +182,7 @@ def check_for_openmp():
 
         "extras_require": {':python_version=="2.6"': ['argparse>=1.2.1'],
                            'docs': ['sphinx', 'sphinxcontrib-autoprogram'],
-                           'tests': ['nose >= 1.0']},
+                           'tests': ['nose >= 1.0', 'pysam']},
```
ctb: I think this is a heavy requirement that is suitable for sandbox but we should rethink for longer term, & certainly prior to merge.  There was that replacement that came across twitter today that we should look at.

ctb:
Nice work, thanks!

setup.py
mr-c: Agreed, but it fails tests due to the import otherwise.

ctb: On Wed, May 20, 2015 at 07:53:43AM -0700, Michael R. Crusoe wrote:
> > @@ -182,7 +182,7 @@ def check_for_openmp():
> >  
> >          "extras_require": {':python_version=="2.6"': ['argparse>=1.2.1'],
> >                             'docs': ['sphinx', 'sphinxcontrib-autoprogram'],
> > -                           'tests': ['nose >= 1.0']},
> > +                           'tests': ['nose >= 1.0', 'pysam']},
> 
> Agreed, but it fails tests due to the import otherwise. 

Yes, we should either skip the test in that case or find a lighterweight
dep, I think.

PR #1020, opened 2015-05-20
Clean up gitignore and make clean a little

kdmurray91:
Just a few housekeeping things:

- `make clean` now uses `rm -f`, which doesn't print an error message if a file doesn't exist.
- Add coverage outputs to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,8 @@ doc/_build
 coverage.xml
 coverage-gcovr.xml
 coverage-debug
+htmlcov/
+diff-cover.html
 khmer-cov.tgz
 cppcheck-result.xml
 MANIFEST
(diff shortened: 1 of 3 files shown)


kdmurray91:
ping @luizirber @ctb @camillescott @mr-c 

- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

kdmurray91:
Should the makefile have a copyright header?

ctb:
LGTM. I will leave for @mr-c to double-check and merge. I don't think the Makefile needs a (c) but it's fine by me if it has one.

mr-c:
LGTM. We might do a copyright header overhaul later so don't worry about it for now.

PR #1039, opened 2015-05-28
try a class for the core processing bit of normalize-by-median

ctb:
So @bocajnotnef here you could replace normalize_by_median_and_check with several functions in Normalizer, build a new Normalizer for each input file, and then accumulate total/discarded in the outermost for loop across all the filenames.  Make some sense?

The basic idea is to replace globals with objects that "package" all of that information in one place.

### DIFF
--- a/scripts/normalize-by-median.py
+++ b/scripts/normalize-by-median.py
@@ -86,48 +86,59 @@ def WithDiagnostics(ifile, batch_size, fp, paired):
         yield batch
 
 
+class Normalizer(object):
+    def __init__(self, desired_coverage, htable, report_fp=None):
+        self.htable = htable
+        self.desired_coverage = desired_coverage
+        self.report_fp = report_fp
+
+        self.total = 0
+        self.discarded = 0
+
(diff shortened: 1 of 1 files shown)


bocajnotnef:
Right, okay. I can run with this. Thanks!

ctb:
On Thu, May 28, 2015 at 08:01:53AM -0700, Jake Fenton wrote:
> Right, okay. I can run with this. Thanks!

p.s. it even passes most of the tests! :)

bocajnotnef:
retest this please

luizirber:
This class is almost a context manager itself, if you rename `__call__` to `__iter__` and implement `__enter__` and `__exit__` for initialization

ctb:
On Fri, May 29, 2015 at 08:59:33AM -0700, Luiz Irber wrote:
> This class is almost a context manager itself, if you rename `__call__` to `__iter__` and implement `__enter__` and `__exit__` for initialization

thanks - that was my suspicion but I decided to leave it for someone else
to confirm :)

Issue #1013, opened 2015-05-18
document PR git cleanup


mr-c:
Here is a faster method tested by @ctb and me. 

If there are merge commits then this approach is always easier than rebasing.

```
git pull # make sure the local is up to date
git pull dib master # get up to date
# fix any merge conflicts
git status # sanity check
git diff dib/master # does the diff look correct? (no merge markers)
git reset --soft dib/master # un-commit the differences from dib/master
git status # sanity check
git commit --all # package all differences in one commit
git status # sanity check
git push # should fail
git push --force # override what's in GitHub's copy of the branch/pull request
# Code review and/or merging from the pull request
```

(`git reset --soft` keeps the tracking for all newly added files )

bocajnotnef:
+1 for having this formally documented

mr-c:
Fixed in #1222

PR #1042, opened 2015-05-28
Minor Fixes to split-paired-reads  mentioned in #763

SherineAwad:


ctb:
jenkins, add to test list

ctb:
please tests this, jenkins

ctb:
jenkins, please test this

ctb:
please test this, jenkins

SherineAwad:
Jenkins test this please

scripts/split-paired-reads.py
```diff
@@ -10,7 +10,7 @@
 De-interleave a file.
 
 Take an interleaved set of reads (/1 and /2), and extract them into separate
-files (.1 and .2).
+files (.2 and .2).
```
ctb: this doesn't seem right...

scripts/split-paired-reads.py
```diff
@@ -122,28 +123,28 @@ def main():
     screed_iter = screed.open(infile, parse_description=False)
 
     # walk through all the reads in broken-paired mode.
-    for index, is_pair, record1, record2 in broken_paired_reader(screed_iter):
-        if index % 100000 == 0 and index:
-            print >> sys.stderr, '...', index
-
+    paired_iter = broken_paired_reader(screed_iter)
```
ctb: why are you making these changes? (record -> read, index -> n?) they seem unnecessary.

ctb:
Please reference the original issue in a comment - that way it will get linked in, like this: #763

ctb:
OK, apart from comments - what else are you planning on doing? thanks :)

SherineAwad:
@ctb I did not finish editing.  nosetest passes  but just asked Jenkins test to see what's going on and it doesn't reply.

scripts/split-paired-reads.py
SherineAwad: For no reason but imagining what the code doing and planning to get them back at the end :)

SherineAwad:
test_split_paired_reads_1_fa(), test_split_paired_reads_2_fq(), test_split_paired_reads_2_mixed_fq_require_pair(), test_split_paired_reads_2_mixed_fq(), test_split_paired_reads_2_mixed_fq_broken_pairing_format()

The above functions in test_scripts.py call split_paired_reads with arg=infile only. Now we make the output files required, should I make any changes in these test functions?

ctb:
Output files should only be required when no input filename is specified, I think.

SherineAwad:
So, if no input is specified, and test is called with no output files as it is, it will fail, right?

ctb:
Input should default to /dev/stdin if not provided, but then require output files; so, yes!

SherineAwad:
So, this should be handled in the testing? I mean if no input is specified, test should pass the output files  too?

ctb:
Yes, it should be handled in the testing.

SherineAwad:
My question is: can I make these changes in the test functions or it is assigned to someone else?
As it is, 'make test' fails when the output files should be used.

ctb:
On Mon, Jun 01, 2015 at 08:26:52AM -0700, Sherine wrote:
> My question is: can I make these changes in the test functions or it is assigned to someone else?
> As it is, 'make test' fails when the output files should be used.

Yep, please make the changes in the functions.

SherineAwad:
@ctb in the master version of split-paired-reads, if the user specify filenames with no path in options -1 and -2 and specified a different output directory in -o different than his current directory. The output files will be thrown in the current directory and the -o directory will be ignored, right? 
Did you mean it, override the output directory too by -1 and -2 ?

ctb:
I think -o should only apply when we can figure out a default file name. So, -o should be ignored when -1 and -2 specified.


--

SherineAwad:
This just makes input default from stdin and requires input files in that case. Plus it updates testing functions accordingly. @ctb please take a look.  

- [x] Is it mergable?
- [x] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

scripts/split-paired-reads.py
```diff
@@ -92,11 +92,13 @@ def main():
 
     infile = args.infile
 
-    check_input_files(infile, args.force)
     filenames = [infile]
     check_space(filenames, args.force)
-
     # decide where to put output files - specific directory? or just default?
+    if os.path.basename(infile) == 'stdin':
```
ctb: I think this should just be 'if infile == '/dev/stdin' - no need to use basename.  Also, check for infile == '-'.

scripts/split-paired-reads.py
```diff
@@ -92,11 +92,13 @@ def main():
...
+        if not (args.output_first and args.output_second):
+            print >>sys.stderr, 'Output files are missing'
```
ctb: error message should say Accepting input from stdin; output files must be provided.

scripts/split-paired-reads.py
```diff
@@ -114,7 +116,6 @@ def main():
 
     fp_out1 = open(out1, 'w')
     fp_out2 = open(out2, 'w')
-
```
ctb: please don't change things accidentally :)

ctb:
Hi @SherineAwad I made some comments.  On the tests, I haven't looked closely, but most of the existing tests should still work, shouldn't they? Please only modify the tests that need to be changed due to differences in the way the script works, and then *add* new tests (by copying and renaming existing tests, and then editing them) to cover the new behavior.

SherineAwad:
@ctb I think you are right. You can take a look pls.

ChangeLog
```diff
@@ -1,3 +1,7 @@
+2015-06-02  Sherine Awad  <drmahmoud@ucdavis.edu>
+   * Fix minor issues in split-paired-reads
```
ctb: Could you describe in a bit more detail what you've done? Also, please put a space between your name and the first line of changes.

scripts/split-paired-reads.py
```diff
@@ -92,11 +92,15 @@ def main():
 
     infile = args.infile
 
-    check_input_files(infile, args.force)
```
ctb: This line needs to be added back in for situations where filenames are provided.

scripts/split-paired-reads.py
```diff
@@ -92,11 +92,15 @@ def main():
...
     filenames = [infile]
     check_space(filenames, args.force)
 
     # decide where to put output files - specific directory? or just default?
+    if infile == '/dev/stdin':
+        if not (args.output_first and args.output_second):
+            print >>sys.stderr, ("Accepting input from stdin;"
```
ctb: please put a space after the ';' - this is printed out as

    Accepting input from stdin;output filenames must be provided.

currently.

scripts/split-paired-reads.py
```diff
@@ -92,11 +92,15 @@ def main():
...
+                                 "output filenames must be provided.")
+            sys.exit(1)
     if args.output_directory:
```
ctb: This should be 'elif' now - you only want to pay attention to output_directory if infile is NOT /dev/stdin.

SherineAwad:
Done !

ctb:
Could you add some tests for the new behavior, please?

SherineAwad:
Yes, I am working on this.

SherineAwad:
@ctb  Please take a look. Thanks.

scripts/split-paired-reads.py
ctb: In general, you should avoid "noise" in the change - here you moved check_input_files down by a line, creating a change for no reason.  It's OK, but please try to avoid in the future. tnx!

ctb:
@SherineAwad looks good; thanks!

ctb:
Hi @SherineAwad I squashed your commits as in #1013 before merging.  To start work on a new issue, do

     git checkout master
     git pull origin master

and then proceed from there!

PR #1033, opened 2015-05-27
Refactor CPython layer so that CountingHash and Hashbits inherit from Hashtable.

ctb:
Note: depends on branch feature/labelgraph.

PR #1044, opened 2015-05-29
Refactor CPython layer so that CountingHash and Hashbits inherit from Hashtable.

ctb:
Re-opened #1033 against master.

- [ ] should khmer_labelhash_init be removed from _khmermodule.cc? What purpose does it serve?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,13 @@
 2015-05-26  Titus Brown  <titus@idyll.org>
 
+   * khmer/_khmermodule.cc: refactor CPython layer so that KHashtable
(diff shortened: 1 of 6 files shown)


ctb:
@mr-c @luizirber review and merge, please? Note question re khmer_labelhash_init.

ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

luizirber:
@ctb about removing `khmer_labelhash_init`: we might need to change this for other classes, so I think we can do this in another PR.

Issue #1037, opened 2015-05-28
write truncation tests for file load functions


ctb:
see test_labelhash.py, function test_save_load_corrupted(), added in #1021; this could usefully be applied to all of the file load functions.  Of course, then we'd need to fix the bugs, too...

Might be interesting to try postpending garbage into the files, too. Are the load functions robust to random crud?

mr-c:
Sweet, fuzz'ing time

ctb:
fixed by #1048.

PR #1031, opened 2015-05-26
Add file type signatures to binary files

mr-c:
Closes #519

PR #1048, opened 2015-05-30
test various load functions

ctb:
Responsive to #1037.


ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

ctb:
Ready for CR and merge @luizirber @camillescott.  Note, it merges cleanly with #1031 with all tests passing, too.

lib/subset.cc
```diff
@@ -1350,20 +1355,22 @@ void SubsetPartition::merge_from_disk(string other_filename)
             diskp = (PartitionID *) (buf + i);
             i += sizeof(PartitionID);
 
-            if (!(*diskp != 0)) {		// sanity check.
-                throw khmer_exception();
-            }
+            assert((*diskp != 0)); // sanity check!
```
camillescott: Thought we were trying to avoid `assert`s in the C++ code?

lib/subset.cc
```diff
@@ -1350,20 +1355,22 @@ void SubsetPartition::merge_from_disk(string other_filename)
...
         memcpy(buf, buf + n_bytes, remainder);
     }
 
+    if (loaded != expected_pmap_size) {
+        delete[] buf;
+        throw khmer_file_exception("error loading partitionmap - "
+                              "invalid # of items");
+    }
+
     delete[] buf;
```
camillescott: Move this up to line 1367 to avoid duplication

lib/subset.cc
ctb: On Thu, Jun 04, 2015 at 06:38:49PM -0700, Camille Scott wrote:
> > @@ -1350,20 +1355,22 @@ void SubsetPartition::merge_from_disk(string other_filename)
> >              diskp = (PartitionID *) (buf + i);
> >              i += sizeof(PartitionID);
> >  
> > -            if (!(*diskp != 0)) {		// sanity check.
> > -                throw khmer_exception();
> > -            }
> > +            assert((*diskp != 0)); // sanity check!
> 
> Thought we were trying to avoid `assert`s in the C++ code?

I was unable to think of a way to trigger this in a test, and this is a
reasonable check on things when debugging or modifying other code.  I think
we're trying to avoid asserts as things that are actually raised for users, but
I don't see any reason not to leave them in as sanity checks.

tests/test_subset_graph.py
```diff
@@ -230,6 +230,43 @@ def test_save_load_merge(self):
...
+        print divvy
+        assert len(divvy) is 3
+        (a, b, c) = divvy
+
+        outfile1 = utils.get_temp_filename('x.pmap')
+        outfile2 = utils.get_temp_filename('y.pmap')
+
+        x = ht.do_subset_partition(a, b)
+        ht.save_subset_partitionmap(x, outfile1)
+        del x
```
camillescott: Not really necessary in python; `del` doesn' actually free the memory, just notifies the gc that it *can* be freed

lib/subset.cc
camillescott: Fair enough!

On Thu, Jun 4, 2015 at 6:46 PM, C. Titus Brown <notifications@github.com>
wrote:

> In lib/subset.cc
> <https://github.com/dib-lab/khmer/pull/1048#discussion_r31782427>:
>
> > @@ -1350,20 +1355,22 @@ void SubsetPartition::merge_from_disk(string other_filename)
> >              diskp = (PartitionID *) (buf + i);
> >              i += sizeof(PartitionID);
> >
> > -            if (!(*diskp != 0)) {		// sanity check.
> > -                throw khmer_exception();
> > -            }
> > +            assert((*diskp != 0)); // sanity check!
>
> On Thu, Jun 04, 2015 at 06:38:49PM -0700, Camille Scott wrote: > @@
> -1350,20 +1355,22 @@ void SubsetPartition::merge_from_disk(string
> other_filename) > diskp = (PartitionID *) (buf + i); > i +=
> sizeof(PartitionID); > > - if (!(*diskp != 0)) { // sanity check. > - throw
> khmer_exception(); > - } > + assert((*diskp != 0)); // sanity check!
> Thought we were trying to avoid `assert`s in the C++ code?
> I was unable to think of a way to trigger this in a test, and this is a
> reasonable check on things when debugging or modifying other code. I think
> we're trying to avoid asserts as things that are actually raised for users,
> but I don't see any reason not to leave them in as sanity checks.
>
> —
> <https://github.com/dib-lab/khmer/pull/1048/files#r31782427>.
>



-- 
Camille Scott

Department of Computer Science
Lab for Data Intensive Biology
University of California, Davis

camille.scott.w@gmail.com

tests/test_subset_graph.py
ctb: On Thu, Jun 04, 2015 at 06:46:32PM -0700, Camille Scott wrote:
> > +        filename = utils.get_test_data('test-graph2.fa')
> > +
> > +        (total_reads, total_kmers) = ht.consume_fasta_and_tag(filename)
> > +        assert total_reads == 3, total_reads
> > +
> > +        divvy = ht.divide_tags_into_subsets(1)
> > +        print divvy
> > +        assert len(divvy) is 3
> > +        (a, b, c) = divvy
> > +
> > +        outfile1 = utils.get_temp_filename('x.pmap')
> > +        outfile2 = utils.get_temp_filename('y.pmap')
> > +
> > +        x = ht.do_subset_partition(a, b)
> > +        ht.save_subset_partitionmap(x, outfile1)
> > +        del x
> 
> Not really necessary in python; `del` doesn' actually free the memory, just notifies the gc that it *can* be freed

It does, however, prevent reuse of the variable :)

lib/subset.cc
ctb: good catch!

camillescott:
Fine fine, keep your extraneous statements ;P LGTM; merging

PR #1052, opened 2015-05-31
A simple in-memory implementation of Sequence Bloom Trees

ctb:
This is a simple implementation of Sequence Bloom Trees, (Solomon & Kingsford; https://github.com/Kingsford-Group/bloomtree).

See https://github.com/dib-lab/khmer/blob/sbt/sbt.py for implementation.


ctb:
Moved to https://github.com/ctb/2015-khmer-sequence-bloom-trees/

Issue #746, opened 2015-01-30
sandbox/unique-kmers.py output needs modification


ctb:
A few things -

* output should include the k-mer size used;
* output should go to stdout, or wherever -o specifies (see readstats.py)

If it makes sense (i.e. not significant memory penalty):

* num k-mers should be reported on a per-file basis as well as an overall summary

ctb:
Fixed in #1053.

PR #1053, opened 2015-05-31
HLL counter adaptation that can monitor streams of sequences

ctb:
A script that outputs all input sequences, so that you can do e.g.

    cat reads.fa | \
            sandbox/unique-kmers-passthru.py /dev/stdin -R predn.txt | \
            scripts/normalize-by-median.py -k 20 -C 20 -x 1e7 -o /dev/stdout /dev/stdin | \
            sandbox/unique-kmers-passthru.py -R postdn.txt /dev/stdin
                 > /dev/null

which is actually pretty useful in our streaming worldview...

Issues that arise:

- [x] should be configurable (default?) behavior on unique-kmers.py
- [x] would be nice to be able to default to '-' for input
- [x] HLLCounter.consume_string doesn't handle lowercase or non-ACGT characters

and probably more.

@luizirber you interested in taking this on? #746 could be tackled at the same time.


ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

ctb:
Ready for review (& merge?) @luizirber

Issue #988, opened 2015-05-08
Properly handle singleton reads in normalize-by-median.


ctb:
This fits into the #732 bucket, improving user experience.

These ideas and issues most recently raised in conversation with @drtamermansour, but leaning on conversations with @camillescott and @mr-c also.

singleton reads/read files are poorly handled by normalize-by-median in workflows and elsewhere. A few thoughts --

1. By default, normalize-by-median should accept and deal with broken-paired properly (pairs dealt with as pairs, singletons dealt with as singletons).

2. -p should require correct pairing, and we will need a new option to force into singleton mode (all sequences treated as singletons) - looking at sample-reads-randomly/#933, --force-single seems fine.

3. We should provide a new option that, with '-p' forcing pairs, allows a singleton file (or files) to be passed in and treated w/o requiring pairing.  Something like --singletons.

ctb:
Note that option number 3 was already implemented in #957 by @susinmotion! Excellent :)

bocajnotnef:
re 2: What does 'correctly paired' mean in this context?

ctb:
Operationally speaking, the broken_paired_reader will return two records if the records are properly paired.  See existing -p logic.

bocajnotnef:
If -p already does correct pairing is that component of 2 done, and we just need force-singletons?

ctb:
yes, assuming you have integrated in broken_paired_reader.

drtamermansour:
This discussion about correct pairing is confusing to me. Please correct me if I am getting this wrong. I think option number 1 means that the script would accept the input data file (without any arguments). The script would expect that the input has mixed proper paired and singletons (e.g. The output of fastq_quality_filter). This mean that the user does not need to extract paired ends from the interleaved files anymore.  Option  2 means the -p will expect a correctly paired end file and will fail with file containing mix of PE and single tones. The --force-single will forcibly utilize the input as single reads.

ctb:
Yep, exactly!

ctb:
Done in #1057 and #1010.

PR #1057, opened 2015-06-01
Refactor normalize-by-median stage 1: broken_paired

bocajnotnef:
In this exciting episode of "Jake plays with khmer" we'll be implementing broken_paired_reader in noramlize-by-median to fix some things.

Relevant to #1006, #988


scripts/normalize-by-median.py
```diff
@@ -55,14 +55,25 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
-    for index, batch in enumerate(batchwise(
-                                  screed.open(ifile, parse_description=False),
-                                  batch_size)):
+    screed_iter = screed.open(ifile, parse_description=False)
+    reader = broken_paired_reader(screed_iter, force_single=single)
 
-        norm.total += batch_size
+    for index, paired_reads, read0, read1 in reader:
+
+        batch = []
```
bocajnotnef: I'm split on this--Currently, I have it set up so that the actual normalize algorithm/implementation doesn't have to be changed, it still uses a `batch` and not two independent reads (or one read, as the case may be).

I feel like it's cleaner this way--otherwise I'd have to overhaul a lot of Normalizer's logic and it'd just look messier for no real gain in functionality. 

@luizirber @ctb @camillescott Opinions?

bocajnotnef:
in the code to handle paired/unpaired data:

```
412         if not args.single_output_file:                                        
413             output_name = os.path.basename(args.unpaired_reads) + '.keep'      
414         outfp = open(output_name, 'w')  
```

Commented it out and nothing breaks and the variable `outfp` doesn't appear to be used anywhere in the code.

bocajnotnef:
Aha, it appears that it's handled within `normalize_by_median_and_check`

Weird.

scripts/normalize-by-median.py
```diff
@@ -55,14 +55,25 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
+        batch.append(read0)
+        if read1 is not None:
+            batch.append(read1)
+
+        if paired_reads:
```
bocajnotnef: unnecessary: remove

scripts/normalize-by-median.py
```diff
@@ -72,25 +83,32 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
 
             print('... in file ' + input_filename, file=sys.stderr)
 
-            if report_fp:
+            if fp:
```
bocajnotnef: change recommended by @luizirber from the last pull:
https://github.com/dib-lab/khmer/commit/a163fa645a8938d557d0e4a586741fddf36b59cf#commitcomment-11477819

scripts/normalize-by-median.py
ctb: On Tue, Jun 02, 2015 at 11:01:08AM -0700, Jake Fenton wrote:
> >  
> > -        norm.total += batch_size
> > +    for index, paired_reads, read0, read1 in reader:
> > +
> > +        batch = []
> 
> I'm split on this--Currently, I have it set up so that the actual normalize algorithm/implementation doesn't have to be changed, it still uses a `batch` and not two independent reads (or one read, as the case may be).
> 
> I feel like it's cleaner this way--otherwise I'd have to overhaul a lot of Normalizer's logic and it'd just look messier for no real gain in functionality. 
> 
> @luizirber @ctb @camillescott Opinions?

*shrug* submit your code and we'll take a look :)

bocajnotnef:
retest this, please

bocajnotnef:
@ctb @luizirber CR, please?

scripts/normalize-by-median.py
```diff
@@ -72,25 +76,25 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
                       1. - (discarded / float(total)), file=fp)
                 report_fp.flush()
 
         # If in paired mode, check that the reads are properly interleaved
-        if paired:
-            if not check_is_pair(batch[0], batch[1]):
-                raise IOError('Error: Improperly interleaved pairs \
-                    {b0} {b1}'.format(b0=batch[0].name, b1=batch[1].name))
-
+        if paired and not paired_reads:
```
ctb: This is confusing language... suggest 'force_paired' for first name, 'is_paired' for second variable name.

scripts/normalize-by-median.py
```diff
@@ -55,11 +53,17 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
 
     index = 0
 
-    for index, batch in enumerate(batchwise(
-                                  screed.open(ifile, parse_description=False),
-                                  batch_size)):
+    screed_iter = screed.open(ifile, parse_description=False)
+    reader = broken_paired_reader(screed_iter, force_single=single)
```
ctb: reader should be passed in; broken_paired_reader should be called outside of WithDiagnostics; filename can be passed in together with reader but should only be used for reporting.

scripts/normalize-by-median.py
```diff
@@ -55,11 +53,17 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
+
+    for index, paired_reads, read0, read1 in reader:
 
-        norm.total += batch_size
+        batch = []
```
ctb: please move batching into normalizer; it's not needed here.

scripts/normalize-by-median.py
```diff
@@ -102,12 +106,8 @@ def __call__(self, input_filename, force_paired=False):
         desired_coverage = self.desired_coverage
         ksize = self.htable.ksize()
 
-        batch_size = 1
-        if force_paired:
-            batch_size = 2
-
-        for batch in WithDiagnostics(input_filename, batch_size,
-                                     self.report_fp, force_paired, self):
+        for batch in WithDiagnostics(input_filename, self.report_fp,
```
ctb: WithDiagnostics should yield as if it's broken_paired_reader, not batches.

ctb:
done.

bocajnotnef:
@ctb Updated; CR/merge?

scripts/normalize-by-median.py
```diff
@@ -55,11 +53,13 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
 
     index = 0
 
-    for index, batch in enumerate(batchwise(
-                                  screed.open(ifile, parse_description=False),
-                                  batch_size)):
+    for index, is_paired, read0, read1 in reader:
+
+        if read1:
```
ctb: if is_paired:

ctb:
Please remove batchwise function.

bocajnotnef:
@ctb Updated (and a streak of five good CI checks)

ctb:
A few explicit goals of my review moving forward, to help guide you in your changes:

first, eliminate as many of the function arguments as possible. You're passing around a lot of information that seems unnecessary. Actually eliminating it is tricky but I'll do my best to help you figure out how :)

second, I'd like you to make as much of the utility code as reusable as possible. I'm particularly interested in the checkpointing/failsafe stuff, and perhaps the diganostics stuff.

More specific points incoming shortly.

bocajnotnef:
Part of me was considering just passing around the `args` from main but that seems like a bad way to do it.

My concern with making more use of the utility code is that it would specialize it more, which would reduce it's re-usability.

ctb:
On Wed, Jun 03, 2015 at 10:07:05AM -0700, Jake Fenton wrote:
> Part of me was considering just passing around the `args` from main but that seems like a bad way to do it.

Yes, I don't mean "package same information up in fewer variables", I mean
"minimize information passing".

> My concern with making more use of the utility code is that it would specialize it more, which would reduce it's re-usability.

I meant the opposite of that :)

bocajnotnef:
Right, agreed on both points. Tricky indeed, but that's the fun.

ctb:
See #1064 for a suggestion; what do you think?

ctb:
Can you try moving the reporting here:

https://github.com/dib-lab/khmer/blob/fd42746179e53c05b4aa3dc35c97f738fe0962bf/scripts/normalize-by-median.py#L210

and here:

https://github.com/dib-lab/khmer/blob/fd42746179e53c05b4aa3dc35c97f738fe0962bf/scripts/normalize-by-median.py#L173

into WithDiagnostics?

ctb:
This is called "programming by undergraduate", in case you're wondering :)

bocajnotnef:
the broken_paired requiring/dealing with paired makes sense--I it was mentioned in the original `broken_pair_reader` pull that it needed a `force_paired`

The first bit shouldn't be too hard to migrate, though as it is it reports per-file so I'll have to find a way to modify a generator's `__exit__`.

The second bit should be easy.

Woo, learning experiences, woo.

ctb:
Don't worry too much about adding tests and so on. Mostly interested in seeing if it looks cleaner this way.

bocajnotnef:
@ctb I did the things. Even managed to pull an arg out of `CatchIOErrors`--all nice and generalizedish now.

scripts/normalize-by-median.py
```diff
@@ -70,31 +66,42 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
-        # If in paired mode, check that the reads are properly interleaved
-        if paired:
-            if not check_is_pair(batch[0], batch[1]):
-                raise IOError('Error: Improperly interleaved pairs \
-                    {b0} {b1}'.format(b0=batch[0].name, b1=batch[1].name))
+        yield index, is_paired, read0, read1
 
-        yield batch
+    # per file diagnostic output
+    if norm.total == 0 and norm.discarded == 0:
```
ctb: When would norm.total == 0 and norm.discarded be nonzero?

scripts/normalize-by-median.py
```diff
@@ -70,31 +66,42 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
+        print('SKIPPED empty file ' + ifilename, file=sys.stderr)
+    if norm.total > 0:
+        print('DONE with {inp}; kept {kept} of {total} or {perc:2}%'
+              .format(inp=ifilename, kept=norm.total - norm.discarded,
+                      total=norm.total, perc=int(100. - norm.discarded /
+                                                 float(norm.total) * 100.)),
+              file=sys.stderr)
+
+    if fp:
```
ctb: holy tabbing, batman!

scripts/normalize-by-median.py
```diff
@@ -70,31 +66,42 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
         self.desired_coverage = desired_coverage
         self.report_fp = report_fp
+        self.force_single = force_single
 
         self.total = 0
         self.discarded = 0
-        self.corrupt_files = []
 
     def __call__(self, input_filename, force_paired=False):
         seq = ""
```
ctb: why is 'seq=""' here?

scripts/normalize-by-median.py
```diff
@@ -102,15 +109,23 @@ def __call__(self, input_filename, force_paired=False):
...
             passed_filter = False
             passed_length = True
 
+            batch = []
+            batch.append(read0)
+            if read1 is not None:
+                batch.append(read1)
+
             for record in batch:
                 if len(record.sequence) < ksize:
```
ctb: note that broken_paired_reader can take a minimum sequence length (ksize), making this length check unused.  Although this may break the pairing code in this case... hmm...

scripts/normalize-by-median.py
ctb: (if you were unsure, you could put in an assertion:

    if norm.total == 0:
        assert norm.discarded == 0
        ...

scripts/normalize-by-median.py
```diff
@@ -146,39 +161,23 @@ def handle_error(error, output_name, input_name, fail_save, htable):
               file=sys.stderr)
 
 
+# TO-DO: find a better way of getting corrupt files out of here
 @contextmanager
-def CatchIOErrors(ifile, ofile, save_on_fail, ht, force, norm):
+def CatchIOErrors(ifile, ofile, save_on_fail, ht, force):
     """
     Context manager to do boilerplate excepting of IOErrors; also does
     upkeep on some statistics and diagnostic output.
```
ctb: this is now an incorrect docstring :)

scripts/normalize-by-median.py
```diff
@@ -146,39 +161,23 @@ def handle_error(error, output_name, input_name, fail_save, htable):
...
     """
-
-    # global corrupt_files, total, discarded, total_acc, discarded_acc
-    caught_error = False
     try:
-        yield
-    except IOError as err:
-        caught_error = True
+        yield False
```
ctb: huh?

scripts/normalize-by-median.py
```diff
@@ -373,28 +376,25 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
-        output_name = args.unpaired_reads
-        if not args.single_output_file:
-            output_name = os.path.basename(args.unpaired_reads) + '.keep'
-        outfp = open(output_name, 'w')
-        total_acc, discarded_acc, corrupt = \
-            normalize_by_median_and_check(
-                args.unpaired_reads, htable, args.single_output_file,
-                args.fail_save, args.paired, args.force, norm, report_fp)
-        corrupt_files += corrupt
+                args.fail_save, p, args.force, norm, report_fp)
```
ctb: report_fp is now unused in normalize_by_median_and_check

scripts/normalize-by-median.py
bocajnotnef: Never, I suppose. Was in the original line so I copied it over.

scripts/normalize-by-median.py
```diff
@@ -189,19 +188,19 @@ def normalize_by_median_and_check(input_filename, htable, single_output_file,
...
         else:
             output_name = single_output_file.name
         outfp = single_output_file
-
     else:
         output_name = os.path.basename(input_filename) + '.keep'
         outfp = open(output_name, 'w')
 
-    with CatchIOErrors(input_filename, outfp, fail_save, htable, force, norm):
+    with CatchIOErrors(input_filename, outfp, fail_save, htable, force) as \
```
ctb: you probably want output_name here rather than outfp. and do you really want it being removed in handle_error if there's a single output name and a failure? Probably should be a test for that, too :)

scripts/normalize-by-median.py
bocajnotnef: I blame pep8.

scripts/normalize-by-median.py
ctb: More generally, the passing of several parameters into normalize_by_median_and_check only to be passed into CatchIOErrors smells bad to me. What about moving CatchIOErrors out to the main loop, outside of normalize_by_median_and_Check? That means you can avoid passing single_output_file, fail_save, and force into normalize_by_median_and_Check.

scripts/normalize-by-median.py
```diff
@@ -209,12 +208,12 @@ def normalize_by_median_and_check(input_filename, htable, single_output_file,
         total = norm.total
```
ctb: what do these two lines of code do?

scripts/normalize-by-median.py
ctb: ooh, and you could avoid passing 'corrupt' back out, too.

scripts/normalize-by-median.py
ctb: > I blame pep8.

nice try :)

scripts/normalize-by-median.py
bocajnotnef: Old changes. Eliminated now.

ctb:
BTW lest you feel despair, it IS looking a lot better, so thank you :)

scripts/normalize-by-median.py
bocajnotnef: Hrm. I thought this is what let me pull `norm` out of the arguments. Now I'm not so sure. Gonna do more testing.

(The idea was that it would yield something useful if there weren't any IOErrors and yeild None when there were, which is detectable outside of the context manager.

Instead, it seems that it will always yield the thing.

I may revert to the "real" way of making a CM because I think then I can give it attributes which I can use to pull out the corrupt files--Otherwise I'd have to pass in an object of some variant.

Python's inability to take referenced variables is frustrating.

scripts/normalize-by-median.py
bocajnotnef: re: corrupt: any insight on how to get the list of corrupt files out of CatchIOErrors without throwing in more args? 

Originally I made the corrupt_files an attribute of the normalizer object, which worked but seemed messy. 

I tried passing in a list that was created out of CIOE but it appears to just copy it, and I can't yield it out magically and CIOE doesn't let me give it attributes (i.e., yielding out "self" and doing "CIOE.corrupt_files" on the outside)

scripts/normalize-by-median.py
bocajnotnef: originally they made things neat for the diagnostics that used to be right after those lines. When I moved the diags I forgot about them.

bocajnotnef:
Not despair, just.. getting frustrated with how python deals with args and copying/not copying.

This is actually rather enjoyable, if you can believe that.

scripts/normalize-by-median.py
bocajnotnef: also, single_output_file would just be replaced by outfp or summuch

scripts/normalize-by-median.py
bocajnotnef: Well. It's not entirely inaccurate.

scripts/normalize-by-median.py
bocajnotnef: interestingly, when I comment that out the test that breaks is double_file_name:

```
WARNING: At least two input files are named test-abund-read-2.fa . (The script normalize-by-med
ian.py can not handle this, only one .keep file for one of the input files will be generated.)
making k-mer counting table
** ERROR: string length must >= the hashtable k-mer size
** Failed on /home/bocajnotnef/git/khmer/tests/test-data/test-abund-read-2.fa: 
** ...dumping k-mer counting table to test-abund-read-2.fa.ct.failed
** ERROR: problem removing corrupt filtered file
** Exiting!

```

bocajnotnef: indeed unnecessary.

unless it's breaking things the tests don't cover.

scripts/normalize-by-median.py
bocajnotnef: Ignore this, I figured it out.

I think.

bocajnotnef:
@ctb I did magic. Do I get a cookie?

ChangeLog
```diff
@@ -1,3 +1,10 @@
+2015-06-02  Jacob Fenton  <bocajnotnef@gmail.com>
+
+   * scripts/normalize-by-median.py: implemented broken_paired_reader
+   * tests/test_scripts.py: modified tests to properly use new args
+   * khmer/utils.py: added force-paired option to broken_paired_reader (@ctb)
+
+
```
bocajnotnef: Hmm. Gotta delete this.

scripts/normalize-by-median.py
```diff
@@ -55,80 +48,94 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
-    for index, batch in enumerate(batchwise(
-                                  screed.open(ifile, parse_description=False),
-                                  batch_size)):
+    # per read diagnostic output
+    for index, is_paired, read0, read1 in reader:
 
-        norm.total += batch_size
-        total = norm.total
-        discarded = norm.discarded
+        if is_paired:
```
ctb: Please move this accounting into the Normalizer object.  No need to mess with norm.total here.

scripts/normalize-by-median.py
```diff
@@ -55,80 +48,94 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
-        # If in paired mode, check that the reads are properly interleaved
-        if paired:
-            if not check_is_pair(batch[0], batch[1]):
-                raise IOError('Error: Improperly interleaved pairs \
-                    {b0} {b1}'.format(b0=batch[0].name, b1=batch[1].name))
+        yield index, is_paired, read0, read1
 
-        yield batch
+    # per file diagnostic output
+    if norm.total == 0:
```
ctb: Would suggest swapping if / else here to put most common case first.

scripts/normalize-by-median.py
ctb: Please fix the tabbing so that the print statement is only indented one block.

scripts/normalize-by-median.py
```diff
@@ -373,28 +331,43 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
+        files.append([args.unpaired_reads, False])
+
+    corrupt_files = []
+
+    for f, p in CheckpointCountingTable(files, args.dump_frequency, htable,
+                                        args.savetable):
+        outfp = None
+
+        # figure out what we're calling out output file
+        if args.single_output_file:
```
ctb: Half of this 'if' can be moved outside the loop; inside the loop, you only need to set the output_name `if not args.single_output_file`.

tests/test_scripts.py
```diff
@@ -601,6 +602,40 @@ def test_normalize_by_median():
...
+def test_normalize_by_median_unforced_badfile():
+    CUTOFF = '1'
+
+    infile = utils.get_temp_filename("potatoes")
+    in_dir = os.path.dirname(infile)
+    script = scriptpath('normalize-by-median.py')
+    args = ['-C', CUTOFF, '-k', '17', '-t', infile]
+    try:
+        (status, out, err) = utils.runscript(script, args, in_dir)
+        raise Exception("Shoulnd't get to this")
```
ctb: spelling

scripts/normalize-by-median.py
bocajnotnef: note: this still needs to be addressed--don't think there's a test in place for handle_error on single_output_file

ctb:
I think I'm going to make myself very unpopular with @bocajnotnef with this thought:

We should remove the IO error handling code that saves the table.

It had two purposes.

One was dealing with the (common) situation where a file that was thought to be paired, wasn't; when this occurred in the middle of a bunch of files, a lot of time and work was wasted. We're dealing with that now by putting in broken-paired reading, which deals much more nicely with the overall problem of keeping pairs.

The other (very infrequent/specific) problem was actual IO errors, caused by HPC hiccoughs that @camillescott was running into.  I'm not actually convinced that this really dealt with that problem but maybe it did.

In any case, there is simply *no* use for a hashtable generated by diginorm that only includes reads partway through a file, and there's no way to roll back the hashtable currently (or probably ever :). The best thing to do is simply (a) not save the hashtable and (b) delete the partial .keep file that remains.  Diginorm (or load-into-counting) can be run on the remaining keep files if the run is to be restarted.  If someone is normalizing lots of files and running into hardware errors, the proper thing to do anyway is normalize the files one by one, and then a subset together, etc. etc.

So, @bocajnotnef, please remove the htable saving and the associated tests :)

bocajnotnef:
this is reminding me rather a lot of the original ReadParser debacle. 

@ctb Do you mean the CheckpointSavingTable or the CatchIOErrors? Or both? Sounds more like CheckpointSaving but you were talking about IOErrors. Sooo. Yeah.

ctb:
Checkpoint saving 

--

bocajnotnef:
Alrighty.

Wooo removing code.

camillescott:
@ctb @bocajnotnef i've got something that will make him even more angry....

Basically, while what you've got cleans things up, there's still a lot of confusing structure -- for example, `WithDiagnostics` twiddling variables in the normalizer object. I propose making use of generators and decorators to streamline things. Here's a skeleton of what it might look like:

```
def normalize_by_median(reader):
    
    def new_reader(input_file, **kwargs):
        for index, paired, r0, r1 in reader(input_file):
            if pass_filter:
                yield index, paired, r0, r1
    
    return new_reader
    
def diagnostics(filter_gen):
    
    def filter_w_diagnostics(input_file, report_fn, **kwargs):
        
        total = 0
        kept = 0
        
        for index, paired, r0, r1 in filter_gen(input_file, **kwargs):
            if paired:
                kept += 2
            else:
                kept += 1
            
            total = index
            
            if index > 0 and index % 100000 == 0:
                # print diagnostics
                pass
            
            yield index, paired, r0, r1
        
        if total == 0:
            print('SKIPPED empty file ' + input_file, file=sys.stderr)
        else:
            # print the things
            # report the things
            pass
    
    return filter_w_diagnostics

if args.action == 'normalize':
    filter_gen = normalize_by_median(broken_paired_reader)
elif args.action == 'filter_abund':
    ...
    
if args.print_diagnostics:
    filter_gen = diagnosics(filter_gen, args.report_fn)

for index, paired, r0, r1 in filter_gen:
    write(paired, r0, r1)
```

This way, any filtering function can be given to the diagnostics function, and we can build up an appropriate generator based on user input.

Feel free to groan loudly :)

ctb:
Ah, yes, very nice!  but I think that's another PR after this one.

@bocajnotnef, get to the point where what you have is mergable, and
then we can talk about next steps on this and other things.

bocajnotnef:
Right. So. Ripping out checkpoint_counting_tables, eliminate relevant
tests, bring up to merge spec and then go into an angsty funk while I
recode the context manages and generators. Cool. This I can do.

bocajnotnef:
Seriously getting massively strong read-parser vibes here.

ctb:
On Thu, Jun 04, 2015 at 04:23:29PM -0700, Jake Fenton wrote:
> Right. So. Ripping out checkpoint_counting_tables, eliminate relevant
> tests, bring up to merge spec and then go into an angsty funk while I
> recode the context manages and generators. Cool. This I can do.

lol

camillescott:
Aww, hey now! Your code actually *works*

On Thu, Jun 4, 2015 at 4:23 PM, Jake Fenton <notifications@github.com>
wrote:

> Seriously getting massively strong read-parser vibes here.
>
>
> On Thu, Jun 4, 2015, 19:23 Jake Fenton <bocajnotnef@gmail.com> wrote:
>
> > Right. So. Ripping out checkpoint_counting_tables, eliminate relevant
> > tests, bring up to merge spec and then go into an angsty funk while I
> > recode the context manages and generators. Cool. This I can do.
> >
> > On Thu, Jun 4, 2015, 19:21 C. Titus Brown <notifications@github.com>
> > wrote:
> >
> >> Ah, yes, very nice! but I think that's another PR after this one.
> >>
> >> @bocajnotnef, get to the point where what you have is mergable, and
> >> then we can talk about next steps on this and other things.
> >>
> >> —
> >> Reply to this email directly or view it on GitHub
> >> <https://github.com/dib-lab/khmer/pull/1057#issuecomment-109090955>.
> >>
> >
>
> —
> <https://github.com/dib-lab/khmer/pull/1057#issuecomment-109091221>.
>



-- 
Camille Scott

Department of Computer Science
Lab for Data Intensive Biology
University of California, Davis

camille.scott.w@gmail.com

ctb:
On Thu, Jun 04, 2015 at 04:27:19PM -0700, Camille Scott wrote:
> Aww, hey now! Your code actually *works*

now, now.

bocajnotnef:
I'm fairly certain there's bits that actually don't work at all but simply
aren't covered by the tests. Specific case: handling "corrupt" files.

bocajnotnef:
@ctb @camillescott CR/merge?

scripts/normalize-by-median.py
ctb: ping @bocajnotnef

bocajnotnef: whoops.

scripts/normalize-by-median.py
```diff
@@ -55,80 +48,94 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
+        print('SKIPPED empty file ' + ifilename, file=sys.stderr)
+    else:
+        print('DONE with {inp}; kept {kept} of {total} or {perc:2}%'
+              .format(inp=ifilename, kept=norm.total - norm.discarded,
+                      total=norm.total, perc=int(100. - norm.discarded /
+                                                 float(norm.total) * 100.)),
+              file=sys.stderr)
 
-        yield batch
+    if fp:
```
ctb: Please de-indent following block.

Also, please use some form of string formatting, either str.format or '%' here -- str.format is fine.  But don't do str(a) + ... + str(b).

scripts/normalize-by-median.py
```diff
@@ -373,28 +312,42 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
+    if args.unpaired_reads:
+        files.append([args.unpaired_reads, False])
+
+    corrupt_files = []
+
+    for f, p in files:
+        outfp = None
+
+        # figure out what we're calling out output file
+        if args.single_output_file:
```
ctb: Please move the bulk of this 'if' outside the for loop and only keep the 'else' as part of:

    if not args.single_output_file: ...

scripts/normalize-by-median.py
```diff
@@ -373,28 +312,42 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
+        else:
+            output_name = os.path.basename(f) + '.keep'
+            outfp = open(output_name, 'w')
+
+        # failsafe context manager in case an input file breaks
+        with CatchIOErrors(f, outfp, args.fail_save, htable, args.force,
+                           corrupt_files):
+
+            # actually do diginorm
+            for record in norm(f, p):
```
ctb: Excellent.  Perhaps one last flipping of classes - or is this getting too close to what @camillescott suggested? :)

* create broken_paired_reader here;
* pass it into WithDiagnostics, along with 'norm';
* wrap norm.__call__ inside of WithDiagnostics;

This should dramatically simplify all the passing around of arguments.

scripts/normalize-by-median.py
```diff
@@ -55,80 +48,94 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
                 for record in batch:
                     seq = record.sequence.replace('N', 'A')
                     self.htable.consume(seq)
                     yield record
             else:
-                self.discarded += batch_size
+                self.discarded += len(batch)
 
 
 def handle_error(error, output_name, input_name, fail_save, htable):
```
ctb: eliminate fail_save args and behavior completely, please.

scripts/normalize-by-median.py
bocajnotnef: Wait, so, move the `for record in norm` into WithDiagnostics?

ctb: try making this work:

            for record in WithDiagnostics(f, report_fp, norm, reader):
                write_record(record, outfp)

bocajnotnef: Right, okay

bocajnotnef:
@ctb Ta-daaaaa.

ctb:
On Fri, Jun 05, 2015 at 12:49:25PM -0700, Jake Fenton wrote:
> @ctb Ta-daaaaa.

you sure you want me to look at it? there's nothing in there that's
embarassing or that you forgot?

scripts/normalize-by-median.py
```diff
@@ -55,166 +48,111 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
-        batch_size = 1
-        if force_paired:
-            batch_size = 2
+        passed_filter = False
 
-        for batch in WithDiagnostics(input_filename, batch_size,
-                                     self.report_fp, force_paired, self):
-            passed_filter = False
-            passed_length = True
+        if is_paired:
```
ctb: I realize I'm just totally nitpicking right now, but how about:

      self.total += 1
      if is_paired:
           self.total += 1

scripts/normalize-by-median.py
```diff
@@ -55,166 +48,111 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
                 print(total + " " + total - discarded + " " +
                       1. - (discarded / float(total)), file=fp)
                 report_fp.flush()
 
-        # If in paired mode, check that the reads are properly interleaved
-        if paired:
-            if not check_is_pair(batch[0], batch[1]):
-                raise IOError('Error: Improperly interleaved pairs \
-                    {b0} {b1}'.format(b0=batch[0].name, b1=batch[1].name))
+        for record in norm(index, is_paired, read0, read1):
```
ctb: how about passing reader into norm?

    for record in norm(reader):
       ...

ctb:
Wow. That turned out much nicer. Good stuff. thx @bocajnotnef. a few comments and rejiggering sugg; let me know.

scripts/normalize-by-median.py
bocajnotnef: Wouldn't that muck up how the generators work?

bocajnotnef:
Every now and then I do something impressive. Just gotta reduce the frequency of embarrassing incidents (a la merge markers in the changelog) and I'm set!

scripts/normalize-by-median.py
ctb: On Fri, Jun 05, 2015 at 12:56:10PM -0700, Jake Fenton wrote:
> >                  print(total + " " + total - discarded + " " +
> >                        1. - (discarded / float(total)), file=fp)
> >                  report_fp.flush()
> >  
> > -        # If in paired mode, check that the reads are properly interleaved
> > -        if paired:
> > -            if not check_is_pair(batch[0], batch[1]):
> > -                raise IOError('Error: Improperly interleaved pairs \
> > -                    {b0} {b1}'.format(b0=batch[0].name, b1=batch[1].name))
> > +        for record in norm(index, is_paired, read0, read1):
> 
> Wouldn't that muck up how the generators work?

nope :)

just have the for loop in WithDiagnostics loop over norm(reader).

bocajnotnef:
@ctb Stuff done, reasonably confident there's nothing embarrassing there.

scripts/normalize-by-median.py
```diff
@@ -55,166 +48,111 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
+        print("{total} {kept} {discarded}"
+              .format(total=norm.total, kept=norm.total - norm.discarded,
+                      discarded=1. - (norm.discarded / float(norm.total))),
+              file=fp)
+        fp.flush()
 
 
 class Normalizer(object):
-    def __init__(self, desired_coverage, htable, report_fp=None):
+    def __init__(self, desired_coverage, htable, report_fp=None,
```
ctb: report_fp is now unused!

scripts/normalize-by-median.py
```diff
@@ -55,166 +48,111 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
                     self.htable.consume(seq)
-                    yield record
+                    yield record, index
             else:
-                self.discarded += batch_size
+                self.discarded += len(batch)
 
 
-def handle_error(error, output_name, input_name, fail_save, htable):
+def handle_error(error, input_name):
```
ctb: handle_error can now be removed and the print incorporated into CatchIOErrors!

scripts/normalize-by-median.py
```diff
@@ -373,28 +293,48 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
+    outfp = None
+
+    if args.single_output_file:
+        if args.single_output_file is sys.stdout:
+            output_name = '/dev/stdout'
+        else:
+            output_name = args.single_output_file.name
+        outfp = args.single_output_file
+
+    for f, p in files:
```
ctb: let's re-expand these to 'filename, require_paired'.

scripts/normalize-by-median.py
```diff
@@ -373,28 +293,48 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
...
-        outfp = open(output_name, 'w')
-        total_acc, discarded_acc, corrupt = \
-            normalize_by_median_and_check(
-                args.unpaired_reads, htable, args.single_output_file,
-                args.fail_save, args.paired, args.force, norm, report_fp)
-        corrupt_files += corrupt
+            output_name = os.path.basename(f) + '.keep'
+            outfp = open(output_name, 'w')
+
+        ksize = htable.ksize()
```
ctb: This can be moved up to where we create/load htable; not needed inside the loop.

ctb:
Do we have a test for corrupt files being removed yet?

scripts/normalize-by-median.py
```diff
@@ -55,166 +48,111 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
                 if med < desired_coverage:
                     passed_filter = True
 
-            if passed_length and passed_filter:
+            if passed_filter:
                 for record in batch:
                     seq = record.sequence.replace('N', 'A')
                     self.htable.consume(seq)
-                    yield record
+                    yield record, index
```
ctb: would suggest removing index (which is not incremented here) and using:

    for index, record in enumerate(norm(reader)):

which is a standard Python pattern.

scripts/normalize-by-median.py
```diff
@@ -55,166 +48,111 @@ def WithDiagnostics(ifile, batch_size, fp, paired, norm):
...
         self.total = 0
         self.discarded = 0
-        self.corrupt_files = []
 
-    def __call__(self, input_filename, force_paired=False):
-        seq = ""
+    def __call__(self, reader):
 
         desired_coverage = self.desired_coverage
         ksize = self.htable.ksize()
```
ctb: note, ksize not used in this function.

ctb:
Reminders & add'l requests:

* remove dump_frequency
* remove fault_tolerant and put -f in force
* remove --report-total-kmers and have this automatically happen (i.e. no 'if args.report_total_kmers')
* recheck docstrings, hint hint.

But should be ready to merge after that :)

scripts/normalize-by-median.py
```diff
@@ -33,188 +33,120 @@
...
+                          total=norm.total,
+                          perc=int(100. - norm.discarded /
+                                   float(norm.total) * 100.)),
                   file=sys.stderr)
 
-            print('... in file ' + input_filename, file=sys.stderr)
+            print('... in file ' + ifilename, file=sys.stderr)
 
-            if report_fp:
+            if fp:
```
bocajnotnef: We do this check twice within WithDiagnostics--Redundant?

bocajnotnef:
@ctb updated minus tests

For the tests--Am I frankenstining a file with mixed paired and unpaired or am I just passing in various input files that bounce between paired and unpaired?

ctb:
@bocajnotnef I'll take a look, but what you should ALWAYS do when you're trying to figure out what to test is look at the results of 'make diff-cover', which here tell me:

    khmer/utils.py (71.4%): Missing lines 122,148
    scripts/normalize-by-median.py (93.3%): Missing lines 62,64,67,299

So those would be your first targets to test.

bocajnotnef:
Ahh, alright. I'll hack on that tonight.

ctb:
- [ ] remove comment about batchwise (lines 39-40)
- [ ] remove docstring comment about WithDiagnostics re "checking for properly paired data"
- [ ] Normalizer __init__ no longer needs force_single arg
- [ ] Normalizer creation shouldn't take report_fp (note, this is a bug, actually)
- [ ] curious, what happens if you feed normalize-by-median.py a non-FASTA/FASTQ file?
- [ ] if an input file is corrupt, you should remove the corresponding .keep file

Otherwise looks good to me!

bocajnotnef:
re: non-FASTA/FASTQ:

```
(env)bocajnotnef@GCU-Caconym:~/git/khmer⟫ normalize-by-median.py scripts/normalize-by-median.py 

|| This is the script 'normalize-by-median.py' in khmer.
|| You are running khmer version 1.4+117.g6957bb9.dirty
|| You are also using screed version 0.8
||
|| If you use this script in a publication, please cite EACH of the following:
||
||   * MR Crusoe et al., 2014. http://dx.doi.org/10.6084/m9.figshare.979190
||   * CT Brown et al., arXiv:1203.4802 [q-bio.GN]
||
|| Please see http://khmer.readthedocs.org/en/latest/citations.html for details.


PARAMETERS:
 - kmer size =    32            (-k)
 - n tables =     4             (-N)
 - min tablesize = 1e+06        (-x)

Estimated memory usage is 4e+06 bytes (n_tables x min_tablesize)
--------
making k-mer counting table
** ERROR: unknown file format for 'scripts/normalize-by-median.py'
** Failed on scripts/normalize-by-median.py: 
** Exiting!
```

Which I believe is @mr-c's magic sniffing detection.

ctb:
@bocajnotnef LOL that's because you fixed the bug I was pointing you at before running it (fixed in 8073421). Sneaky of you ;)

scripts/normalize-by-median.py
```diff
@@ -33,188 +33,114 @@
...
-        norm.total += batch_size
-        total = norm.total
-        discarded = norm.discarded
-
-        if index > 0 and index % 100000 == 0:
-            print('... kept {kept} of {total} or {perc:2}%'
-                  .format(kept=total - discarded,
-                          total=total,
-                          perc=int(100. - discarded / float(total) * 100.)),
+        if norm.total > 0 and norm.total % 10000 == 0:
```
bocajnotnef: @ctb Thoughts on this? I was finding it difficult to get 100000 kept reads with our testing data but I also thought it made more sense to report based on total reads gone through and not the number kept.

camillescott: My thoughts: 10,000 is too small, and will result in a stupendous quantity of output for median-large datasets; reporting based on total is the way to go, because it guarantees output at a steady interval.

bocajnotnef: Mm, it is rather small. 

New problem: How would I test that line without having a truly massive dataset?

Or, do I get a truly massive dataset?

camillescott: This'll be covered by the test for expected output that is being merged
with #862

On Sun, Jun 7, 2015 at 8:09 PM, Jake Fenton <notifications@github.com>
wrote:

> In scripts/normalize-by-median.py
> <https://github.com/dib-lab/khmer/pull/1057#discussion_r31884284>:
>
> >
> > -        norm.total += batch_size
> > -        total = norm.total
> > -        discarded = norm.discarded
> > -
> > -        if index > 0 and index % 100000 == 0:
> > -            print('... kept {kept} of {total} or {perc:2}%'
> > -                  .format(kept=total - discarded,
> > -                          total=total,
> > -                          perc=int(100. - discarded / float(total) * 100.)),
> > +        if norm.total > 0 and norm.total % 10000 == 0:
>
> Mm, it is rather small.
>
> New problem: How would I test that line without having a truly massive
> dataset?
>
> Or, do I get a truly massive dataset?
>
> —
> <https://github.com/dib-lab/khmer/pull/1057/files#r31884284>.
>



-- 
Camille Scott

Department of Computer Science
Lab for Data Intensive Biology
University of California, Davis

camille.scott.w@gmail.com

bocajnotnef: Ahh, excellent. I don't have to worry about it, then.

bocajnotnef:
@ctb Updated--Everything is covered except a thing that @camillescott said would be covered in #862. I made a behavior change in the incremental reporting; instead of every 100000 reads returned by diginorm it'd be every 100000 reads sent into diginorm. (see here: https://github.com/dib-lab/khmer/pull/1057/files#diff-6b3b058584d44911b3cf3fe21d9c9af4R51)

Other 'n that, should be mergable.

scripts/normalize-by-median.py
camillescott: Note: seeing as you've implemented a specific diagnostics handler, you
*could* parameterize this... :)

bocajnotnef: Shh, noo, I just finished getting rid of a bunch of args.

(Fair point, though... Personally, I vote to wait until I move on to the
refactor structure you suggested.)

camillescott: Yeah, I agree. Let's get this one merged first.

khmer/utils.py
```diff
@@ -130,6 +134,8 @@ def broken_paired_reader(screed_iter, min_length=None, force_single=False):
                 n += 2
                 record = None
             else:                                   # orphan.
+                if require_paired:
+                    raise ValueError("Unpaired reads when force-paired is set")
```
ctb: s/force-paired/require_paired/, pls

khmer/utils.py
```diff
@@ -138,6 +144,8 @@ def broken_paired_reader(screed_iter, min_length=None, force_single=False):
 
     # handle the last record, if it exists (i.e. last two records not a pair)
     if prev_record:
+        if require_paired:
+            raise ValueError("Last two records weren't paired!")
```
ctb: same error message as the other one, please!

tests/test_scripts.py
```diff
@@ -601,6 +602,94 @@ def test_normalize_by_median():
...
+    in_dir = os.path.dirname(infile)
+
+    shutil.copyfile(utils.get_test_data('single-read.fq'), infile)
+
+    script = scriptpath('normalize-by-median.py')
+    args = ['-C', CUTOFF, '-k', '17', '-p',  infile]
+    try:
+        (status, out, err) = utils.runscript(script, args, in_dir)
+        raise Exception("Shouldn't get to this")
+    except:
```
ctb: This will catch Exception as well as AssertionError (which is what is thrown by runscript). Suggest catching AssertionError specifically.

tests/test_scripts.py
```diff
@@ -601,6 +602,94 @@ def test_normalize_by_median():
...
+
+    shutil.copyfile(utils.get_test_data('test-large.fa'), infile)
+
+    script = scriptpath('normalize-by-median.py')
+    args = ['-C', '1', '-k', '17', '--force-single', '-p', '-R',
+            outfile, infile]
+    try:
+        (status, out, err) = utils.runscript(script, args, in_dir)
+        raise Exception("Shouldn't get to this")
+    except:
```
ctb: Same problem: don't catch _all_ exceptions...

ctb:
- [ ] write test for -o /dev/stdout

ctb:
Otherwise LGTM! Nice work!

bocajnotnef:
in your pull, you changed `args.single_output_file` to `args.single_output_file.name` and stuff's breakin' everywhere.

I'm reverting it to just `single_output_file`, 'cause argparse is supposed to open the given string anyway, turning `single_output_file` into a file object which may or may not be the `stdout` file, which is a `sys` attribute.

ctb:
OK, but what I'm saying is it wasn't working in the tests :).

bocajnotnef:
Right, I went back and actually read things and now understand. Gah. 

Context, it's important.

ctb:
Well, I mean, I'm wrong a lot. So always double check ;)

bocajnotnef:
Note: the test `normalize_by_median_stdout` is simply named wrong--it's a streaming test, not a stdout test.

bocajnotnef:
Also, I'm running into all kinds of issues with removing the `if args.single_output_file` block--if `/dev/stdout` or any of it's aliases are passed in then `single_output_file` becomes a CString and not a file object and thus doesn't have a `.name` attribute.

bocajnotnef:
retest this please, Jenkins

bocajnotnef:
@ctb Additional opinions, please.

ctb:
Huh, the tests aren't passing on my laptop!?

bocajnotnef:
Have you been doing strange things with your virtualenv? (i.e. testing screed RCs?)

ctb:
it doesn't seem screed related.

bocajnotnef:
Well. They seem to be passing on my laptop, Jenkins and Athyra.

Would try a clean env--if that doesn't do it, might be an OSX specific issue (which CI won't catch 'cause the OSX host is still offline)

ctb:
Yep passing on athyra. How odd.

ctb:
Python version issue; works on Python 2.7.6, not on Python 2.7.2.

bocajnotnef:
Strange. Shall I investigate more?

bocajnotnef:
@ctb Merged in fix.

ctb:
OK, will review.

ctb:
One minor problem is that tests.test_scripts.test_normalize_by_median_stdout_3 and tests.test_scripts.test_normalize_by_median_stdout_2 leak test output. Can't think of a good way to fix that tho.

ctb:
OK: consider what happens if the following is run:

    normalize-by-median.py -o xxx data/100k-filtered.fa  data/100k-filtered.fa 

at least two undesirable things happen, one actually destructive.  What are they, @bocajnotnef?

ctb:
...and CAN YOU FIX THEM? WILL YOU ACCEPT THE CHALLENGE??

ctb:
Actually, *three* undesirable things happen, which has to do with the tests leaking. So you can fix everything in one swell foop.

bocajnotnef:
0. the "are there multiple inputs with the same name" check will fail since it doesn't pay attention to the `single_output_file` flag
 1. It doesn't check to see if the *same file* is entered into the program, regardless of its basename
 2. the broken multiple inputs with the same name check doesn't do anything to prevent overwriting the output of the first file with whichever name--i.e. if we have `normalize-by-median.py data1/100k-filtered.fa  data2/100k-filtered.fa` the out from `data2` will overwrite that of `data1`--that said, this doesn't involve the `-o`flag
 3. for some reason, if we have `single_output_file` set we don't just point `outfp` at it, we open another file handle to the new file. That's silly.

I'm guessing we somehow duplicate sending output to stdout because of the weird logic/`outfp` direction in the `if args.single_output_file` block.

bocajnotnef:
(and of course I'll accept the challenge--I'm getting paid to solve it.)

bocajnotnef:
Wait, no--because if we do `-o /dev/stdout' for some reason the argparser doesn't make that a file handle and just returns a string.

What's up with that?

ctb:
Might it be a string... What does 'type' say?

--

bocajnotnef:
Nvm. tests passing. 

Gonna steam of consiousness this in the gitter channel while I hack on it. I'll @mention you here when I'm done.

ctb:
Suggest cleaning up the non stdout case first. Suspect rest will be resolved easily after that.

--

bocajnotnef:
@ctb Fixed some other things that looked super bad but I can't figure out why the tests are leaking--figuring it has something to do with '-' vs '/dev/stdout' but can't fathom what.

ctb:
- [ ] switch to median_at_least from #862

ctb:
You can just remove test_normalize_by_median_stdout_2; it's not testing the 'sys.stdout' branch of that if statement, so it's not a useful test on top of  test_normalize_by_median_stdout_3.

ctb:
@bocajnotnef, if you can clean up that if statement and remove that test, I'll merge.

tests/test_scripts.py
```diff
@@ -15,6 +15,7 @@
 from cStringIO import StringIO
 import traceback
 from nose.plugins.attrib import attr
+from nose.tools import assert_raises
```
mr-c: Is this used anymore?

bocajnotnef: Indeed not.

mr-c:
Please wait for #1031 to get merged first. After that you can update `doc/whats-new-2.0.rst` with the changes to the CLI.

bocajnotnef:
Wilco.

ctb:
@bocajnotnef get it mergable and we'll see. There's no reason to wait on #1031 to finish this PR off.

bocajnotnef:
Mkay. Will do so tonight.

bocajnotnef:
@ctb Merge, please?

ctb:
Whew. Thanks, @bocajnotnef!

bocajnotnef:
Huzzah!

Wondering: What was the "one fell swoop" solution?

ctb:
On Tue, Jun 09, 2015 at 12:45:12PM -0700, Jake Fenton wrote:
> Wondering: What was the "one fell swoop" solution?

All of those problems were related; you got to it eventually :)

bocajnotnef:
Hmm. Thought the test was still leaking, which is what confused me.. Oh well.

ctb:
On Tue, Jun 09, 2015 at 12:46:36PM -0700, Jake Fenton wrote:
> Hmm. Thought the test was still leaking, which is what confused me.. Oh well.

It's not, at least not on Mac OS X.

Issue #1027, opened 2015-05-25
interleave-reads.py SRA naming bug


macmanes:
root@ip-172-31-27-66:/mnt/reads# head -4 test1.fq
    @SRR797058.3 HWI-ST600:227:C0WR4ACXX:7:1101:17167:2000/1
    NCTACCAAAAAAATGCCCGATAATTCTGACCATTCCTTCCTCATTCTCGTCTGGCGTTTGGTCACGACGCACGATACCTTCTGCACTTGTCAAGACAGCGG
    +
    #00@#################################################################################################

and 

    root@ip-172-31-27-66:/mnt/reads# head -4 test2.fq
    @SRR797058.3 HWI-ST600:227:C0WR4ACXX:7:1101:17167:2000/2
    CTTGACAAGAGCAGAAGTTATCTTGCCTCGGGACCAAACGCCAGACGAGCACGAGGGAGCGATCGTCCGCATTAGCCGGCATTCTTTTGCTAGCAGATCGG
    +
    =?###################################################################################################

attempt at interleaving.

    root@ip-172-31-27-66:/mnt/reads# interleave-reads.py test1.fq test2.fq -o inter.fq
    
    || This is the script 'interleave-reads.py' in khmer.
    || You are running khmer version 1.4
    || You are also using screed version 0.8
    ||
    || If you use this script in a publication, please cite EACH of the following:
    ||
    ||   * MR Crusoe et al., 2014. http://dx.doi.org/10.6084/m9.figshare.979190
    ||
    || Please see http://khmer.readthedocs.org/en/latest/citations.html for details.
    
    Interleaving:
            test1.fq
            test2.fq
    ... 0 pairs
    ERROR: This doesn't look like paired data! SRR797058.3 HWI-ST600:227:C0WR4ACXX:7:1101:17167:2000/1/1 SRR797058.3 HWI-ST600:227:C0WR4ACXX:7:1101:17167:2000/2/2

Seems like an extra `/1` or `/2` is being appended.

macmanes:
Apparently the space in the fastQ header is tripping things up, as a simple sed command removing the space makes the interleaving process work.

`sed -i 's_3 H_3-H_g' test{1,2}.fq`

mr-c:
Drat. Thanks for the report

ctb:
Not yet fixed - see comments in #1088.  The right-hand side needs to be checked properly.

kdmurray91:
PR to fix is #1173

[...]

PR #1088, opened 2015-06-12
Fix bug in interleave-reads.py mentioned in #1027

SherineAwad:
Fix bug mentioned here:
https://github.com/dib-lab/khmer/issues/1027


SherineAwad:
@ctb  please?
check_is_left() is called in interleave-reads.py, if we have: 
name='SRR797058.3 HWI-ST600:227:C0WR4ACXX:7:1101:17167:2000/1'
and we call:
lhs, rhs = _split_left_right(name)
it returns: lhs = 'SRR797058.3' and  rhs = 'HWI-ST600:227:C0WR4ACXX:7:1101:17167:2000/1'
so  check_is_left(name) will return false because the lhs doesn’t end with ‘/1' and rhs doesn't start with ‘:1'. 
Same problem with check_is_right().
should I fix this? and in this PR?

SherineAwad:
@ctb 
same issue with check-is-pair?
Isn't left as plain english? So check_is_left should check rhs ends with/1? ..etc 
Changing these functions will miss up many tests?

SherineAwad:
@ctb 
Ignore my second comment please. Just wanna know where should I handle the different format. In interleave-reads or utils.py?

SherineAwad:
Jenkins test this please

SherineAwad:
Ready for review
- [x] Is it mergable?
- [x] Did it pass the tests?
  Check for code coverage.
  fixing as needed.
  changes were made?

ctb:
Hi @SherineAwad could you please add a test before asking for review? Thx.

SherineAwad:
Oh, I see. Will do !

SherineAwad:
= Added some test files!  Do you require test files be created on the fly? 
Other than this; Ready for review !

ctb:
@SherineAwad don't have time to review today, but looks good on a quick skim!

one question - why the file mode change?

(scripts/interleave-reads.py 100755 → 100644)

please change it back :)

ctb:
On Sat, Jun 13, 2015 at 02:58:18PM -0700, Sherine wrote:
> = Added some test files!  Do you require test files be created on the fly? 
> Other than this; Ready for review !

(no need for test files to be created on fly :)

ctb:
Thanks, @SherineAwad! Looks good.

bede:
I've just run into this bug analysing some ENA data. Thanks for the fix @SherineAwad

khmer/utils.py
```diff
@@ -55,6 +55,10 @@ def check_is_pair(record1, record2):
     elif lhs1 == lhs2 and rhs1.startswith('1:') and rhs2.startswith('2:'):
         return True
 
+    # handle @name seq/1
+    elif lhs1 == lhs2 and rhs1.endswith('/1') and rhs2.endswith('/2'):
```
kdmurray91: Do we need to check that `rhs1[:-2] == rhs2[:-2]`? I.e., that the read name is equal.

The same check as L47-52 above

kdmurray91: Because `lhs1 ==lhs2` will always be true, even if pairing is off, as that's the sample ID of the file.

Also, sorry I didn't look at this earlier :)

ctb: argh, yes, we need to do that.

kdmurray91: OK, I'll make a PR

kdmurray91: See PR #1173

PR #1090, opened 2015-06-12
Adapt Makefile rules for py3 changes

luizirber:
# WIP

Makefile rules were outdated, trying to fix it. (see inline comments)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@
 #  and documentation
 # make coverage-report to check coverage of the python scripts by the tests
 
-CPPSOURCES=$(wildcard lib/*.cc lib/*.hh khmer/_khmermodule.cc)
+CPPSOURCES=$(wildcard lib/*.cc lib/*.hh khmer/_khmer.cc)
 PYSOURCES=$(wildcard khmer/*.py scripts/*.py)
 SOURCES=$(PYSOURCES) $(CPPSOURCES) setup.py
 DEVPKGS=sphinxcontrib-autoprogram pep8==1.5.7 diff_cover \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -39,9 +39,9 @@ install-dependencies:
 	pip install --upgrade $(DEVPKGS)
 
 ## sharedobj   : build khmer shared object file
-sharedobj: khmer/_khmermodule.so
+sharedobj: khmer/_khmer.so
```
luizirber: This is not true for py3, since it creates the dynamic library as `_khmer.cpython-34m.so`, for example.

Makefile
```diff
@@ -154,7 +154,7 @@ diff_pylint_report: pylint_report.txt
 # We need to get coverage to look at our scripts. Since they aren't in a
 # python module we can't tell nosetests to look for them (via an import
 # statement). So we run nose inside of coverage.
-.coverage: $(PYSOURCES) $(wildcard tests/*.py) khmer/_khmermodule.so
+.coverage: $(PYSOURCES) $(wildcard tests/*.py) $(wildcard khmer/_khmer*.so)
```
luizirber: This can be improved. We can create separate coverage runs and combine the results for the complete coverage.

luizirber:
@mr-c ideas?

mr-c:
Make use of http://stackoverflow.com/questions/4933285/how-to-detemine-python-version-in-makefile ?

mr-c:
@luizirber ping?

luizirber:
@mr-c I will add these changes to #1108

PR #1111, opened 2015-06-22
tox & Travis CI support

mr-c:


mr-c:
To run tox locally I had to deactivate the current virtualenv and make sure I started with a clean build tree

.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: python
+python: 2.7
+
```
luizirber: Since we don't need sudo, we can set
`sudo: false`
and use the new Docker infrastructure (which is way faster).

More info: http://cdunn2001.blogspot.com/2015/06/using-c11-in-travis-ci.html (this also show how to install packages in Ubuntu without using sudo)

Issue #1120, opened 2015-06-24
Allow for skipping of pysam tests


ctb:
Also see https://github.com/dib-lab/khmer/pull/1019#discussion_r30709115

cc @bocajnotnef

bocajnotnef:
Any way we could build a 

```
try:
    import
except:
    # skip test
```

kinda thing?

ctb:
+1

ctb:
Will be fixed by #750 (see #1121).

mr-c:
Done

PR #1121, opened 2015-06-25
Implemented pysam skipability

bocajnotnef:
issue: #1120
starting PR comment: https://github.com/dib-lab/khmer/pull/1019#discussion_r30709115

added stuff within the import tests to skip the test (or really just prevent it from failing) if the pysam module is not installed.

@ctb


bocajnotnef:
relevant to #750

ctb:
Ahh, apologies, I didn't realize that it was a pysam import error in the test_import_all test.  But...

This is an ugly and rather overbroad hack :). We don't want to check for failed pysam import in *generic* test running code!

Instead, I suggest moving 'import pysam' into the 'main()' function, and if/when we write an actual exec test for the script, we can make that test skip when pysam import fails.

bocajnotnef:
Agreed, it is an ugly hack. 

I'll go move the things.

bocajnotnef:
An embarassing number of commits later, tests are passing. 

@ctb Better?

sandbox/readaligner_pairhmm_train.py
```diff
@@ -28,6 +27,11 @@ def trusted_str(cov, trusted_cutoff):
 
 
 def main():
+    try:
```
ctb: why try/pass here? just import it. tests should still succeed as we don't actually run this script in the tests.

bocajnotnef: I swear I'd dropped the try/except... Gah.

bocajnotnef: Aha, I see why I had that.

If the import fails the exec test fails:

```
test_sandbox_scripts.py: test import readaligner_pairhmm_train.py ... FAIL

======================================================================
FAIL: test_sandbox_scripts.py: test import readaligner_pairhmm_train.py
----------------------------------------------------------------------
  File "/home/bocajnotnef/git/env/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/bocajnotnef/git/khmer/tests/test_sandbox_scripts.py", line 83, in __call__
    "{0}".format(traceback))
AssertionError: (u"/home/bocajnotnef/git/khmer/sandbox/readaligner_pairhmm_train.py cannot be exec'd", u"<module 'traceback' from '/usr/lib/python2.7/traceback.pyc'>")
```

ctb: OK, this is a bit of a hack, but if you move the import down below parse_args, you'll be fine.

bocajnotnef:
@ctb Updated

sandbox/readaligner_pairhmm_train.py
```diff
@@ -28,6 +27,7 @@ def trusted_str(cov, trusted_cutoff):
 
 
 def main():
+
```
ctb: extraneous blank line -> pls remove

ctb:
thx!

sandbox/readaligner_pairhmm_train.py
```diff
@@ -37,6 +36,8 @@ def main():
 
     args = parser.parse_args()
 
+    import pysamFOO
```
mr-c: ??? No, this does not work. Why was this merged?

ctb: Hah, I guess that's the problem with allowing it to be skipped!

mr-c: I thought we found a Python only sam parser? All I can come up with is
https://pypi.python.org/pypi/libsam/ which doesn't fill me with confidence

On Tue, Jul 28, 2015 at 11:00 AM C. Titus Brown <notifications@github.com>
wrote:

> In sandbox/readaligner_pairhmm_train.py
> <https://github.com/dib-lab/khmer/pull/1121#discussion_r35678960>:
>
> > @@ -37,6 +36,8 @@ def main():
> >
> >      args = parser.parse_args()
> >
> > +    import pysamFOO
>
> Hah, I guess that's the problem with allowing it to be skipped!
>
> —
> <https://github.com/dib-lab/khmer/pull/1121/files#r35678960>.
>
-- 

ctb: From https://twitter.com/ctitusbrown/status/626093435817193472,

https://github.com/mdshw5/simplesam
https://github.com/brentp/hts-python/

Issue #1125, opened 2015-06-29
some tests leak stdout, which is Bad.


ctb:
```
tests.test_script_arguments.test_create_countgraph_3 ...
** ERROR: khmer only supports k-mer sizes <= 32.

ok
```

and

```
tests.test_script_arguments.test_create_nodegraph_3 ...
** ERROR: khmer only supports k-mer sizes <= 32.

ok
```

and

```
tests.test_functions.test_check_file_status_kfile ... ERROR: Input file /var/folders/y6/s1r3q87933s6gjdrtmts1npc0000gq/T/khmertest_OSEr6M/thisfiledoesnotexist does not exist
Exiting
ok
```

and

```
tests.test_functions.test_check_file_status_kfile_force ... ERROR: Input file /var/folders/y6/s1r3q87933s6gjdrtmts1npc0000gq/T/khmertest_qMiKqT/thisfiledoesnotexist does not exist
ok
```

ctb:
(yes, yes, mea culpa. but I didn't want to deal with it at the time :)

bocajnotnef:
Ooh, low hanging fruit.

I'll nab this. (Mostly because that leak is annoying the crap outta me.)

bocajnotnef:
Additionally, this test is leaking into the main directory (~/git/khmer):
`tests.test_scripts.test_extract_paired_reads_4_output_files ... FAIL`

bocajnotnef:
Hmm. Strike my last, doesn't appear to be leaking anymore.. Weird.

bocajnotnef:
Closing #1125, fixed by #1137

Issue #1129, opened 2015-06-29
There are two test_diginorm_basic_functionality_4()'s in the diginorm test suite


bocajnotnef:
https://github.com/dib-lab/khmer/blob/master/tests/test_normalize_by_median.py#L569
https://github.com/dib-lab/khmer/blob/master/tests/test_normalize_by_median.py#L600

Fix: in the second one, change the "4" to a "5"

bocajnotnef:
Closing #1129, fixed by #1137

PR #1137, opened 2015-07-01
Fixed tests output leak

bocajnotnef:
see https://github.com/dib-lab/khmer/issues/1125

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -3,6 +3,7 @@
    * khmer/khmer_args.py: Changed some error printing to be exceptions
    * tests/{test_script_arguments,test_functions}.py: changed tests to expect
    exceptions and/or to redirect stderr where appropriate.
+   * tests/test_normalize_by_median.py: changed to not duplicate a test
 
 2015-06-30  Titus Brown  <titus@idyll.org>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2015-06-30  Jacob Fenton  <bocajnotnef@gmail.com>
+
+   * khmer/khmer_args.py: Changed some error printing to be exceptions
+   * tests/{test_script_arguments,test_functions}.py: changed tests to expect
+   exceptions and/or to redirect stderr where appropriate.
+
 2015-06-30  Titus Brown  <titus@idyll.org>
 
    * setup.cfg: changed 'stop=TRUE' to 'stop=FALSE', so that tests do not
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,9 +1,9 @@
 2015-06-30  Jacob Fenton  <bocajnotnef@gmail.com>
 
-   * khmer/khmer_args.py: Changed some error printing to be exceptions
-   * tests/{test_script_arguments,test_functions}.py: changed tests to expect
-   exceptions and/or to redirect stderr where appropriate.
+   * tests/{test_script_arguments,test_functions}.py: changed tests to use
+   stderr redirection to prevent leaks
    * tests/test_normalize_by_median.py: changed to not duplicate a test
+   * tests/test_script_arguments.py: changed tests to use stderr redirection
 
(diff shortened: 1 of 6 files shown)


bocajnotnef:
@ctb @luizirber CR/merge, please?

bocajnotnef:
also https://github.com/dib-lab/khmer/issues/1129

tests/test_functions.py
```diff
@@ -153,19 +154,30 @@ def test_extract_hashbits_info():
...
+    old_stderr = sys.stderr
+    sys.stderr = capture = StringIO()
+    
     try:
         check_input_files(fn, False)
     except SystemExit:
-        check_file_status_exited = True
-    assert check_file_status_exited
-
+        sys.stderr = old_stderr
```
ctb: I would suggest using try/finally here:

```
try:
   ...
except:
   ...
finally:
   sys.stderr = old_stderr

File: tests/test_functions.py [hunk 2]
```diff
@@ -153,19 +154,30 @@ def test_extract_hashbits_info():
...
     fn = utils.get_temp_filename('thisfiledoesnotexist')
+    
+    old_stderr = sys.stderr
+    sys.stderr = capture = StringIO()
+
     try:
         check_input_files(fn, True)
     except OSError:
         assert False
+
```
ctb: and use `finally` here, too.

tests/test_normalize_by_median.py
```diff
@@ -581,7 +580,8 @@ def test_diginorm_basic_functionality_4():
                     infile)
 
     script = 'normalize-by-median.py'
-    args = list(CUTOFF) + list(PAIRING) + ['-k', '15', infile]
+    args = list(CUTOFF) + ['-k', '15', infile]
```
ctb: why this change?

khmer/khmer_args.py
```diff
@@ -151,8 +151,7 @@ def create_nodegraph(args, ksize=None, multiplier=1.0):
     if ksize is None:
         ksize = args.ksize
     if ksize > 32:
-        print_error("\n** ERROR: khmer only supports k-mer sizes <= 32.\n")
-        sys.exit(1)
+        raise Exception("** ERROR: khmer only supports k-mer sizes <= 32.")
```
ctb: ok, but I *wanted* it to be a print_error, not an Exception...

bocajnotnef: Ahh, sorry. I'll switch it back to the old behavior and use pipe redirection to catch the errs.

tests/test_normalize_by_median.py
bocajnotnef: requiring pairing error'd out--there's unpaired reads in the input (which makes sense since -p used to be used to include single sets of paired/unpaired stuff and this test was never actually running, so it didn't break when we made the change within NBM)

bocajnotnef:
@ctb Updated

bocajnotnef:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

bocajnotnef:
Weird, this shouldn't be mergable as it is...

tests/test_normalize_by_median.py
ctb: ok thx, good catch :)

ChangeLog
```diff
@@ -1,3 +1,10 @@
+2015-06-30  Jacob Fenton  <bocajnotnef@gmail.com>
+
+   * khmer/khmer_args.py: Changed some error printing to be exceptions
```
ctb: Fix ChangeLog to reflect new commits, pls!

ctb:
Apart from the ChangeLog fix, this LGTM.  Please go ahead and merge after updating the changelog.

ctb:
And thanks for the pep8 fixes...!

bocajnotnef:
Merged #1137

PR #1138, opened 2015-07-01
Clean up & simplify trim-low-abund by using generators

ctb:
See #1010 (normalize by median refactor) for inspiration.


ctb:
@bocajnotnef your input welcome!

bocajnotnef:
https://github.com/dib-lab/khmer/blob/cleanup/semistreaming/scripts/trim-low-abund.py#L254 and
https://github.com/dib-lab/khmer/blob/cleanup/semistreaming/scripts/trim-low-abund.py#L299 are essentially the same thing that relies mostly on data pulled from the trimmer--would be nice to integrate it into `Trimmer` somehow.

ctb:
I think most of what's left to do is:

- [ ] update ChangeLog
- [ ] test code coverage
- [ ] add help text for new arguments

ctb:
Ready for review & maybe merge @luizirber @camillescott

scripts/trim-low-abund.py
```diff
@@ -63,18 +63,9 @@
                          check_valid_file_exists, add_output_compression_type,
                          get_file_writer)
 
-DEFAULT_NORMALIZE_LIMIT = 20
+DEFAULT_TRIM_AT_COVERAGE = 20
 DEFAULT_CUTOFF = 2
```
camillescott: Just a thought, but considering the emphasis we put on our our default settings, it might be valuable to keep these in their own file in the main lib, like `defaults.py` or some such, where they can be imported and interrogated.

scripts/trim-low-abund.py
```diff
@@ -134,12 +125,190 @@ def get_parser():
...
+                        default=DEFAULT_DIGINORM_COVERAGE,
+                        help="Coverage threshold for --diginorm")
+    parser.add_argument('--single-pass', default=False, action='store_true',
+                        help="Do not do a second pass across the low coverage "
+                        "data")
+
     return parser
 
 
+class ReadBundle(object):
```
camillescott: This should also probably be in the main lib

scripts/trim-low-abund.py
ctb: I think it's good to have them local to the script, for comprehension reasons.
Maybe if we make absolutely sure to print them out and indicate where they
are set...

scripts/trim-low-abund.py
ctb: At the moment it's not used by any other script so that seemed premature
to me.

scripts/trim-low-abund.py
```diff
@@ -175,122 +364,85 @@ def main():
...
-
     tempdir = tempfile.mkdtemp('khmer', 'tmp', args.tempdir)
     print('created temporary directory %s; '
           'use -T to change location' % tempdir, file=sys.stderr)
 
+    trimmer = Trimmer(ct, not args.variable_coverage, args.cutoff,
+                      args.trim_at_coverage)
+    if args.diginorm:
+        trimmer.do_normalize = True
+        trimmer.diginorm_coverage = int(args.diginorm_coverage)
```
luizirber: These two can be initialized in Trimmer.__init__ (defaulting to False and 0?)

scripts/trim-low-abund.py
```diff
@@ -175,122 +364,85 @@ def main():
...
-                                     args.gzip, args.bzip)
-        else:
-            trimfp = get_file_writer(args.output, args.gzip, args.bzip)
+        pass2fp = open(pass2filename, 'w')
 
-        pass2list.append((filename, pass2filename, trimfp))
+        # construct output filenames
+        outfp = args.output
+        if outfp is None:
+            outfp = open(os.path.basename(filename) + '.abundtrim', 'wb')
```
luizirber: `outfp` and `trimfp` are never closed. What do you think about using context managers to control this?

scripts/trim-low-abund.py
```diff
@@ -110,9 +101,9 @@ def get_parser():
                         help='remove k-mers below this abundance',
                         default=DEFAULT_CUTOFF)
 
-    parser.add_argument('--normalize-to', '-Z', type=int,
-                        help='base cutoff on this median k-mer abundance',
-                        default=DEFAULT_NORMALIZE_LIMIT)
+    parser.add_argument('--trim-at-coverage', '-Z', type=int,
```
luizirber: I didn't know this worked, but
```python
parser.add_argument('--trim-at-coverage', '--normalize-to', '-Z', type=int,
```
also works (if there is anyone using the `--normalize-to` flag, it won't break for them)

tests/test_script_output.py
```diff
@@ -0,0 +1,192 @@
...
+
+def test_trim_low_abund_k21_M1e7_C4_variable_Z15():
+    infile = utils.get_test_data('simple-genome-reads.fa')
+    outfile = utils.get_temp_filename('out')
+    utils.runscript('trim-low-abund.py', ['-C', '4', '-k', '21',
+                                          '-V', '-Z', '15',
+                                          '-M', '1e7', '-o', outfile,
+                                          infile])
+
+    hash = _calc_md5(open(outfile, 'rb'))
```
luizirber: This is nitpicking, but let's close this file after using it:

```python
with open(outfile, 'rb') as output:
    hash = _calc_md5(output)
```

(Bonus points if you avoid using `hash` as a variable name, since it is a built-in function)

scripts/trim-low-abund.py
ctb: +1

scripts/trim-low-abund.py
ctb: I don't know how to use context managers here if 'outfp' is only sometimes created in the loop?

ctb: In any case, fixed in cc3b3e93ab

ctb:
Ready for review & merge @luizirber

ctb:
Ready for review & merge.

khmer/kfile.py
```diff
@@ -232,6 +232,11 @@ def get_file_writer(file_handle, do_gzip, do_bzip):
     if do_gzip and do_bzip:
         raise ValueError("Cannot specify both bzip and gzip compression!")
 
+    if file_handle is sys.stdout:         # needed for Python 3/gzip.
```
luizirber: Should this be removed, since it is being worked on #1367 ?

ctb:
Sorry, thought I'd committed that change already - now done. Ready once more!

PR #1159, opened 2015-07-08
Escape incorrect signature as hex values

luizirber:
`test_labelhash.py:test_load_wrong_filetype` is failing sometimes. It is somewhat hard to reproduce, but the issue is that the file signature can contain unprintable characters and the Python exception (`OSError`) is not being built properly.

This PR just print the char[4] string in hexadecimal. It is a bit weird, because the error message says it should be 'OXLI', which is not hexadecimal...

Previous (and this is non-deterministic, file signature can be different values):
`Incorrect file signature >a (m while reading labels/tags from tests/test-data/all-A.fa Should be: OXLI`
Now:
`Incorrect file signature 0x3e612028 while reading labels/tags from tests/test-data/all-A.fa Should be: OXLI`

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-07-08  Luiz Irber  <khmer@luizirber.org>
+
+   * lib/{counting,hashbits,hashtable,labelhash,subset}.cc: print hexadecimal
+   representation of the signature read from the file.
+
 2015-07-06  Luiz Irber  <khmer@luizirber.org>
 
    * sandbox/collect-reads.py: Set a default value for coverage based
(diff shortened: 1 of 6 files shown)


luizirber:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

luizirber:
Ready for review and merge @mr-c

mr-c:
Thanks! Ugh, I hate the underlying error though.

luizirber:
There is a lot of copy-and-paste on the save/load code over
{labelhash,hashbits,counting,subset,hashtable}.cc, maybe move most of it to
hashtable.cc?

ctb:
On Wed, Jul 08, 2015 at 12:56:09PM -0700, Luiz Irber wrote:
> There is a lot of copy-and-paste on the save/load code over
> {labelhash,hashbits,counting,subset,hashtable}.cc, maybe move most of it to
> hashtable.cc?

Please file an issue!

PR #1163, opened 2015-07-09
No more complaining about FIFO's being empty (addresses #1147)

### LINKED ISSUE #1147: kfile.check_valid_file_exists() should return silently if it's a FIFO
see logic in kfile.check_input_files() for example.

As it is, running `normalize-by-median` on a FIFO warns that the FIFO is empty.

bocajnotnef:
fixes #1147 

- [x] Is it mergeable?
- [x] Did it pass the tests?
  fixing as needed.
  changes were made?

@ctb @luizirber CR/merge please.

### DIFF
--- a/khmer/kfile.py
+++ b/khmer/kfile.py
@@ -149,7 +149,7 @@ def check_valid_file_exists(in_files):
     """
     for in_file in in_files:
         if os.path.exists(in_file):
-            if os.stat(in_file).st_size > 0:
+            if os.stat(in_file).st_size >= 0:
                 return
             else:
                 print('WARNING: Input file %s is empty' %
--- a/khmer/kfile.py
+++ b/khmer/kfile.py
@@ -150,8 +150,7 @@ def check_valid_file_exists(in_files):
     for in_file in in_files:
         if os.path.exists(in_file):
             mode = os.stat(in_file).st_mode
-            if os.stat(in_file).st_size > 0 and not S_ISBLK(mode) or\
-               S_ISFIFO(mode):
+            if os.stat(in_file).st_size > 0 or S_ISBLK(mode) or S_ISFIFO(mode):
                 return
             else:
                 print('WARNING: Input file %s is empty' %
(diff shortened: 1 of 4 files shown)


mr-c:
Needs a test

bocajnotnef:
@mr-c Updated

khmer/kfile.py
```diff
@@ -149,7 +149,7 @@ def check_valid_file_exists(in_files):
     """
     for in_file in in_files:
         if os.path.exists(in_file):
-            if os.stat(in_file).st_size > 0:
+            if os.stat(in_file).st_size >= 0:
```
ctb: am I missing something? if the file size is 0 this will not print a warning, right? don't we want the warning in that case?

ctb:
I don't understand this pull request at all.

- the fix doesn't address FIFOs, and seems to apply to all empty files;
- there's no test that breaks when the "fix" to kfile.py is applied;
- the new test doesn't break when I revert the change to kfile;

So,

- write a test that breaks with your current changes to kfile.py line 152.
- revert that change, which is a bad one.
- please address FIFO size specifically, as in #1147.

Or, alternatively, tell me that I'm wrong and convince me :)

bocajnotnef:
Misunderstood how stat works in this case. I'll sort it.

bocajnotnef:
@ctb updated. pep8 and tests passing locally, test breaks when change is reverted and kfile still complains about empty files. 


Jenkins linux build appears to be hanging/breaking for other reasons.

ctb:
OK - could you please 

* write a test that breaks with your previous changes to kfile.py line 152

as well?

ctb:
(the current code looks good, although if you wanted to use parantheses to put an explicit grouping around the 'and not' clause in the if statement, it would make it easier to read.)

bocajnotnef:
As in, the bad changes I made? (Allowing all empty files through)

ctb:
yes.

bocajnotnef:
tests/pep8 passing locally--jenkins is borked for other reasons

@ctb Updated; Merge, please?

mr-c:
@bocajnotnef is this urgent? I'd rather fix Jenkins right now.

bocajnotnef:
@mr-c Nope, is not urgent.

Issue #847, opened 2015-02-25
split-paired-reads.py has nonstandard behavior with orphaned reads


camillescott:
By default, split-paired-reads.py mixes orphaned reads into their respective .1 / .2 files, which is a nonstandard behavior -- orphaned reads should be put into their own files. Tools like bowtie and Trinity fail when they encounter orphaned reads mixed into split paired files.

ctb:
We talked about this in person, and the conclusion was that we cannot alter this until khmer 2.0; but the -p was added in #818 to force the expected behavior.

Two thoughts -

1. shall we update the documentation to make this clear?
2. @mr-c, what's the right way to punt these kinds of issues to khmer 2.0 release?

ctb:
For 2.0,

- [ ] make -p default;
- [ ] upgrade extract-paired-reads to properly handle streaming input and specification of output files;
- [ ] in the error message that results from -p, mention that extract-paired-reads can be used to fix;

@camillescott @mr-c the alternative here is to add an option to split-paired-reads to sideline or trash orphans. I think this makes the script too complicated so am -0 on it but would appreciate your thoughts.

ctb:
Closed by #1164.

PR #747, opened 2015-01-30
gzip/bzip2 output options

b-wyss:
Feature creation in response to #505

PR #1164, opened 2015-07-10
Add --output-orphaned option to split-paired-reads.py

camillescott:
Currently, so far as I can tell, there's no way to both split paired reads into .1 and .2 files while also outputting orphans to their own file without doing messy pipe stuff. This adds that feature to split-paired-reads.py with an extra command line option.


ctb:
This should fix #847, right?

ctb:
@camillescott may I snarf this?

ctb:
ping @camillescott

ctb:
Taking it and making it my own.

mr-c:
LGTM

ctb:
- [ ] streaming_io tests
- [ ] maybe add in gzip stuff from #747

ctb:
- [ ] test -0 without filename
- [ ] test -0 without filename and with output-directory
- [ ] test -0 with filename
- [ ] test -p and evaluat eerror message
- [ ] compression stuff (bzip2/gzip)

ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

Ready for review @mr-c @bocajnotnef @luizirber

ctb:
Fixes #847.

ChangeLog
```diff
@@ -1,3 +1,14 @@
+2015-08-04  Titus Brown  <titus@idyll.org> and
+Camille Scott  <camille.scott.w@gmail.com>
```
mr-c: Can you indent this to match your name?

scripts/split-paired-reads.py
```diff
@@ -37,16 +38,16 @@ def get_parser():
...
 
     Alternatively, you can specify the filenames directly with
     :option:`-1`/:option:`--output-first` and
     :option:`-2`/:option:`--output-second`, which will override the
-    :option:`-o`/:option:`--output-dir` setting on a file-specific basis.
+    :option:`-d`/:option:`--output-dir` setting on a file-specific basis.
 
-    :option:`-p`/:option:`--force-paired` will require the input file to
-    be properly interleaved; by default, this is not required.
+    :option:`-0`/:option:'`--output-orphans` will allow broken-paired format,
```
mr-c: extra extraneous single quote after the 2nd `:option:`

scripts/split-paired-reads.py
```diff
@@ -128,47 +131,48 @@ def main():
         # Use default filename created above
         fp_out2 = get_file_writer(open(out2, 'wb'), args.gzip, args.bzip)
 
+    # put orphaned reads here, if -0!
+    allow_orphans = False
```
mr-c: Why not use `args.output_orphaned` everywhere you check `allow_orphans`?

mr-c:
Otherwise, this LGTM

scripts/split-paired-reads.py
ctb: On Tue, Aug 04, 2015 at 09:52:19AM -0700, Michael R. Crusoe wrote:
> > @@ -128,47 +131,48 @@ def main():
> >          # Use default filename created above
> >          fp_out2 = get_file_writer(open(out2, 'wb'), args.gzip, args.bzip)
> >  
> > +    # put orphaned reads here, if -0!
> > +    allow_orphans = False
> 
> Why not use `args.output_orphaned` everywhere you check `allow_orphans`?

It's partly a leftover of some now-removed logic that permitted -0 without
a filename, but what made me keep it in was this code:

`require_paired=not allow_orphans`

which I think would be uglier and less clear if it was
`not args.output_orphaned`.  What do you think?

scripts/split-paired-reads.py
ctb: On Tue, Aug 04, 2015 at 09:50:34AM -0700, Michael R. Crusoe wrote:
> >  
> > -    :option:`-p`/:option:`--force-paired` will require the input file to
> > -    be properly interleaved; by default, this is not required.
> > +    :option:`-0`/:option:'`--output-orphans` will allow broken-paired format,
> 
> extra extraneous single quote after the 2nd `:option:`

how do you find this stuff!? fixed :)

ctb:
- [ ] standardize on --output instead of --out (see #1188)

scripts/split-paired-reads.py
mr-c: I'm okay with `not args.output_orphaned`

ctb: On Tue, Aug 04, 2015 at 03:54:24PM -0700, Michael R. Crusoe wrote:
> > @@ -128,47 +131,48 @@ def main():
> >          # Use default filename created above
> >          fp_out2 = get_file_writer(open(out2, 'wb'), args.gzip, args.bzip)
> >  
> > +    # put orphaned reads here, if -0!
> > +    allow_orphans = False
> 
> I'm okay with `not args.output_orphaned` 

That doesn't address "here's my reasoning, what do you think" but I'll give
it a whirl ;)

ctb:
Fixes #1188.

Ready for review @mr-c.

mr-c:
Great, thanks!

Issue #1169, opened 2015-07-10
Argparse arg naming inconsistencies throughout codebase


bocajnotnef:
https://github.com/dib-lab/khmer/blob/master/scripts/trim-low-abund.py#L86 
vs
https://github.com/dib-lab/khmer/blob/master/oxli/build_graph.py#L34

We should settle on a single thing; I nominate `output_filename`'

@mr-c @luizirber @mr-c @camillescott thoughts?

ctb:
On Fri, Jul 10, 2015 at 01:05:36PM -0700, Jake Fenton wrote:
> https://github.com/dib-lab/khmer/blob/master/scripts/trim-low-abund.py#L86 
> vs
> https://github.com/dib-lab/khmer/blob/master/oxli/build_graph.py#L34
> 
> We should settle on a single thing; I nominate `output_filename`'

Why not have the controlling script specify the directory/location to
check in this case?

bocajnotnef:
we'd have to add an arg to the function which just seems excessive when it already takes in the `args` object from the controlling script.

ctb:
On Fri, Jul 10, 2015 at 01:13:02PM -0700, Jake Fenton wrote:
> we'd have to add an arg to the function which just seems excessive when it already takes in the `args` object from the controlling script.

presumably this happens elsewhere in the codebase - maybe look through and see
whether it looks like the proposed global standardization causes problems
anywhere.

or, replace 'args' with a list of the directories being considered.

bocajnotnef:
we rely on `_calculate_tablesize` which requires the args object (to determine if args.max_memory_usage is set, which we could work around I suppose); otherwise +1 to the list.

ctb:
did you look through the codebase to see whether standardizing on names in `args` makes sense?

bocajnotnef:
I believe it does--especially if we're getting in the habit of tossing `args` objects about to do information gathering/checking on them.

increases modularity of the scripts and would also probably make it easier to migrate to an oxli-like API.

luizirber:
Explicit is better than implicit, we are using only three arguments from the `args`. Why not refactor the function to take these three arguments instead?

mr-c:
+1 for never passing the args object around

On Thu, Jul 16, 2015 at 11:48 AM Luiz Irber <notifications@github.com>
wrote:

> Explicit is better than implicit, we are using only three arguments from
> the args. Why not refactor the function to take these three arguments
> instead?
>
> —
[...]

Issue #1167, opened 2015-07-10
Get rid of kfile.check_space call in load-into-counting


ctb:
load-into-counting doesn't output anything other than a counting graph, so this check is unnecessary.  Also see #1166.

bocajnotnef:
Taking this (and #1166)

ctb:
`git grep check_space\( scripts | cut -d: -f 1` yields the following files which this check should not be in:

```
scripts/abundance-dist-single.py
scripts/count-overlap.py
scripts/load-into-counting.py
scripts/make-initial-stoptags.py
scripts/partition-graph.py
scripts/sample-reads-randomly.py
```

so it should be taken out of all of them, @bocajnotnef  - thx!

bocajnotnef:
Ah, didn't realize that extended beyond load. (as in, that it was equally
unnecessary in those files.)

Will remove.

bocajnotnef:
`load-into-counting` has a call to `check_space_for_hashtable` but not `check_space`--I left the hashtable call in.

ctb:
Fixed in #1170.

Issue #1166, opened 2015-07-10
kfile.check_space_for_hashtable only examines the current directory...


ctb:
...not the location where we actually want to save the hashtable. This is broken behavior.

bocajnotnef:
Taking this

ctb:
Fixed in #1170.

Issue #668, opened 2014-11-20
free space check incorrectly fails on HPCC


mr-c:
```
|| This is the script 'load-into-counting.py' in khmer.
|| You are running khmer version 1.1
[...]
Estimated memory usage is 2.2e+11 bytes (n_tables x min_tablesize)
--------
ERROR: Not enough free space on disk for saved table files;       Need at least 215.040499418 GB more.
       Table size: 215.0 GB
       Free space: 0.0 GB
```

It turns out that I didn't have write permissions to the directory I wanted to output to so perhaps that was the culprit.

bocajnotnef:
testing; case: can't write to `/home/root` (surprise)

```
(env)bocajnotnef@GCU-Caconym:~/git/khmer$ load-into-counting.py /home/root/table tests/test-data/test-large.fa 

|| This is the script 'load-into-counting.py' in khmer.
|| You are running khmer version 2.0-rc1+116.g2dacd90
|| You are also using screed version 0.9
||
|| If you use this script in a publication, please cite EACH of the following:
||
||   * MR Crusoe et al., 2014. http://dx.doi.org/10.6084/m9.figshare.979190
||   * Q Zhang et al., http://dx.doi.org/10.1371/journal.pone.0101271
||   * A. Döring et al. http://dx.doi.org:80/10.1186/1471-2105-9-11
||
|| Please see http://khmer.readthedocs.org/en/latest/citations.html for details.


PARAMETERS:
 - kmer size =    32 		(-k)
 - n tables =     4 		(-N)
 - max tablesize = 1e+06 	(-x)

Estimated memory usage is 4e+06 bytes (n_tables x max_tablesize)
--------

** WARNING: tablesize is default!
** You probably want to increase this with -M/--max-memory-usage!
** Please read the docs!

ERROR: File /home/root/table does not have write permission; exiting
```

bocajnotnef:
scenario: mounted SD card to `/home/root` as root--still don't have write permissions:

```
(env)bocajnotnef@GCU-Caconym:~/git/khmer$ sudo umount /media/bocajnotnef/Mercury 
(env)bocajnotnef@GCU-Caconym:~/git/khmer$ sudo mount /dev/mmcblk0p1 /home/root/
(env)bocajnotnef@GCU-Caconym:~/git/khmer$ sudo ls /home/root/
System Volume Information  testdir
[...]

PR #1170, opened 2015-07-10
Fix weird things with space checks in some files (see #1167 #1166)

SensibleSalmon:
See #1166, #1167 

Blocked by: #1169

Lightweight argparse cleanup, making space checks use the right directory instead of cwd.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2015-07-16  Jacob Fenton  <bocajnotnef@gmail.com>
+
+   * khmer/{kfile,khmer_args}.py: refactored information passing, made it so
+   space checks happen in the right directory
+   * oxli/build_graph.py,sandbox/collect-reads.py,scripts/{count-overlap,
+   abundance-dist-single,filter-abund-single,load-into-counting,
+   make-initial-stoptags,normalize-by-median,partition-graph,
+   sample-reads-randomly,trim-low-abund}.py,tests/test_script_arguments.py:
+   changed to use new arg structure
+
 2015-07-10  Jacob Fenton  <bocajnotnef@gmail.com>
 
    * oxli/functions.py: changed estimate functions to use correct letter
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,12 +1,16 @@
-2015-07-20  Jacob Fenton  <bocajnotnef@gmail.com>
+2015-07-21  Jacob Fenton  <bocajnotnef@gmail.com>
 
    * khmer/{kfile,khmer_args}.py: refactored information passing, made it so
-   space checks happen in the right directory
-   * oxli/build_graph.py,sandbox/collect-reads.py,scripts/{count-overlap,
+   space checks happen in the right directory.
+   * oxli/build_graph.py,sandbox/collect-reads.py,scripts/{
    abundance-dist-single,filter-abund-single,load-into-counting,
-   make-initial-stoptags,normalize-by-median,partition-graph,
-   sample-reads-randomly,trim-low-abund}.py,tests/test_script_arguments.py:
-   changed to use new arg structure
+   normalize-by-median,trim-low-abund}.py,tests/test_script_arguments.py:
+   changed to use new arg structure for checking hashtable save space.
+   * oxli/functions.py,scripts/saturate-by-median.py: updated error message
+   to mention --force option.
+   * scripts/{count-overlap,load-into-counting,make-initial-stoptags,
+   partition-graph,sample-reads-randomly}.py: removed unnecessary call to
+   check_space.
 
 2015-07-20  Titus Brown  <titus@idyll.org>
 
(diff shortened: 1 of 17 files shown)


ctb:
@bocajnotnef do you see anything in the code that would lead to this problem?

http://lists.idyll.org/pipermail/khmer/2015-July/000736.html

I'm stumped...

bocajnotnef:
I recall one of the checks was only checking the directory the script was
called from, not the directory of the given output filename. I forget which
issue specifically but it's one of the ones involved in this pull.

bocajnotnef:
I'll dig up which issue in a minute.

ctb:
OK, could you track it down, please? thanks!

bocajnotnef:
Issue #1166, line 120ish in kfile.py (https://github.com/dib-lab/khmer/pull/1170/files#diff-46d242f8c7f4ca46fd5c784bd24184c7R120)

```
 """Check we have enough size to write a hash table."""                      
    hash_size = khmer_args._calculate_tablesize(args, hashtype)                 
                                                                                
    cwd = os.getcwd()                                                           
    dir_path = os.path.dirname(os.path.realpath(cwd))                           
    target = os.statvfs(dir_path)                             
```

We grab `cwd` and not `os.path.dirname(args.out)` or summuch.

bocajnotnef:
in this case a `--force` should solve the issue until a fix gets implemented.

ctb:
By my reading, that's not the problem in the email, tho?

--

bocajnotnef:
Looks to be; error is: `ERROR: Not enough free space on disk for saved table files;  ` (from the email).

Output from kfile on line ~130 (https://github.com/dib-lab/khmer/blob/master/khmer/kfile.py#L130): 

```
    if size_diff > 0:
        print("ERROR: Not enough free space on disk "
              "for saved table files;"
              "       Need at least %s GB more."
              % (float(size_diff) / 1e9,), file=sys.stderr)
```

bocajnotnef:
Ahh, I see--He is calling norm-by-med from his `mnt` directory. I'd missed that.

ctb:
Yep. I wonder if os.stat on a mount point returns parent fs info?

--

bocajnotnef:
It's behaving as if what I said is the problem but since he's calling norm-by-med on the mounted partition this shouldn't be happening. The only explanation I can think of is if he installed khmer to the system (or a virtualenv) and `os.getcwd()` is somehow getting the directory of where khmer is installed and not where it's being called from--which seems extremely weird.

I'm ruling out that it's something to do with `os.statvfs` since it's correctly reporting the amount of free space on the system partition.

Could it be that `os.path.dirname(os.path.realpath(cwd))` is getting root as the parent FS and is checking there...?

bocajnotnef:
Some testing in ipython indicates it's at least a similar issue; `Mercury` is an SD card I use for spare storage on my laptop. Attempting to stat as kfile does yields:

```
In [12]: dir = os.path.dirname(os.path.realpath('/media/bocajnotnef/Mercury'))

In [13]: res = os.statvfs(os.path.realpath(dir))

In [14]: dir
Out[14]: '/media/bocajnotnef'
```

Which, then if you stat `/media/bocajnotnef` you'd get the size of my root filesystem and not the SD card.

bocajnotnef:
Did some quick testing in iPython--This doesn't appear to be it.

Ran the kfile space check on a file in my homedir, got:

```
In [5]: cwd = os.getcwd()

In [6]: cwd
Out[6]: '/home/bocajnotnef'

In [8]: res = os.statvfs(os.path.realpath(cwd))

In [9]: fs = res.f_frsize * res.f_bavail

In [10]: fs
Out[10]: 75171151872
```

Running the same on a file on an SD card:

```
In [18]: dir = os.path.dirname(os.path.realpath('/media/bocajnotnef/Mercury/test.file'))

In [19]: dir
Out[19]: '/media/bocajnotnef/Mercury'

In [20]: res = os.statvfs(dir)

In [21]: fs = res.f_frsize * res.f_bavail

In [22]: fs
Out[22]: 67091300352
```

I have ~70gigs free on my system partition and the SD card is 64 gigs, so those numbers look about right.

ctb:
On Tue, Jul 14, 2015 at 08:00:44AM -0700, Jake Fenton wrote:
> Did some quick testing in iPython--This doesn't appear to be it.

I concur.

Weird, well, I will let them know.

--titus

bocajnotnef:
@ctb Updated; CR/merge?

ctb:
Ref https://github.com/dib-lab/khmer/issues/1169#issuecomment-122082228, what do you think about breaking it out like this?

```
tablesize = khmer_args.calculate_tablesize(args)
check_space_for_hashtable(tablesize)
```

in the various scripts?

I definitely foresee a future where tablesize is determined from more args, but the check space function doesn't need to know that - it just needs to know the estimated size.

bocajnotnef:
+1 for that solution.

bocajnotnef:
@mr-c @ctb CR, please?

I feel I should add a test to avoid the previous behaviour (Testing the wrong directory's space) but I'm unsure of how to come up with a test scenario for this--we can't assume that whatever system we're testing on has a convienently mounted drive for us to use.

ctb:
Ref #618.

ctb:
Also, can you take a look at #668?

bocajnotnef:
I'd need access to the HPCC (or a similar setup) to verify the behaviour
still exists (or have @mr-c confirm)

ctb:
On Sat, Jul 18, 2015 at 01:07:06PM -0700, Jake Fenton wrote:
> I'd need access to the HPCC (or a similar setup) to verify the behaviour
> still exists (or have @mr-c confirm)

You might first check to see what the behavior is for disks to which
you have read-only access.

bocajnotnef:
Seems to be what we'd expect--that is, it yells at you that you don't have permissions. (comments in #668)

ctb:
Yes, but it doesn't cause the same error that @mr-c saw.  So it looks like there's some oddity on HPC systems (?).

ctb:
I think 'calculate_tablesize' should just take 'args', so as to accomodate things like `-U/--unique-kmers`. Otherwise LGTM on a first pass!

Note that you could fix #1179 rather casually at this point, too.

bocajnotnef:
What's strange is khmer thinks there was no free space at all--that doesn't
look like it's reading the wrong directory.

I'm wondering if not having read permissions would show up as having no
free space in a stat.

Though, there isn't anything about not having read perms in the issue.

ctb:
On Sun, Jul 19, 2015 at 08:15:28AM -0700, Jake Fenton wrote:
> What's strange is khmer thinks there was no free space at all--that doesn't
> look like it's reading the wrong directory.
> 
> I'm wondering if not having read permissions would show up as having no
> free space in a stat.

Maybe? 'df' shows file size for things you don't have read perms to,
I think. But it's a good point and easy enough to test.

> Though, there isn't anything about not having read perms in the issue.

True.

bocajnotnef:
I'll do some testing in a bit, as well as integrating -U

bocajnotnef:
Well, that wasn't it.

```
ERROR: File /tmp/testingdir/table does not have write permission; exiting
(env)1 bocajnotnef@GCU-Caconym:~/git/khmer⟫ ls /tmp/testingdir/
ls: cannot open directory /tmp/testingdir/: Permission denied
(env)2 bocajnotnef@GCU-Caconym:~/git/khmer⟫ 
```

it is also possible to stat the directory:

```
In [9]: dir_path = os.path.dirname(os.path.realpath('/tmp/testingdir/table'))

In [10]: target = os.statvfs(dir_path)

In [11]: free_space = target.f_frsize * target.f_bavail

In [12]: free_space
Out[12]: 75911532544
```

ctb:
hey @bocajnotnef, can you also fix the `%s` to be a `%.1f` in `check_space_for_hashtable`, around line 135 of khmer/kfile.py?

```
Need at least %s GB more.
```

and then check this box:

- [ ] fixed GB output to truncate float

bocajnotnef:
figured out weirdness with HPCC stuff--this should fix it.

@mr-c @ctb @camillescott CR/Merge, please?

ChangeLog
```diff
@@ -7,7 +17,6 @@
    * tests/test_normalize_by_median.py: updated/added tests for reporting.
 
 2015-07-17  Jacob Fenton  <bocajnotnef@gmail.com>
-
```
ctb: ?

bocajnotnef:
@ctb Updated

ctb:
Cleaned up minor slop in 4ecd60b; merging.

ctb:
BTW, @bocajnotnef, don't put the issue numbers in the pull request subject; you can't click on 'em. Better to just put them in the description somewhere, so as to link issues & PR.

bocajnotnef:
Ahh, alright. Wilco.

PR #1176, opened 2015-07-17
Promote unique-kmers.py to scripts

luizirber:
There weren't any tests for the `unique-kmers.py` script. Since I was adding tests, why not promote it from sandbox?

I think this can be merged before khmer 2.0, but it needs some work. The output is a bit wonky, sometimes there is output being sent to stdout that would be better in stderr.

(There were two bugs: a missing import for `write_record`, and using `to_print` before it was initialized.

PR #1177, opened 2015-07-17
Streaming unique-kmers.py using consume_fasta

luizirber:
Revert `unique-kmers.py` changes and use consume_fasta again. Needed to write a `write_record` equivalent in C++ for this, currently living in `lib/HLLCounter.cc`, but should probably go in `lib/read_parsers.cc`?

(don't merge this before #1176 )


luizirber:
Reopened as #1204

Issue #393, opened 2014-04-19
Investigate/implement streaming approaches, more generally


ctb:
(Updated in #1206)

In partial response to ivory.idyll.org/blog/2014-pycon.html, can we put our protocols on a streaming basis by systematically introducing streaming functionality into khmer?

There are ways to do this even for e.g. Trimmomatic using clever Unix socket tricks.

Also see #149.

ctb:
This should be relevant: https://github.com/ctb/2015-experimental-graphalign/blob/master/khmer_api.py#L136

What fun!

ctb:
Also: http://ivory.idyll.org/blog/2015-a-generator-api-for-khmer.html

ctb:
Streaming fixes & tests in #1186; this will finish off most of the straightforward practical issues.

PR #1186, opened 2015-07-21
Update scripts to handle stdin.

### LINKED ISSUE #700: state of streaming I/O in khmer, mark 2
Now that SeqAn has landed we support reading from a streamed FASTQ or FASTA using the ReadParser interface: both uncompressed and gziped. Streaming of a bzip2 compressed file using ReadParser does not work natively though if it is a single file it can be piped through a decompressor and into our scripts. The bzip2 issue is fixed in the currently under-development version of SeqAn (2.0), see https://github.com/seqan/seqan/issues/707#issuecomment-64371262 for a discussion.

Screed supports streaming from uncompressed FASTQ and FASTA files in https://github.com/ged-lab/screed/pull/11 which is waiting for review and merging. Bzip2 compressed files are also supported in streaming mode but due to deficiencies in Python 2.x gzip files are not natively streamable. We could backport code from Python 3.x to work around that. http://bugs.python.org/file15619/gzip_7471_py27.diff

How to stream: specify `/dev/stdin` as the input filename and/or `/dev/stdout` as the output filename.

Scripts that use ReadParser can abbreviate `/dev/stdin` as `-`. This could be ported to screed as well as supported as the output filename.

[edited to remove zip archive support as that has been dropped]

### LINKED ISSUE #1055: All scripts/ possible should support '-' input
This is for better, more productive streaming.

Should '-' be default? Probably not.

Also, always allow '-' in output.

ctb:
Fixes problems with scripts that don't support stdin, which should be supported as input for streaming awesomeness.

Fixes #1055. Will close #700, too. And after this commit we should update #393.


ctb:
@mr-c curious what you think of https://github.com/dib-lab/khmer/blob/update/read_handling/tests/test_streaming_io.py - I judged it too cumbersome to make use of our existing test_scripts infrastructure, but it's important to have these... thoughts?

ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

ctb:
Missing coverage:
```
khmer/kfile.py (66.7%): Missing lines 161
scripts/extract-long-sequences.py (100%)
scripts/filter-abund.py (71.4%): Missing lines 87,89
scripts/interleave-reads.py (100%)
scripts/normalize-by-median.py (33.3%): Missing lines 349,351
scripts/sample-reads-randomly.py (33.3%): Missing lines 112,114
scripts/split-paired-reads.py (66.7%): Missing lines 103
scripts/trim-low-abund.py (60.0%): Missing lines 134,136
```

ctb:
Will close #700.

ctb:
Ready for review @bocajnotnef @mr-c.

Two notes:
* there's one line that's putatively uncovered (line 161 in kfile.py) but it is definitely executed - I tried puttin in an assert 0 there to check.  So this is a bug in coverage analysis.
* Right now the new test_streaming_io.py file won't run properly after install. I think this is OK; thoughts welcome.

luizirber:
Please hold the merge, this is broken in Python 3.

luizirber:
See #1191 for py3 fixes.

ctb:
It all runs on py3 now. ping @mr-c @bocajnotnef

luizirber:
I really like `run_shell_cmd`, can we start using it instead of `runscriptredirect`?

ctb:
On Wed, Jul 22, 2015 at 08:39:23AM -0700, Luiz Irber wrote:
> I really like `run_shell_cmd`, can we start using it instead of `runscriptredirect`?

It seems to do the same thing but be less complicated, so why not? ;)
Still, let's not be in a hurry to change things that are currently working...

bocajnotnef:
I'm seeing a few pep8 violations.

Do we want to migrate all existing streaming tests (there's a few in test_normalize_by_median) into test_streaming_io?

khmer/kfile.py
```diff
@@ -25,7 +25,7 @@ def check_input_files(file_path, force):
     """
     mode = None
 
-    if file_path == '-':
+    if file_path == '-' or file_path == '/dev/stdin':
```
mr-c: adding `or S_ISCHR(mode)` to line 45 (plus import on line 15) and not hardcoding the path here will be more effective and flexible.

khmer/kfile.py
```diff
@@ -157,7 +157,9 @@ def check_valid_file_exists(in_files):
     or non-existent.
     """
     for in_file in in_files:
-        if os.path.exists(in_file):
+        if in_file == '-' or in_file == '/dev/stdin':
```
mr-c: '/dev/stdin' passes `os.path.exists()` unless it truly doesn't exist in which case it would be a real error (and a screwed up system).

mr-c: Also, I'm a bit uncomfortable with the structure here. Are we allowing a mixture of STDIN & named input files? Seems messy.

scripts/abundance-dist.py
```diff
@@ -87,7 +87,9 @@ def main():
     print('HT sizes:', hashsizes, file=sys.stderr)
     print('outputting to', args.output_histogram_filename, file=sys.stderr)
 
-    if os.path.exists(args.output_histogram_filename):
+    if args.output_histogram_filename in ('-', '/dev/stdin'):
+        pass
+    elif os.path.exists(args.output_histogram_filename):
```
mr-c: Same comment as before with regards to `/dev/stdin` existing; not needed

scripts/count-median.py
```diff
@@ -91,7 +91,10 @@ def main():
     ksize = htable.ksize()
 
     print('writing to', output_filename, file=sys.stderr)
-    output = open(output_filename, 'w')
+    if output_filename in ('-', '/dev/stdin'):
+       output = sys.stdout
+    else:
+       output = open(output_filename, 'w')
```
mr-c: I disagree. It is a nice default but one should be allowed to stream in and write to a file. Especially bad as we say to the user in line 93 that we are writing to the file name they specified but we immediately go and do something different.

mr-c: perhaps you meant `if output_filename in ('-', '/dev/stdout'):` in which case +1 :-)

tests/test_streaming_io.py
```diff
@@ -0,0 +1,448 @@
...
+from . import khmer_tst_utils as utils
+from .test_scripts import _make_counting
+import subprocess
+import os.path
+import difflib
+
+
+def scriptpath():
+    path = os.path.join(os.path.dirname(__file__), "../scripts")
+    return path
```
mr-c: Did you test post-install?

mr-c:
Seems so easy! Does it work across all our OS / Python matrix and for both
devs and users?

On Tue, Jul 21, 2015 at 4:47 PM C. Titus Brown <notifications@github.com>
wrote:

> @mr-c <https://github.com/mr-c> curious what you think of
> https://github.com/dib-lab/khmer/blob/update/read_handling/tests/test_streaming_io.py
> - I judged it too cumbersome to make use of our existing test_scripts
> infrastructure, but it's important to have these... thoughts?
>
> —
> <https://github.com/dib-lab/khmer/pull/1186#issuecomment-123510713>.
>
-- 

mr-c:
"Right now the new test_streaming_io.py file won't run properly after
install. I think this is OK; "
Ah, not Okay :-)


On Wed, Jul 22, 2015 at 5:59 AM C. Titus Brown <notifications@github.com>
wrote:

> Ready for review @bocajnotnef <https://github.com/bocajnotnef> @mr-c
> <https://github.com/mr-c>.
>
> Two notes:
>
>    - there's one line that's putatively uncovered (line 161 in kfile.py)
>    but it is definitely executed - I tried puttin in an assert 0 there to
>    check. So this is a bug in coverage analysis.
>    - Right now the new test_streaming_io.py file won't run properly after
>    install. I think this is OK; thoughts welcome.
>
> —
> <https://github.com/dib-lab/khmer/pull/1186#issuecomment-123711557>.
>
-- 

khmer/kfile.py
ctb: Nothing we didn't do before any change I made! Punt to an issue if you want to discuss.

ctb:
There's nothing OS-specific about it.  I'll see what I can do about the
post-install stuff.

ctb:
On Wed, Jul 22, 2015 at 10:08:05AM -0700, Jake Fenton wrote:
> I'm seeing a few pep8 violations.

I'm running 'make pep8' *shrug*.

> Do we want to migrate all existing streaming tests (there's a few in test_normalize_by_median) into test_streaming_io?

No reason to do it that I can see; file an issue if you want to talk about it :)

scripts/abundance-dist.py
ctb: Ah, but here the logic is going to exit *because* /dev/stdin already exists. So I'll leave this one in.

scripts/count-median.py
ctb: Indeed :)

ctb:
OK, fixed the issues pointed out by @mr-c, other than the question of test_streaming_io.py. I'll work on that on my next flight (leaving ~now :).

ctb:
Victory! @bocajnotnef @mr-c @luizirber review (and maybe merge), please!

mr-c:
I'm reviewing this

mr-c:
The issue with the new test runner is that `coverage` can't measure code coverage inside the subshell.

```
make coverage report

Name                             Stmts   Miss Branch BrMiss  Cover
------------------------------------------------------------------
khmer/__init__                     149      2     32      4    97%
khmer/kfile                         82      6     34      5    91%
khmer/khmer_args                   132      7     49     13    89%
khmer/thread_utils                 137      8     40      1    95%
khmer/utils                         81      0     44      0   100%
oxli/__init__                       18      1      2      1    90%
oxli/build_graph                    51      1     10      1    97%
oxli/functions                      82      0     30      2    98%
scripts/abundance-dist              67     67     22     22     0%
scripts/abundance-dist-single      101    101     35     35     0%
scripts/annotate-partitions         40     40      6      6     0%
scripts/count-median                53     53     16     16     0%
scripts/count-overlap               43     43     10     10     0%
scripts/do-partition               124    124     32     30     1%
scripts/extract-long-sequences      22     22      8      8     0%
scripts/extract-paired-reads        68     68     22     22     0%
scripts/extract-partitions         144    144     66     66     0%
scripts/fastq-to-fasta              33     33     12     12     0%
scripts/filter-abund                62     62     14     14     0%
scripts/filter-abund-single         67     67     14     14     0%
scripts/filter-stoptags             49     49     10     10     0%
scripts/find-knots                  67     67     12     12     0%
scripts/interleave-reads            73     73     26     26     0%
scripts/load-graph                  11     11      2      2     0%
scripts/load-into-counting          91     91     22     22     0%
scripts/make-initial-stoptags       53     53     10     10     0%
scripts/merge-partitions            42     42     10     10     0%
scripts/normalize-by-median        178    178     66     66     0%
scripts/partition-graph            103    103     26     24     2%
scripts/readstats                   90     90     18     18     0%
scripts/sample-reads-randomly       91     91     46     46     0%
scripts/split-paired-reads          79     79     26     26     0%
scripts/trim-low-abund             173    173     52     52     0%
------------------------------------------------------------------
TOTAL                             2656   1949    824    606    27%
```

ctb:
I'm confused by this comment; nothing has changed in this regard, I've just added tests.

--

ctb:
Err, just to be clear: nothing in this PR should affect existing code coverage.

--

ctb:
The tests are failing due to pylint stuff that I didn't change...

scripts/abundance-dist.py
mr-c: Your _output_ filename shouldn't be `/dev/stdin`. I think you meant `/dev/stdout`

scripts/count-median.py
```diff
@@ -91,7 +91,10 @@ def main():
     ksize = htable.ksize()
 
     print('writing to', output_filename, file=sys.stderr)
-    output = open(output_filename, 'w')
+    if output_filename in ('-', '/dev/stdout'):
+        output = sys.stdout
```
mr-c: Shouldn't the output argument on line 63 be set to argparse.FileType instead of this?

scripts/normalize-by-median.py
```diff
@@ -344,6 +344,11 @@ def main():  # pylint: disable=too-many-branches,too-many-statements
         else:
             output_name = args.single_output_file.name
         outfp = args.single_output_file
```
mr-c: The logic around single output which may-or-may not be STDOUT vs one output per input with '.keep' appended is a mess.

Since single_output_file is of `type=argparse.FileType('w')` then we can't re-open it (or cause it to be re-opened). So no setting `output_name = '/dev/stdout'`

What really should happen is that `if args.single_output_file:` then `outfp = args.single_output_file` and `output_name = args.single_output_file.name` end of story.

scripts/oxli
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+#
+# This file is a part of khmer, http://github.com/ged-lab/khmer/, and is
+# Copyright (C) Michigan State University, 2015. It is licensed under the
```
mr-c: New file. Copyright UC Davis.

scripts/sample-reads-randomly.py
```diff
@@ -108,6 +108,10 @@ def main():
         output_filename = output_file.name
     else:
         filename = args.filenames[0]
+        if filename == '/dev/stdin' or filename == '-':
```
mr-c: ` ('-' in args.input_filenames or '/dev/stdin' in args.input_filenames)` would be more consistent with other changes you've made

scripts/abundance-dist.py
ctb: yep.

scripts/normalize-by-median.py
ctb: The logic seems reasonably straightforward to me and I don't see how to simplify it easily.  I can't tell if I'm misreading it or if you're misreading it :). Help?

Also note that `.name` doesn't exist in certain cases that are hit by the tests. @bocajnotnef has been dealing with this by just applying str(args.single_output_file), which I guess works.

bocajnotnef: str(fileobj) doesn't actually work in a useful way.

```
In [1]: f = open('file', 'w')

In [2]: str(f)
Out[2]: "<open file 'file', mode 'w' at 0x7f703005a4b0>"
```

mr-c: I'll work on this with @bocajnotnef after this PR is merged

tests/test_normalize_by_median.py
```diff
@@ -336,10 +337,25 @@ def test_normalize_by_median_double_file_name():
 
     try:
         (status, out, err) = utils.runscript(script, args, in_dir)
+        raise Exception("Shouldn't get to this")
```
mr-c: `assert 0` would be kinder

ctb: See next  'except'. :)

mr-c: Right, that makes me un-happy too :-( Why is runscript now throwing AssertionErrors?

tests/khmer_tst_utils.py
```diff
@@ -53,7 +53,27 @@ def cleanup():
     cleanup_list = []
 
 
+def scriptpath(scriptname='interleave-reads.py'):
```
mr-c: Why the default value for scriptname? Seems dangerous.

ctb: On Mon, Jul 27, 2015 at 09:48:23AM -0700, Michael R. Crusoe wrote:
> > @@ -53,7 +53,27 @@ def cleanup():
> >      cleanup_list = []
> >  
> >  
> > +def scriptpath(scriptname='interleave-reads.py'):
> 
> Why the default value for scriptname? Seems dangerous.

See comment immediately below.  We need something, but it doesn't matter what.
Better approaches welcome, but passing in exactly the scriptname we need
is ugly and will probably fall afoul of copy-pasta.

tests/khmer_tst_utils.py
```diff
@@ -53,7 +53,27 @@ def cleanup():
...
+    # note - it doesn't matter what the scriptname is here, as long as
+    # it's some khmer script present in this version of khmer.
+
+    path = os.path.join(os.path.dirname(__file__), "../scripts")
+
+    if os.path.exists(os.path.join(path, scriptname)):
+        return path
+
+    for path in os.environ['PATH'].split(':'):
+        if os.path.exists(os.path.join(path, scriptname)):
```
mr-c: For what circumstance do we need to search the system path? Seems dangerous.

ctb: On Mon, Jul 27, 2015 at 09:51:47AM -0700, Michael R. Crusoe wrote:
> > @@ -53,7 +53,27 @@ def cleanup():
> >      cleanup_list = []
> >  
> >  
> > +def scriptpath(scriptname='interleave-reads.py'):
> > +    "Return the path to the scripts, in both dev and install situations."
> > +
> > +    # note - it doesn't matter what the scriptname is here, as long as
> > +    # it's some khmer script present in this version of khmer.
> > +
> > +    path = os.path.join(os.path.dirname(__file__), "../scripts")
> > +
> > +    if os.path.exists(os.path.join(path, scriptname)):
> > +        return path
> > +
> > +    for path in os.environ['PATH'].split(':'):
> > +        if os.path.exists(os.path.join(path, scriptname)):
> 
> For what circumstance do we need to search the system path? Seems dangerous.

This is a refactor of existing code (merged by you, I believe :), so we were
already doing this.  I couldn't think of a better way. Punt to an issue if you
want to discuss, but this isn't new functionality.

ctb: And yes, searching the path is dangerous, and led to an annoying bug with the new oxli functionality when testing in dev mode on athyra (hence the oxli script addition)!

mr-c: Ah, I see; this was pulled up from `runscriptredirect`.

mr-c: The difference is that now we are consulting the path for all script running not just the piped streaming tests.

ctb: > The difference is that now we are consulting the path for all script running not just the piped streaming tests.

Uh, no, I don't think so?

https://github.com/dib-lab/khmer/blob/master/tests/khmer_tst_utils.py#L71

(All of this is just refactoring, nothing new.)

mr-c: Drat, you are right.

ctb: heh, yes, that's how I feel about the code.

tests/khmer_tst_utils.py
```diff
@@ -64,17 +84,16 @@ def _runscript(scriptname, sandbox=False):
...
         if sandbox:
-            paths = [os.path.join(os.path.dirname(__file__), "../sandbox")]
+            path = os.path.join(os.path.dirname(__file__), "../sandbox")
         else:
-            paths = [os.path.join(os.path.dirname(__file__),
-                                  "../scripts")]
-            paths.extend(os.environ['PATH'].split(':'))
-        for path in paths:
-            scriptfile = os.path.join(path, scriptname)
+            path = scriptpath()
```
mr-c: should be `path = scriptpath(scriptname)`

ctb: On Mon, Jul 27, 2015 at 10:51:16AM -0700, Michael R. Crusoe wrote:
> >          else:
> > -            paths = [os.path.join(os.path.dirname(__file__),
> > -                                  "../scripts")]
> > -            paths.extend(os.environ['PATH'].split(':'))
> > -        for path in paths:
> > -            scriptfile = os.path.join(path, scriptname)
> > +            path = scriptpath()
> 
> should be `path = scriptpath(scriptname)`

*shrug* we're using scriptpath() elsewhere for cleanliness reasons. I actually
think we should remove 'scriptname' from the argument list and just hardcode
it, but I left it in just in case.

ctb: Bear in mind that ultimately this will just be �oxli� all the time�

mr-c:
This branch fails the automated release testing: http://ci.oxli.org/job/khmer-releasetest-ondemand/4/label=linux/console

mr-c:
So the tests pass from Git downloaded source; from pip pointing at Git but fail to run after installing from a tarball.

```
cd ../testenv3/
source bin/activate
pip install -U setuptools==3.4.1
pip install khmer*tar.gz  # from `make dist`
pip install nose
tar xzf khmer*tar.gz
cd khmer*
make dist
make test
```

mr-c:
I've pushed a fix for the test breakage; testing now

mr-c:
Setuptools does indeed create a proper `oxli` entry point upon install, yes. Oxli was being tested before and the tests would fail if a `./setup.py install` had not been done prior due to that entry point needing to be created by setuptools.

All install mechanism use setuptools so the pkg_resources use will be valid however I agree that having a dummy script when a versioned one will be created by setuptools is rather messy.

ctb:
On Tue, Jul 28, 2015 at 10:23:59AM -0700, Michael R. Crusoe wrote:
> Setuptools does indeed create a proper `oxli` entry point upon install, yes. Oxli was being tested before and the tests would fail if a `./setup.py install` had not been done prior due to that entry point needing to be created by setuptools.
> 
> All install mechanism use setuptools so the pkg_resources use will be valid however I agree that having a dummy script when a versioned one will be created by setuptools is rather messy.

Well, wait - it's clearly necessary when in dev mode (developing and
running tests in the git repo directory), but I don't think it should be
distributed.  So, what about keeping scripts/oxli but ditching distribution of
it and removing the dependency of setuptools?

mr-c:
Setuptools has been a dependency for two years. Happy to keep script in tree but not install it (letting setuptools make its versioned version)

ctb:
On Tue, Jul 28, 2015 at 10:27:18AM -0700, Michael R. Crusoe wrote:
> Setuptools has been a dependency for two years. Happy to keep script in tree but not install it (letting setuptools make its versioned version)

What, in order to build khmer? Right, ok :).  Well, keep it in and if I run
into any problems we can revisit.

tests/test_normalize_by_median.py
mr-c: Fixed in 9cf880dff1bbb6e69df4a96bbd09d324a9f8eae5

tests/test_normalize_by_median.py
```diff
@@ -79,12 +79,8 @@ def test_normalize_by_median_unpaired_final_read():
 
     script = 'normalize-by-median.py'
     args = ['-C', CUTOFF, '-k', '17', '-p', infile]
-    try:
-        (status, out, err) = utils.runscript(script, args, in_dir)
-        raise Exception("Shouldn't get to this")
-    except AssertionError as e:
-        out = str(e)
-        assert "ERROR: Unpaired reads when require_paired" in out, out
+    (status, out, err) = utils.runscript(script, args, in_dir, fail_ok=True)
```
ctb: please add assertion on return code, too!

mr-c: Fixed globally, thanks.

ctb: great!

mr-c:
Jenkins, retest this please

mr-c:
Passes the tests except for code coverage; `make diff-cover` is clean

mr-c:
This looks good to me and I think it should be the next PR merged. What do you say @ctb ?

ctb:
Sure, is there any reason not to? :)

On Tue, Jul 28, 2015 at 03:04:44PM -0700, Michael R. Crusoe wrote:
> This looks good to me and I think it should be the next PR merged. What do you say @ctb ?
> 
> ---
> https://github.com/dib-lab/khmer/pull/1186#issuecomment-125766763
-- 

mr-c:
Jenkins, retest this please

mr-c:
The release notes are going to be murder, I can see that now. Merged!

ctb:
;) w00t!

mr-c:
Gah, Jenkins wasn't testing the correct commit (!!) and this broke the build. Thank you @luizirber for the heads up.

Issue #1185, opened 2015-07-21
move docstrings from _khmer.cc to khmer/__init__.py


mr-c:
Easier to write and will help pylint, pep257 and friends

PR #1187, opened 2015-07-21
Migrating docstrings from cpython to pythonland

bocajnotnef:
as per #1185

Join me on my journey to liberate the docstrings from the oppression of the cpython! Together we can return them to their glorious homeland!


khmer/__init__.py
```diff
@@ -287,4 +287,27 @@ class HLLCounter(_HLLCounter):
     """
 
     def __len__(self):
-        return self.estimate_cardinality()
+        """Return the current estimation."""
+        return self._estimate_cardinality()
+
```
mr-c: Please don't rename the CPython methods. Just call the method directly `return self.estimate_cardinality()`

standage:
Has recent reorganization rendered this PR unnecessary?

betatim:
Unsure, but I would wait for the cythonisation and then add docstrings to that instead of moving from cpython to python by refreshing this PR.

standage:
Closing as no longer relevant.

Issue #1034, opened 2015-05-27
Remove get_kadian_count functionality.


ctb:
No longer of interest.

mr-c:
Fixed in #1194

PR #1194, opened 2015-07-23
Save _n_occupied in graph structures (Hashtable and descendants)

### LINKED ISSUE #1093: Verify that n_occupied etc. numbers are correct and saved in both countgraph & nodegraph
See discussion in #429.

- [ ] functions should be inspected and tested;
- [ ] we need to save these functions in countgraph and nodegraph

ctb:
Also removed unused count-overlap.py and associated code.

Ultimately should fix #1093, but ... see discussion.

Also fixes replace-string fubar committed to tests/test_hashbits.py in bd9b52e1.


ctb:
OK, so here's the problem - 

there are two numbers that we track in Hashbits at the C++ level, `_n_occupied` and `_n_unique_kmers`.  The first is the number of hashtable bins that are nonzero (hashtable occupancy) and the second is the number of times we've had to flip a bit to 1 in at least one hashtable (so, an estimate of the number of unique k-mers in the data set).

We don't save either of them in save(...), so they get reset to 0 on a load - not good.  This is what #1093 is about.

But, now the problem is that _n_unique_kmers cannot properly be updated in update_from, which updates one Hashbits data structure from another.  I can't think of any appropriate behavior here.
(_n_occupied can be recalculated easily by just iterating across the hashtable entries.)

So I think we have three options:

1. have an _n_unique_kmers that is only accurate until you do an update_from, at which point we... reset it to zero? Blech.
2. drop the calculation of _n_unique_kmers and only report on occupancy, which is good enough to warn about memory usage (and from which we can probably estimate n_unique_kmers... hmmm).
3. suggest that people interested in estimates of number of unique k-mers use HyperLogLog to do so, which is saveable/loadable/mergable. Downside: extra compute/extra code when they want to do so.
4. simply include a HyperLogLog counter in the Hashtable data structure. Downside: extra compute, all the time.

I'm leading towards combined (2) and (3), but would welcome thoughts @mr-c @luizirber.  It would be interesting to think about how to estimate the number of unique k-mers from the hashtable occupancy, I 'spose...

Also, separate note: CountingHash computes n_occupied when asked, while Hashbits keeps n_occupied up to date as k-mers are added.  Any thoughts on which is the right way to go? I've been planning to go with the Hashbits approach, but that requires being a bit careful about thread safety.

ctb:
cc @camillescott.

I'll plan to go with (2), eliminating _n_unique_kmers, next time I work on this - won't be for a few days prob.

lib/khmer.hh
```diff
@@ -54,7 +54,7 @@ private:\
 #   define CIRCUM_MAX_VOL 200	// @CTB remove
 
 #   define SAVED_SIGNATURE "OXLI"
-#   define SAVED_FORMAT_VERSION 4
+#   define SAVED_FORMAT_VERSION 5
```
mr-c: Gah, do we need per-file-type versioning?

mr-c:
The PR as it is written (though not described) looks good to me.

lib/khmer.hh
ctb: On Tue, Jul 28, 2015 at 02:37:13PM -0700, Michael R. Crusoe wrote:
> > @@ -54,7 +54,7 @@ private:\
> >  #   define CIRCUM_MAX_VOL 200	// @CTB remove
> >  
> >  #   define SAVED_SIGNATURE "OXLI"
> > -#   define SAVED_FORMAT_VERSION 4
> > +#   define SAVED_FORMAT_VERSION 5
> 
> Gah, do we need per-file-type versioning?

Yeah, I kinda cringed. Maybe we could do *per release* file versioning,
so this would only be incremented after a release?

mr-c: That's only one step better but acceptable for this next release as we
broke everything

On Tue, Jul 28, 2015 at 2:41 PM C. Titus Brown <notifications@github.com>
wrote:

> In lib/khmer.hh
> <https://github.com/dib-lab/khmer/pull/1194#discussion_r35704979>:
>
> > @@ -54,7 +54,7 @@ private:\
> >  #   define CIRCUM_MAX_VOL 200	// @CTB remove
> >
> >  #   define SAVED_SIGNATURE "OXLI"
> > -#   define SAVED_FORMAT_VERSION 4
> > +#   define SAVED_FORMAT_VERSION 5
>
> On Tue, Jul 28, 2015 at 02:37:13PM -0700, Michael R. Crusoe wrote: > @@
> -54,7 +54,7 @@ private:\ > # define CIRCUM_MAX_VOL 200 // @CTB
> <https://github.com/CTB> remove > > # define SAVED_SIGNATURE "OXLI" > -#
> define SAVED_FORMAT_VERSION 4 > +# define SAVED_FORMAT_VERSION 5 Gah, do we
> need per-file-type versioning?
> Yeah, I kinda cringed. Maybe we could do *per release* file versioning, so
> this would only be incremented after a release?
>
> —
> <https://github.com/dib-lab/khmer/pull/1194/files#r35704979>.
>
-- 

ctb:
Discussion with @mr-c:

* modifying the file format to include n_occupied is a fine idea;
* on merge we can then recalculated n_occupied accurately;
* n_unique_kmers should still be reported but not saved, because it is useful and we have no code that breaks it currently in the scripts/;

The tentative conclusion is to not remove n_unique_kmers now but in the future look for appropriate workarounds, while dealing correctly with n_occupied.

ctb:
Working on #1034 (remove get_kadian_count) in this as well.

ctb:
TODO:

- [ ] add tests for counting n_occupied()
- [ ] fix broken countgraph tests
- [ ] review ChangeLog entries

mr-c:
Jenkins, retest this please

ctb:
OK, back to working on this.

ctb:
Note 15651da, in which number of unique k-mers reported had been systematically underestimated by CountingHash/Countgraph (so, e.g. load-into-counting unique k-mers reported was generally significantly underestimated)

ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

Ready for review & hopefully merge, @mr-c.

mr-c:
Net deletion of 479 lines; I love it. Thanks!

PR #1204, opened 2015-07-29
Streaming unique-kmers.py using consume_fasta

luizirber:
Revert `unique-kmers.py` changes and use consume_fasta again. I wrote a `write_record` equivalent in C++ for this, currently living in `lib/HLLCounter.cc`, but I think this should probably go in `lib/read_parsers.cc`.

(reopening #1177 now that #1176 was merged)

### DIFF
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -4440,13 +4440,16 @@ hllcounter_consume_string(khmer_KHLLCounter_Object * me, PyObject * args)
 }
 
 static PyObject * hllcounter_consume_fasta(khmer_KHLLCounter_Object * me,
-        PyObject * args)
+        PyObject * args, PyObject * kwds)
 {
     const char * filename;
     PyObject * output_records_o = NULL;
+    char * kwlist[] = {"filename", "stream_out", NULL};
+
     bool output_records = false;
 
-    if (!PyArg_ParseTuple(args, "s|O", &filename, &output_records_o)) {
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "s|O", kwlist,
+          &filename, &output_records_o)) {
         return NULL;
     }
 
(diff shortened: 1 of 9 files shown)


ctb:
cc #1053

luizirber:
Update: I moved `write_record` to `Read.write_to`, it takes an `ostream` as argument (I'm using `std::cout` on `HLLCounter.consume_fasta`).

luizirber:
- [x] Is it mergeable?
- [x] Did it pass the tests?
  fixing as needed.
  changes were made?

luizirber:
ready for review @ctb @mr-c @bocajnotnef

khmer/_khmer.cc
```diff
@@ -4575,16 +4575,22 @@ static PyObject * hllcounter_consume_fasta(khmer_KHLLCounter_Object * me,
         PyObject * args)
 {
     const char * filename;
+    PyObject * output_records_o = NULL;
+    bool output_records = false;
 
-    if (!PyArg_ParseTuple(args, "s", &filename)) {
+    if (!PyArg_ParseTuple(args, "s|O", &filename, &output_records_o)) {
```
luizirber: Should `output_records` be a keyword on this method? I think it makes more sense, because calling it is a bit weird (check https://github.com/dib-lab/khmer/pull/1204/files#diff-7f09af630bd7b293c0827bba1d411617R120 for an example).

This also raises the question: do we want to do this in other places? We do have a lot of default arguments in the extension glue code that are not exposed nicely to Python land...

lib/read_parsers.cc
```diff
@@ -20,6 +20,21 @@ namespace khmer
 namespace read_parsers
 {
 
+void
+Read::write_to(std::ostream& output)
```
luizirber: I think `write_to` is a better name than `write_record`, since the invocation is something like `read.write_to(std::cout)`. Thoughts?

mr-c: +1

khmer/_khmer.cc
mr-c: keyword is fine by me; how about a post-2.0 issue for the 2nd question?

luizirber:
jenkins, test this please

mr-c:
LGTM. Merge once the tests pass.

Issue #1117, opened 2015-06-23
Implementing --max-mem and auto'd hashtable agrs


bocajnotnef:
Related to: #1050, #390, #1106, #732 

Currently: @ctb is implementing --max-mem in #1050 where, given a memory ceiling, automagically calculates graph sizes

#1106 also has methods for automagically calculating graph sizes given number of unique kmers, memory ceiling and/or desired FP rate. 

So, my question: is it worth including the methods added in #1106 more widely if #1050 is going to address some/most of the usability?

If so, I move that the argparser stuff currently here https://github.com/dib-lab/khmer/pull/1106/files#diff-c9f39ec665e36b5911296d1f4028270cR61 gets moved into the oxli module so it can be added to scripts easily (a la threading args). Some additional processing stuff will need to be added to the scripts that implement this but it should be along the lines of "if memory options then use returned values"

Sticking points:
 # making sure #1106 doesn't overlap with #1050 in terms of function--#1050 has --max-mem already, so #1106 might be limited to tablesize estimations and FP-rate based estimations.
 # implementing the size calculations via argparser without duplicating code--Should be fairly easy to just use the added oxli functions but we all know how "fairly easy" stuff tends to go when I get involved.

@ctb @camillescott @luizirber @mr-c @macmanes @kdmurray91 @drtamermansour Thoughts?

ctb:
On Tue, Jun 23, 2015 at 03:53:21PM -0700, Jake Fenton wrote:
> Related to: #1050, #390, #1106, #732 
> #1106 also has methods for automagically calculating graph sizes given number of unique kmers, memory ceiling and/or desired FP rate. 
> 
> So, my question: is it worth including the methods added in #1106 more widely if #1050 is going to address some/most of the usability?

Question: from reading through #1106 it looks to me like all of your methods
require some information (desired FP rate, or number of k-mers) along with
desired memory usage.  Is that correct?

bocajnotnef:
Yes. Specifically, number of kmers paired with either desired FP or desired memory.

ctb:
On Tue, Jun 23, 2015 at 04:00:45PM -0700, Jake Fenton wrote:
> Yes. Specifically, number of kmers paired with either desired FP or desired memory.

So in that case I see no conflict in the feature sets.  I think #1050 will
be (is already) a sound platform on which to build in your additional
functionality.  No?

ctb:
[...]

Issue #1146, opened 2015-07-06
hashtable arg estimation functions may be wrong


bocajnotnef:
see #1106 #1117 

```
(env)1 bocajnotnef@GCU-Caconym:~/git/khmer⟫ python scripts/normalize-by-median.py -U 83 -x 17 tests/test-data/test-filter-abund-Ns.fq                                                         

|| This is the script 'normalize-by-median.py' in khmer.
|| You are running khmer version 2.0-rc1+53.gf24a2e0.dirty
|| You are also using screed version 0.9
||
|| If you use this script in a publication, please cite EACH of the following:
||
||   * MR Crusoe et al., 2014. http://dx.doi.org/10.6084/m9.figshare.979190
||   * CT Brown et al., arXiv:1203.4802 [q-bio.GN]
||
|| Please see http://khmer.readthedocs.org/en/latest/citations.html for details.


PARAMETERS:
 - kmer size =    32            (-k)
 - n tables =     4             (-N)
 - max tablesize =    17        (-x)

Estimated memory usage is 68 bytes (n_tables x max_tablesize)
--------
*** Warning: The given tablesize is be too small!
*** Estimating false positive rate to be 0.100049378972
making countgraph
DONE with tests/test-data/test-filter-abund-Ns.fq; kept 4 of 4 or 100%
output in test-filter-abund-Ns.fq.keep
Total number of unique k-mers: 3
**
** ERROR: the graph structure is too small for 
** this data set.  Increase data structure size
** with --max_memory_usage/-M.
**
** Do not use these results!!
**
** (estimated false positive rate of 45.698;
max allowable 0.800
```

TL;DR: Estimation funcs say we'll have an FP of  0.100049378972
Script says we have an FP of 45.698

One of those algorithms is horribly wrong. 

~~@ctb @luizirber I have no idea how the math behind any of these functions work--Advice, please?~~

@camillescott  gave enlightenment--pondering required.

ctb:
What was the enlightenment? :)

bocajnotnef:
wiki article on how bloom filters worked.

mr-c:
@bocajnotnef  Is this still an issue?

bocajnotnef:
Unsure--Need to do further investigating.

ctb:
Addressed in #1214.

PR #1214, opened 2015-07-30
Cleaning up automatic argument setting

### LINKED ISSUE #1179: Deal appropriately with --unique-kmers argument
The addition of `-U/--unique-kmers` should not have applied to *all scripts* (see https://github.com/dib-lab/khmer/pull/1126/files#diff-2ebccdb74ba5df72b22cff1bd70d9696R63) if it was going to be ignored by all of them!  We should either make it work for all scripts (which should be easy after #1050, no?) or move the parameter addition into the one bit of code that uses it on `oxli/functions.py`.

SensibleSalmon:
Ref #1117 for expansion of automatic memory usage determination, this time based on estimated number of unique k-mers. Inspired by user experience upgrade goals (#732) and hopefully will fix #1179 (dealing with `-U/--unique-kmers`).

This resolves @ctb's comment https://github.com/dib-lab/khmer/issues/1117#issuecomment-126351063:

> OK, well, let's start a new pull request :). Here are a few requested changes:
> - change do_sanity_checking name to check_fp_rate
> - should this function only be called if -U is set? For now, I think so...
> - change oxutils shortname to oxfuncs, since it's `oxli.functions' you're importing.
> - get rid of "optimization" language in comments - it's automatic argument picking, not optimization :)
>   (The goal of all of this is to enable script readability.)
> 
> Is there a reason that this code isn't called generically in create_countgraph/other khmer_args functions? That seems like a better place for it than adding it into each script individually, but I'm probably missing something.

My response to the last: 

> The desired_fp rate varies from script to script, which is why we currently call it on a script-by-script basis--that said, it shouldn't be too hard to pass that along to create_countgraph.

Next steps TBD


oxli/functions.py
```diff
@@ -112,14 +112,13 @@ def optimal_args_output_gen(unique_kmers, fp_rate):
     return "\n".join(to_print)
 
 
-def do_sanity_checking(args, desired_max_fp):
+def check_fp_rate(args, desired_max_fp):
     """
     simple function to check if the restrictions in the args (if there are any)
```
ctb: (please also update this - it's not simple anymore :) and also is currently limited to looking at unique_kmers.

Also, you can deindent most of the function by changing that first 'if' to

```
if not args.unique_kmers:
   return

File: oxli/functions.py [hunk 2]
```diff
@@ -144,7 +143,7 @@ def do_sanity_checking(args, desired_max_fp):
                 print("*** Estimating false positive rate to be {0}".format(
                     res.fp_rate), file=sys.stderr)
             else:
-                print("*** INFO: set memory ceiling using auto optimization.",
+                print("*** INFO: set memory ceiling automatically.",
```
ctb: 'from estimated number of unique kmers'

ctb:
I like this code so far; nice work. I think the next step should be to encapsulate the call to oxli.functions in khmer.khmer_args create* functions, with a default fp_rate of 0.10. That should then implement it for all the scripts without further modifications, yah?

Can you also add a long argument --fp-rate (no short version) that lets people override script-configured fp-rate from the command line?

That should let us do:

     normalize-by-median.py --fp-rate 0.8 --unique-kmers 1e7 ...

as a way of setting memory parameters. Make sense?

bocajnotnef:
Yup, sounds good. I'll get that rolling.

ctb:
Be sure to remove this: https://github.com/dib-lab/khmer/blob/master/scripts/normalize-by-median.py#L321 and make that a general warning in khmer_args.py.

bocajnotnef:
Should we sanity check for things like ending up with a negative amount of reccomended memory? 
cause I was messing around to test the args and the function happily reported it set the memory ceiling to -238 bytes.

Somehow, this then resulted in an FP rate of 2.08.

bocajnotnef:
What's strange is if I set desired fp rate to be 0.004 or 0.00004 the estimated FP rate comes out to be 0.208

ctb:
well, the equations are approximations. so for small numbers of k-mers weird
things may happen.

I would say, if -x is going to be less than 1 mb in size, increase it to 
1 mb in size.

On Mon, Aug 03, 2015 at 10:38:37AM -0700, Jake Fenton wrote:
> What's strange is if I set desired fp rate to be 0.004 or 0.00004 the estimated FP rate comes out to be 0.208
> 
> ---
> https://github.com/dib-lab/khmer/pull/1214#issuecomment-127341822
-- 

bocajnotnef:
Currently, we check to see if we have `args.unique_kmers` in the scripts and in the function--we should stick to one location.

I vote for doing the check in the function and have just all scripts do `args = oxfuncs.check_fp_rate(args, fp)`

ctb:
On Mon, Aug 03, 2015 at 02:00:07PM -0700, Jake Fenton wrote:
> Currently, we check to see if we have `args.unique_kmers` in the scripts and in the function--we should stick to one location.
> 
> I vote for doing the check in the function and have just all scripts do `args = oxfuncs.check_fp_rate(args, fp)`

Why do you vote that?  (I agree, just curious as to your reasoning.)

bocajnotnef:
Reduces points of failure to a single place--less repeated code in the scripts. If we're gonna do the same check in fifteen places to run a thing, just have the thing do the check.

ctb:
On Mon, Aug 03, 2015 at 02:07:20PM -0700, Jake Fenton wrote:
> Reduces points of failure to a single place--less repeated code in the scripts. If we're gonna do the same check in fifteen places to run a thing, just have the thing do the check.

+1. Let's see how it works in practice :)

bocajnotnef:
@ctb Updated--Should I further expand to other htable-using scripts?

ctb:
BTW, please make --unique-kmers a float - that lets us use scientific notation to set it, e.g. `-U 1e9`.

oxli/functions.py
```diff
@@ -112,43 +112,66 @@ def optimal_args_output_gen(unique_kmers, fp_rate):
...
 
 
-def do_sanity_checking(args, desired_max_fp):
+def check_fp_rate(args, desired_max_fp):
     """
-    simple function to check if the restrictions in the args (if there are any)
-    make sense--If not, complain. If no restrictions are given, add some that
-    make sense.
-    Takes in args and desired max FP rate
+    function to check if the desired_max_fp rate makes sense given specified
```
ctb: Capitalize 'function'

ctb:
Please see https://github.com/dib-lab/khmer/pull/1214#issuecomment-126696076; was there a reason you didn't go with that for normalize-by-median?

bocajnotnef:
Oh, nuts. Nope, no reason. I just had that conversation with Michael, too.

I'll pull the function call out to the create* functions.

bocajnotnef:
Also, should I be making use of the logging framework we added? 'cause with the addition of the autoarg stuff we break normalize being quiet.

bocajnotnef:
(which also reminds me, the tests for normalize being quiet don't actually assert that there be nothing in the output, just that there aren't certian things in the output)

ctb:
On Tue, Aug 04, 2015 at 11:28:46AM -0700, Jake Fenton wrote:
> (which also reminds me, the tests for normalize being quiet don't actually assert that there be nothing in the output, just that there aren't certian things in the output)

No, I fixed that.

ctb:
On Tue, Aug 04, 2015 at 11:28:23AM -0700, Jake Fenton wrote:
> Also, should I be making use of the logging framework we added? 'cause with the addition of the autoarg stuff we break normalize being quiet.

Please fix -q on normalize-by-median, but don't institute it in other
scripts, please.

bocajnotnef:
I'm running into circular import issues between khmer_args and oxli/__init__. Apparently this is indicative of needing to refactor package structure--Which I'm inclined to agree with.

So, unless there's mass objection, I'm gonna migrate the autoarg stuff out of oxli and into khmer_args

ctb:
On Tue, Aug 04, 2015 at 11:54:26AM -0700, Jake Fenton wrote:
> I'm running into circular import issues between khmer_args and oxli/__init__. Apparently this is indicative of needing to refactor package structure--Which I'm inclined to agree with.
> 
> So, unless there's mass objection, I'm gonna migrate the autoarg stuff out of oxli and into khmer_args

OK short term fix.

bocajnotnef:
long term punt to issue?

ctb:
On Tue, Aug 04, 2015 at 11:55:53AM -0700, Jake Fenton wrote:
> long term punt to issue?

yep.

ctb:
(but it will be obvious, no need to punt to an issue; part of oxli
refactoring)

bocajnotnef:
Alright. Cool.

bocajnotnef:
Problem: if we call the check_fp_rate function from the funcs that create hashtables we won't be able to warn people if they're loading a hashtable AND using the automatic args (since we won't ever create a hashtable if we're loading one)

Suggested solution: we make the --load-table arg and the -U arg (and/or the --max-mem arg) mutually exclusive in argparse.

ctb:
Sounds like a plan.

bocajnotnef:
new problem: the args I need to make mutually exclusive are added in two separate places.

I'm gonna eat and ponder

bocajnotnef:
Alright. Making a mutually exclusive group is not feasible. Michael reccomended making a function to do the checking and to just call that function somewhere in all the scripts but I feel like if there's the solution we should just revert to the behaviour where we call the check_fp_rate function in all the scripts.

OPen to suggestions.

bocajnotnef:
fix by adding "you're doing silly things" logic to the check_load_table (ish) thing in khmer_args

bocajnotnef:
@ctb Moved the stuff to the place and fixed the things

Someday that'll be a commit message.

scripts/normalize-by-median.py
```diff
@@ -26,7 +26,7 @@
 import textwrap
 from khmer import khmer_args
 from contextlib import contextmanager
-from oxli import functions as oxutils
+from oxli import functions as oxfuncs
```
ctb: no longer needed or used.

khmer/khmer_args.py
```diff
@@ -44,6 +47,153 @@ class ComboFormatter(argparse.ArgumentDefaultsHelpFormatter,
...
+        res = estimate_optimal_with_K_and_f(args.unique_kmers,
+                                            desired_max_fp)
+        if args.max_tablesize and args.max_tablesize < res.htable_size:
+            print("*** Warning: The given tablesize is too small!",
+                  file=sys.stderr)
+            print("*** Estimating false positive rate to be {0} from estimated"
+                  " number of unique kmers ".format(res.fp_rate,
+                                                    file=sys.stderr))
+        else:
+            if res.mem_use < 1000000:  # one megabyteish
```
ctb: perhaps best to use 1e6 here, for readability.

khmer/khmer_args.py
```diff
@@ -44,6 +47,153 @@ class ComboFormatter(argparse.ArgumentDefaultsHelpFormatter,
...
+        math.log(1 - des_fp_rate ** (1 / float(int_n_tables)))))
+    mem_cap = ht_size * int_n_tables
+    fp_rate = (1 - math.exp(-num_kmers / float(ht_size))) ** int_n_tables
+
+    res = namedtuple("result", ["num_htables", "htable_size", "mem_use",
+                                "fp_rate"])
+    return res(int_n_tables, ht_size, mem_cap, fp_rate)
+
+
+def optimal_args_output_gen(unique_kmers, fp_rate):
```
ctb: optimal_args_report maybe?

oxli/build_graph.py
```diff
@@ -52,8 +52,9 @@ def main(args):
     for fname in args.input_filenames:
         check_input_files(fname, args.force)
 
-    # if optimization args are given, do optimization
-    args = functions.do_sanity_checking(args, 0.01)
+    # if automatic arg info is given, set args automagically
```
ctb: Is this still needed since we use create_nodegraph below?

khmer/khmer_args.py
```diff
@@ -149,6 +311,7 @@ def calculate_tablesize(args, hashtype, multiplier=1.0):
 
 
 def create_nodegraph(args, ksize=None, multiplier=1.0):
+    args = check_fp_rate(args)
```
ctb: be sure to allow passthru of optimal fp rate into create_nodegraph.

sandbox/optimal_args_hashbits.py
```diff
@@ -20,11 +20,11 @@
...
 
 import khmer
-from khmer.khmer_args import build_hashbits_args
-from khmer.khmer_args import (report_on_config, info, add_threading_args)
+from khmer.khmer_args import (report_on_config, info, add_threading_args,
+                              build_hashbits_args)
 from khmer.kfile import check_input_files, check_space
 from khmer.kfile import check_space
-from oxli.functions import optimal_args_output_gen as output_gen
+from khmer.khmer_args import optimal_args_output_gen as output_gen
```
ctb: in general, try to avoid renaming functions this way - makes finding them tricky.

khmer/khmer_args.py
```diff
@@ -110,6 +263,15 @@ def __call__(self, parser, namespace, values, option_string=None):
 ** Your values for ksize, n_tables, and tablesize
 ** will be ignored.'''.format(hashfile=values))
 
+            if getattr(namespace, 'unique_kmers') != 0 or \
+               getattr(namespace, 'max_memory_usage') is not None or \
```
ctb: you can remove 'is not None'. 0 (which is the alternative 'false' here) is not a sane argument either.

khmer/khmer_args.py
```diff
@@ -110,6 +263,15 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+               getattr(namespace, 'max_tablesize') is not None:
+
+                print("\n*** WARNING: you have specified a number of unique"
```
ctb: How about:

you have asked that the graph size be calculated automatically (-U or -M), 
but are loading an existing graph. Size will NOT be set automatically.

khmer/khmer_args.py
```diff
@@ -110,6 +263,15 @@ def __call__(self, parser, namespace, values, option_string=None):
 ** Your values for ksize, n_tables, and tablesize
 ** will be ignored.'''.format(hashfile=values))
 
+            if getattr(namespace, 'unique_kmers') != 0 or \
+               getattr(namespace, 'max_memory_usage') is not None or \
+               getattr(namespace, 'max_tablesize') is not None:
```
ctb: why max_tablesize here?

khmer/khmer_args.py
bocajnotnef: I'm conflicted on how to manage the fp rate setting--Currently, all scripts default to an fp of 0.1 but that can be overidden via the --fp-rate arg, which is in the args object and is passed along.

If we want to actually pass in a default rate we'd have to add an arg to at least both create functions--Which isn't necessarily bad, but I'd like to avoid it if possible.

We could have the scripts instead modify the fp_rate attribute of the args... Though I don't feel like that's a good solution either

ctb: On Wed, Aug 05, 2015 at 10:03:57AM -0700, Jake Fenton wrote:
> > @@ -149,6 +311,7 @@ def calculate_tablesize(args, hashtype, multiplier=1.0):
> >  
> >  
> >  def create_nodegraph(args, ksize=None, multiplier=1.0):
> > +    args = check_fp_rate(args)
> 
> I'm conflicted on how to manage the fp rate setting--Currently, all scripts default to an fp of 0.1 but that can be overidden via the --fp-rate arg, which is in the args object and is passed along.
> 
> If we want to actually pass in a default rate we'd have to add an arg to at least both create functions--Which isn't necessarily bad, but I'd like to avoid it if possible.
> 
> We could have the scripts instead modify the fp_rate attribute of the args... Though I don't feel like that's a good solution either

There's science in them thar hills, so we need something like the following:

* if users supply fp-rate, assume they know what they are doing;
* otherwise, default to the one specified by the script;
* if one isn't specified by the script, default to the global one;

So, yes, we need an arg added to both create functions; there's no avoiding it.

bocajnotnef: Alright. Moar args it is.

bocajnotnef:
Found a problem.

So, we moved the code to check if we have automatic args (-U and/or -M) and --loadtable into the function that handles if loadtable is called.

That doesn't work if I structure my command: `normalize-by-median.py --loadtable table -U 40 -M 50 somefile`

It executes the function to handle --loadtable before it parses -U and -M into the namespace.

ctb:
On Wed, Aug 05, 2015 at 11:21:29AM -0700, Jake Fenton wrote:
> Found a problem.
> 
> So, we moved the code to check if we have automatic args (-U and/or -M) and --loadtable into the function that handles if loadtable is called.
> 
> That doesn't work if I structure my command: `normalize-by-median.py --loadtable table -U 40 -M 50 somefile`
> 
> It executes the function to handle --loadtable before it parses -U and -M into the namespace.

Hey, super cool!  Nice catch.

Does that mean we need a separate check for loadtable args in the fp rate
fn?

bocajnotnef:
Yeah, which is doable, it just needs to be wrapped in a try/except (since the args object won't always have --loadtable as an option)

ctb:
On Wed, Aug 05, 2015 at 11:24:11AM -0700, Jake Fenton wrote:
> Yeah, which is doable, it just needs to be wrapped in a try/except (since the args object won't always have --loadtable as an option)

c.f. 'getattr'.

bocajnotnef:
should I remove the check in LoadAction, then, since there isn't a guaruntee of it firing?

bocajnotnef:
Drat, wait, we can't put it in the fp rate function 'cause it doesn't get called if we load a table.

Back to the original issue.

bocajnotnef:
I'm tempted to just move the check_fp_rate call into report_on_config (since we do that everywhere) and just have report_on_config return a modified arg object.

bocajnotnef:
Or if we wanted to be SUPER insane-o we could add a function similar to LoadAction to -U/-M that runs the same check as --loadtable so if either one is called first it'll do the check.

ctb:
On Wed, Aug 05, 2015 at 11:32:22AM -0700, Jake Fenton wrote:
> Or if we wanted to be SUPER insane-o we could add a function similar to LoadAction to -U/-M that runs the same check as --loadtable so if either one is called first it'll do the check.

+1

bocajnotnef:
getattr still borks for me if the stuff isn't there

```
  File "/home/bocajnotnef/git/khmer/khmer/khmer_args.py", line 205, in __call__
    if getattr(namespace, 'loadtable'):
AttributeError: 'Namespace' object has no attribute 'loadtable'

```

ctb:
On Wed, Aug 05, 2015 at 11:49:24AM -0700, Jake Fenton wrote:
> getattr still borks for me if the stuff isn't there
> 
> ```
>   File "/home/bocajnotnef/git/khmer/khmer/khmer_args.py", line 205, in __call__
>     if getattr(namespace, 'loadtable'):
> AttributeError: 'Namespace' object has no attribute 'loadtable'

see: documentation on getattr.

bocajnotnef:
Aha.

bocajnotnef:
retest this, please

bocajnotnef:
@ctb That was less insane than I was expecting. 

- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

Ready for merge

khmer/khmer_args.py
```diff
@@ -110,6 +272,16 @@ def __call__(self, parser, namespace, values, option_string=None):
 ** Your values for ksize, n_tables, and tablesize
 ** will be ignored.'''.format(hashfile=values))
 
+            if getattr(namespace, 'unique_kmers') != 0 or \
```
ctb: is there a reason not to use 'namespace.unique_kmers' and 'namespace.max_memory_usage' here?

sandbox/optimal_args_hashbits.py
```diff
@@ -84,7 +84,7 @@ def main():
         if not False:
             sys.exit(1)
 
-    to_print = output_gen(unique_kmers,fp_rate)
+    to_print = optimal_args_report(unique_kmers, fp_rate)
```
ctb: how about graphsize_args_report?

khmer/khmer_args.py
bocajnotnef: I was just using the syntax I saw elsewhere. Referencing the attributes directly should be safe here.

bocajnotnef:
@ctb Updated

khmer/khmer_args.py
```diff
@@ -110,6 +272,15 @@ def __call__(self, parser, namespace, values, option_string=None):
 ** Your values for ksize, n_tables, and tablesize
 ** will be ignored.'''.format(hashfile=values))
 
+            if namespace.unique_kmers != 0 or namespace.max_memory_usage:
```
ctb: eliminate != 0

ctb:
This looks pretty nice and clean to me, but I may not have a chance to dig into it today.  My main comment is that I think the script-level testing should be expanded a bit - think about it and let me know what you decide :).

bocajnotnef:
I'm running into a lot of problems with having things like the LoadAction calls--or any function that gets called when there's an arg. It introduces too many cases to keep track of (for such things as handling args that shouldn't be set together or things like handling--quiet, especially since we can only configure the logger once we parse the args but when we parse the args we spew output everywhere).

Working around having these actions is beginning to get non-trivial, so I'm going to factor it all out into a check_conflicting_args function that I'm gonna have something or another call (maybe report_on_config...?)

bocajnotnef:
(something tells me doing this refactor will be non-trivial regardless)

bocajnotnef:
Realizing this is probably what I was supposed to do regardless.

Oh well.

bocajnotnef:
retest this, please

tests/test_normalize_by_median.py
```diff
@@ -78,7 +93,7 @@ def test_normalize_by_median_quiet():
     shutil.copyfile(utils.get_test_data('test-abund-read-2.fa'), infile)
 
     script = 'normalize-by-median.py'
-    args = ['-C', CUTOFF, '-k', '17', '--quiet', infile]
+    args = ['-C', CUTOFF, '-k', '17', '--quiet', '-M', '1e6', infile]
```
bocajnotnef: Because I made it so report_on_config executed again I had to increase the tablesize to stop it from warning during the quiet test. Normally, I'd then go "But it shouldn't make any noise when silenced anwyay" but these are warnings, which as I recall we wanted to ignore --quiet.

bocajnotnef:
retest this, please

bocajnotnef:
@mr-c Ready for merge

ctb:
Misc review thoughts
- [ ] do we need to update doc/user/scripts.txt globally somehow, or does this just happen automatically?
- [ ] look closely at the various complaint functions & add_loadhash_args
- [ ] do we need to override default FPR anywhere in the scripts?
- [ ] check -q behavior on normalize-by-median
- [ ] read and understand https://github.com/dib-lab/khmer/pull/1214/files#r36478352
- [ ] fix misspelling of `Reccomended`
- [ ] remove `umers` abbreviation viz. `test_oxli_build_graph_umers_arg` & elsewhere

Train of thought --

normalize-by-median code removal is good!

ctb:
(I will do these things)

ctb:
Nicely done; I like the cleaned up loadhash-checking code, in particular.

ctb:
Nicely done; I like the cleaned up loadhash-checking code, in particular.

ctb:
Nicely done; I like the cleaned up loadhash-checking code, in particular.

ctb:
Nicely done; I like the cleaned up loadhash-checking code, in particular.

ctb:
Nicely done; I like the cleaned up loadhash-checking code, in particular.

ctb:
Fixes #1117 and #1179, addresses issues in UX (was #732) & edge cases in #1146.

bocajnotnef:
Thanks!

PR #1235, opened 2015-08-08
suppress display of -x and -N on argparse help

### LINKED ISSUE #1232: Can we remove -x and -N options from -h and doc/user/scripts.txt output?
With the new -M and -U we don't need them any more and I think we would prefer they be hidden.

ping @mr-c

ctb:
Fixes #1232.

### DIFF
--- a/tests/test_script_arguments.py
+++ b/tests/test_script_arguments.py
@@ -122,6 +122,34 @@ def test_check_tablespace_nodegraph(graph_type, exp_buckets):
     assert sizestr == exp_buckets
 
 
+def test_normal_help(capsys):
+    # check -x and -N are hiden by default with --help
+    parser = khmer_args.build_graph_args()
+
+    with pytest.raises(SystemExit):
+        parser.parse_args(['-h'])
+
+    out, err = capsys.readouterr()
+    assert "--max-tablesize" not in out
+    assert '--n_tables' not in out
+
+
+def test_expert_help(capsys):
+    # check -x and -N are hidden by default but appear with --help-expert
+    old_argv = sys.argv[:]
+    sys.argv.append('--help-expert')
+    parser = khmer_args.build_graph_args()
+
+    with pytest.raises(SystemExit):
+        parser.parse_args(['-h', '--help-expert'])
+
+    out, err = capsys.readouterr()
+    assert "--max-tablesize" in out
+    assert '--n_tables' in out
+
+    sys.argv = old_argv
+
+
 def test_check_space_force():
     fakelump_fa = utils.get_test_data('fakelump.fa')
 
--- a/tests/test_script_arguments.py
+++ b/tests/test_script_arguments.py
@@ -123,7 +123,7 @@ def test_check_tablespace_nodegraph(graph_type, exp_buckets):
 
 
 def test_normal_help(capsys):
-    # check -x and -N are hiden by default with --help
+    # check -x and -N are hidden by default with --help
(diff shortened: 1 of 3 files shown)


ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

ctb:
Ready for review & merge, @mr-c. I think this is the final planned code change, yes?

ctb:
retest this please, jenkins.

mr-c:
Re: https://github.com/dib-lab/khmer/issues/1232#issuecomment-128987924 I'd rather not supress these options at this time.

Maybe amend them to remind people that there is an easier way?

mr-c:
Bumping to 2.0+

standage:
@ctb So to be clear, the idea is to hide `-x/-N` for typical user but have some kind of verbose help that includes a description of the "power" options?

ctb:
On Mon, Jun 27, 2016 at 08:10:41PM -0700, Daniel Standage wrote:
> @ctb So to be clear, the idea is to hide `-x/-N` for typical user but have some kind of verbose help that includes a description of the "power" options?

Yep.

betatim:
Refreshed the PR. It looked like this was basically ready to go and then got forgotten?

ctb:
We left it here: https://github.com/dib-lab/khmer/pull/1235#issuecomment-129261058 - we didn't like the SUPPRESS approach. Alternatives?

betatim:
Ah, that is a fail on reading comprehension for me :)

What you want is to hide it by default for `blah.py --help` but show it when you do `blah.py --HELP` (or what ever)? Off the top of my head the way to do this is to peek at `argv` and detect that the flag for "really help me this time" is present and swap from suppressing to showing the help message. Pretty hacky.

> Maybe amend them to remind people that there is an easier way?

Seems much easier/less likely to break.

ctb:
On Thu, Nov 17, 2016 at 07:01:27AM -0800, Tim Head wrote:
> Ah, that is a fail on reading comprehension for me :)
> 
> What you want is to hide it by default for `blah.py --help` but show it when you do `blah.py --HELP` (or what ever)? Off the top of my head the way to do this is to peek at `argv` and detect that the flag for "really help me this time" is present and swap from suppressing to showing the help message. Pretty hacky.
> 
> > Maybe amend them to remind people that there is an easier way?
> 
> Seems much easier/less likely to break.

or,

* have the default output (when run with no args) not show them;
* have --help show them under "advanced arguments"

or something like that.

You COULD do something like use args.SUPPRESS, and then add an argument
--help-expert, which shows all the arguments you're not supposed to use
except in case of emergency... hmm, I like that.

betatim:
Added `--help-expert`, somewhat hacky. Ideas how to make it nicer? Otherwise ready for review @ctb

standage:
I think this should go into 2.1.

betatim:

tests/test_script_arguments.py
```diff
@@ -122,6 +122,34 @@ def test_check_tablespace_nodegraph(graph_type, exp_buckets):
     assert sizestr == exp_buckets
 
 
+def test_normal_help(capsys):
+    # check -x and -N are hiden by default with --help
```
standage: hiden --> hidden

betatim:
Shall we wait for @ctb to hit the button then?

standage:
Nope, just had to wait for the build to complete after my typo fix. Merging now.

Issue #144, opened 2013-08-31
Add good tests for scripts/extract-partitions.py & refactor


ctb:
This is a really ugly file... apologies in advance.

ctb:
No longer claimed.

bocajnotnef:
I can take this on if extract-partitions still needs a refactor.

ctb:
On Mon, Aug 10, 2015 at 11:03:57AM -0700, Jake Fenton wrote:
> I can take this on if extract-partitions still needs a refactor.

please!

bocajnotnef:
Looking at the CI coverate ( http://ci.oxli.org/job/khmer-master/418/label=linux/cobertura/_default_/scripts_extract_partitions_py/ ) I'm seeing near-compelte coverage on extract-partitions, so I'm just going to go ahead with the refactor.

ctb:
On Thu, Aug 13, 2015 at 10:10:13AM -0700, Jake Fenton wrote:
> Looking at the CI coverate ( http://ci.oxli.org/job/khmer-master/418/label=linux/cobertura/_default_/scripts_extract_partitions_py/ ) I'm seeing near-compelte coverage on extract-partitions, so I'm just going to go ahead with the refactor.

For testing on this kind of thing, I would like to ask that (prior to merge)
we:

* identify a few different data sets to run the through do-partition etc.
  (the ones in data/ are probably ok)
* run extract-partitions, old version
* run extract-partitions, new version
* make sure results are bytewise identical, OR that we understand
  *exactly* why they differ.

(good test coverage is not enough for correctness.)

bocajnotnef:
Okay. Sounds good.

bocajnotnef:
Ran a modified variant of the "partitioning large datasets" ( https://khmer.readthedocs.org/en/v1.4.1/user/partitioning-big-data.html#running-on-an-example-data-set ) over the 25k Casava reads in /data, using master's extract-partition and the refactored version. Diffing the dist files and the group files show's they're the same.

bocajnotnef:
(modified script here for reference: https://gist.github.com/bocajnotnef/0bd649939c60a7b7ed72)

ctb:
On Fri, Aug 14, 2015 at 10:34:10AM -0700, Jake Fenton wrote:
[...]

PR #1251, opened 2015-08-13
Refactor extract-partitions

bocajnotnef:
Generators galore! Classes are fun! Wooo!

Ref #144 (man, that's old)

TO-DO:
 - [x] Pull out all actual processing into classes/generators and just let the script handle IO, etc.
 - [ ] Waffle: Move stuff into oxli?
 - [x] Shove things like min_size in init when we build the extractor


bocajnotnef:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?

bocajnotnef:
@ctb @camillescott @luizirber CR and thoughts on migrating the generators to oxli?

luizirber:
Check https://github.com/dib-lab/khmer/compare/rfact/extract-part...rfact/extract-part_sugg for a suggestion on how to define the generators and context managers. Defining the `__call__` of a class works, but it looks a bit weird to me...

scripts/extract-partitions.py
```diff
@@ -83,15 +84,159 @@ def get_parser():
     return parser
 
 
-# pylint: disable=too-many-statements
-def main():  # pylint: disable=too-many-locals,too-many-branches
+class PartitonedReadIterator(object):
```
ctb: spelling, please ;)

"Partitioned"
"Generator"

tests/test_scripts.py
```diff
@@ -1349,6 +1349,34 @@ def test_extract_partitions():
     assert len(parts) == 1, len(parts)
 
 
+def test_extract_paired_inconsitent_formats():
```
ctb: "inconsistent"

bocajnotnef:
@luizirber The tricky thing is in some cases I need to do the setup in __init__, which I don't think is possible when using the decorator.

Also, cleaned up, tested, ready for merge (unless we want to migrate stuff to oxli)

luizirber:
@bocajnotnef you can do setup before the yield and teardown after the yield, what kind of things do you need to save for later?

bocajnotnef:
We pull statistics from the ReadGenerator, here; https://github.com/dib-lab/khmer/pull/1251/files#diff-2b097bf3eed5cf5518cea07f5c4fb00eR316

luizirber:
I see. But can you use `__iter__` instead of `__call__`? Similar to what we do in [screed][1], for example.

[1]: https://github.com/dib-lab/screed/blob/ddcbcbd7618efb3a90064fae4531140387e62b52/screed/openscreed.py#L115

bocajnotnef:
I don't feel like that'd clean it up any, and as it is the functionality seems clear. (Plus, we did it this way in normalize-by-median_

ctb:
What's the reason to migrate stuff to oxli?

bocajnotnef:
Just seemed logical since what's essentially happening is the script is
being broken down into functions that do the processing and a main function
that makes it run as a script.

standage:
test this please jenkins

standage:
LGTM.

standage:
Thanks @bocajnotnef!

Issue #1256, opened 2015-08-15
load-into-countgraph rename messes up multiple versions.


ctb:
If you have an older version of khmer, and you install 2.0-pre, load-into-counting runs the old version and load-into-countgraph runs the new version.

More generally, we need to discuss script *renaming* in a top level issue.  All of our documentation and examples - much of it not linked into khmer directly - depends on these script names.

I'm -1 on changing script names at this point in the release cycle.

If there are any other such renames, please mention them here.

ctb:
The underlying logic behind the -1 is user focused, BTW - changing command line options is frustrating for users but necessary sometimes (and a major version # definitely is a good reason).  However, if we change the script name users don't know what script to look for, and they may accidentally run older versions that still import the newer package code.  There are probably other undesirable outcomes.  So I think it's a bigger decision than "hey, I'd like to rename this script to be consistent with our internal nomenclature."

ctb:
Same with load-into-nodegraph rename. Commits 2d621f6 and fa75f9cb342.

ctb:
Here are the commands to reproduce the bug that alerted me to this; run these on a blank Ubuntu 14.04
```
sudo apt-get update && \
sudo apt-get -y install screen git curl gcc make g++ python-dev unzip \
           default-jre pkg-config libncurses5-dev r-base-core \
           r-cran-gplots python-matplotlib sysstat python-virtualenv \
           python-setuptools cmake

cd
python -m virtualenv env
source env/bin/activate
pip install -U setuptools
pip install khmer==1.4.1

# at some later point, decide to update:
git clone https://github.com/dib-lab/khmer.git
cd khmer
make install

cd /mnt
load-into-counting.py ...
```

This last command fails with:
```
Traceback (most recent call last):
  File "/home/ubuntu/env/bin/load-into-counting.py", line 26, in <module>
    from khmer.kfile import check_space_for_hashtable
ImportError: cannot import name check_space_for_hashtable
```

This seems broken to me - I didn't realize that pip didn't lock things to version (i.e. I would have expected load-into-counting.py to load the khmer library from 1.4.1).  But it also adds a reason to not rename scripts, I think.

An alternative to renaming scripts is to produce scripts with multiple names, e.g. load-into-counting and load-into-countgraph, OR have a replacement 'load-into-counting' that says "use load-into-countgraph" - 

I'm -0 on changing the name at all, I'm -1 on eliminating load-into-counting.

ctb:
[...]

Issue #1181, opened 2015-07-19
Document broken-paired-reads behavior for 2.0


ctb:
See #733 for list of relevant scripts.

bocajnotnef:
Would this just go in `whats-new-2.0`?

ctb:
Yep. But leave it for now; changes yet to come... (split-paired-reads #1164 changes)

mr-c:
@ctb  ping

PR #1269, opened 2015-08-19
updated whats-new-2.0.rst

mr-c:
From the conversation at #1256

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,4 +1,4 @@
-2015-08-19  Michael R. Crusoe  <crusoe@ucdavis.edu>
+2015-08-25  Michael R. Crusoe  <crusoe@ucdavis.edu>
 
    * doc/whats-new-2.0.rst: Many updates from `diff`ing the `--help` output of
    version 1.4.1 vs now.
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-08-18  Michael R. Crusoe  <crusoe@ucdavis.edu>
+
+   * doc/whats-new-2.0.rst: Many updates from `diff`ing the `--help` output of
+   version 1.4.1 vs now.
+
 2015-08-27  Titus Brown  <titus@idyll.org>
 
   * doc/dev/getting-started.rst: fixed a few misspellings.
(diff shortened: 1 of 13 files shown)


mr-c:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?
  tested for streaming IO?)

mr-c:
@ctb This basically is the heart of the upcoming release notes. You can review it now or @camillescott can do so and I can reserve time for you to review before I release 2.0.

ctb:
I'll try to get to it today or over the weekend.  I've already looked over
most of the differences so I think it's basically good :)

ctb:
Thoughts on review --

--force-single on diginorm conflicts with --force_single on sample-reads-randomly. See @standage comment on this: https://github.com/dib-lab/khmer/pull/933#issuecomment-94997181.  I prefer force-single, but regardless of what we decide we should be consistent with at least this one option.

Under "New behavior", the text that occurs under "Mixed-pair sequence file format support" should be moved here and then referred to below.  Or something.  Regardless, the significant switch to support broken-paired behavior across the codebase should be mentioned in New Behavior somehow.

Under new behavior, we probably want to mention streaming I/O, no?  All the '-' behavior that we added?

Also under new behavior, the -M and -U options should be mentioned.  In fact, it looks like most of the text from incompatible changes should be moved up under new behavior, and only the switch from --min-tablesize to --max-tablesize needs to be mentioned in incompatible changes.

"it's helper" => "its helper"

mr-c:
To match the existing script (since that is the current trend) I've picked `--force_single` over `--force-single` though my personal preference is for the later.

mr-c:
@ctb ping

ctb:
As in, ready for review?

--

mr-c:
@ctb Yes, it is ready for your review of my latest changes. Do you feel that #1181 is satisfied by this PR?

ctb:
I've made some changes; if they look good to you, we can merge. I think we probably will need more docs for #1181 so let's leave that open, but punt to > 2.0.

mr-c:
Thanks; merged

PR #1293, opened 2015-09-05
make version test more specific

mr-c:
The test for the `--version` option & citation output is dynamically generated so as to avoid hard coding the lists of scripts that we provide. The folder the scripts get installed to may have other non-khmer scripts in them; this pull request ensures that we only test khmer scripts for khmer functionality by looking for the word 'khmer' on the second line of the script. This works due to the recent copyright audit #1289.

Currently the tests fail after installation from PyPI with `pip`. This PR fixes that.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2015-09-05  Michael R. Crusoe  <crusoe@ucdavis.edu>
+
+   * tests/test_scripts.py: make the script version test more specific; double
+   check that script in question is from the 'khmer' project based upon the
+   second line of the file.
+
 2015-09-04  Michael R. Crusoe  <crusoe@ucdavis.edu>
 
    * tests/khmer_tst_utils.py: look in "EGG-INFO" for scripts before checking
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -3711,7 +3711,7 @@ def test_unique_kmers_multiple_inputs():
     assert 'Total estimated number of unique 20-mers: 4170' in err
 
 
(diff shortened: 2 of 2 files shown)


mr-c:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?
  tested for streaming IO?)

mr-c:
I tested this by making a v2.0-rc7 tag locally and then uploading that to testpypi directly and running the tests on an Ubuntu 14.04 box. After merging this PR the entire release procedure will be restarted with a new release candidate.

mr-c:
@camillescott @luizirber @ctb Please review & merge; thanks!

ctb:
I don't see anything in this that's critical for the 2.0 release, and
would prefer that 2.0 come out before any more changes.

On the code itself, I would suggest putting in a comment in the source code
explaining the purpose (which is to look at the copyright header, right?)

ctb:
After reading some earlier e-mail, it appears this fixes a bug with the pypi
testing and hence should be in 2.0.  But I don't understand what the bug is or
why this fixes it; could you elaborate?

mr-c:
@ctb I've updated the PR description to answer your questions: https://github.com/dib-lab/khmer/pull/1293#issue-105056087

mr-c:
And I've added a commit giving the test a better name + explanatory inline comments.

ctb:
K. Happy to have it merged; en route so may not get to it myself.

--

ctb:
LGTM.

PR #1301, opened 2015-09-11
Sequence Bloom Tree changes

camillescott:
General collection of changes to help with the SBT implementation.


camillescott:
See https://github.com/dib-lab/khmer/pull/1302 instead

PR #1306, opened 2015-09-14
Deal with empty input on unique-kmers.py

### LINKED ISSUE #1305: unique-kmers.py core dumps on empty input
```
cat < /dev/null | unique-kmers.py -k 20 -
```

yields

```
terminate called without an active exception
Aborted (core dumped)
```

luizirber:
Fixes #1305

## Checklist

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/lib/hllcounter.cc
+++ b/lib/hllcounter.cc
@@ -424,6 +424,8 @@ void HLLCounter::consume_fasta(
                 // Iterate through the reads and consume their k-mers.
                 try {
                     read = parser->get_next_read();
+                } catch (khmer::StreamReadError) {
+                    break;
                 } catch (read_parsers::NoMoreReadsAvailable) {
                     break;
                 }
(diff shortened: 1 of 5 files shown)


lib/hllcounter.cc
```diff
@@ -424,6 +424,8 @@ void HLLCounter::consume_fasta(
                 // Iterate through the reads and consume their k-mers.
                 try {
                     read = parser->get_next_read();
+                } catch (khmer::StreamReadError) {
+                    break;
```
luizirber: I'm just continuing on `StreamError`, but should it stop and report an error? @ctb

ctb:
I don't know!

lib/hllcounter.cc
mr-c: I would let this be thrown and wrap it into a Python exception in `_khmer.cc`

luizirber:
ready for review and merge @ctb @standage @betatim

lib/hllcounter.cc
standage: I agree: we probably don't want to be silent about errors in the input data.

PR #1313, opened 2015-09-28
new file 'test_read_handling.py'

ljcohen:
removed interleave-reads, split-paired-reads, and extract-paired-reads functions from 'test_scripts.py' and put them in new file


tests/test_scripts.py
```diff
@@ -88,7 +88,7 @@ def test_load_into_counting_autoargs_0():
     outfile = utils.get_temp_filename('table')
     infile = utils.get_test_data('test-abund-read-2.fa')
 
-    args = ['-U', '1e7', '--fp-rate', '0.08', outfile, infile]
+    args = ['-U', '1ine7', '--fp-rate', '0.08', outfile, infile]
```
luizirber: undo this change, that's why the tests are failing.

PR #1316, opened 2015-09-29
HLL refactor: C++11 simplifications

luizirber:
- Better constant initialization (and avoiding race conditions too...)
- Use range loops when possible.
- Avoid hard to read code (`accumulate`, I'm looking at you).

## Checklist
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


lib/hllcounter.cc
```diff
@@ -269,36 +187,33 @@ HLLCounter::HLLCounter(double error_rate, WordLength ksize)
...
 double HLLCounter::get_erate()
 {
-    return 1.04 / sqrt(this->m);
+    return 1.04 / sqrt(ncounters);
 }
 
 void HLLCounter::set_erate(double error_rate)
 {
-    if (count(this->M.begin(), this->M.end(), 0) != this->m) {
+    if (count(begin(counters), end(counters), 0) != ncounters) {
```
betatim: Do you want to move this to a method with a name like `_all_counters_zero()`? It took me a moment to work out what it was doing and the code is reused below.

lib/hllcounter.cc
```diff
@@ -307,68 +222,69 @@ void HLLCounter::set_erate(double error_rate)
...
     HashIntoType x = khmer::_hash_murmur(value);
-    uint64_t j = x & (this->m - 1);
-    this->M[j] = std::max(this->M[j], get_rho(x >> this->p, 64 - this->p));
+    uint64_t j = x & (ncounters - 1);
+    counters[j] = std::max(counters[j],
+                           get_rho(x >> ncounters_log2, 64 - ncounters_log2));
 }
 
-unsigned int HLLCounter::consume_string(const std::string &inp)
+unsigned int HLLCounter::consume_string(const std::string &s)
```
betatim: is this a function for ["consenting adults"](https://github.com/dib-lab/khmer/pull/1511#issuecomment-274820149)? As in it should not check your input for you? If not, can we use https://github.com/dib-lab/khmer/blob/master/lib/read_parsers.cc#L48-L64 to do the cleaning?

lib/hllcounter.cc
```diff
@@ -481,16 +397,16 @@ bool HLLCounter::check_and_normalize_read(std::string &read) const
...
-    if (read.length() < this->_ksize) {
+    if (read.length() < _ksize) {
         return false;
     }
 
-    for (unsigned int i = 0; i < read.length(); i++) {
-        read[i] &= 0xdf; // toupper - knock out the "lowercase bit"
-        if (read[i] == 'N') {
-            read[i] = 'A';
+    for (auto &ch: read) {
```
betatim: as above, should we be checking here and if yes can we use one of the centralised functions to do it?

lib/hllcounter.hh
```diff
@@ -84,32 +84,32 @@ public:
...
 private:
     double _Ep();
     double alpha;
-    int p;
-    int m;
+    int ncounters;
+    int ncounters_log2;
     WordLength _ksize;
-    std::vector<int> M;
+    std::vector<uint8_t> counters;
```
betatim: For my education: being able to count to 255 is enough or why the change?

lib/hllcounter.cc
luizirber: Yup, I started poking HLL because I wanted to change the sequence parsing to use the proper cleaning from `read_parsers.cc`, will fix!

betatim: Nice. So this means `consume_string` like functions should check and clean the string given to them. The downside is that users who use khmer's sequence loading functionality will "clean" the sequence twice (in read parser and consume_string)

ctb: See #1590 and https://github.com/dib-lab/khmer/issues/1541#issuecomment-274828660 - I think we should only check input on bulk loading functions, not `consume_string`. So the toupper functionality should be removed for consistency.

betatim: Also matches the test here https://github.com/dib-lab/khmer/pull/1511/files/32fcf25137937de2dc52d214724f87312434c51a#diff-8e86ee1950f0fb1c11b9da2d3a59a088R193 :)

lib/hllcounter.hh
luizirber: `counters[i]` keep track of how many leading zeroes occurred in the hash, for the index `i`.
```
00010...01010000101010
\---------------/\---/
                  index (5 bits, ncounters = 2**5 = 32): 10
\-/ 3 leading zeroes

counters[10] = max(counters[10], 3)
```
So, the value of the counters is at most the length of the hash (64 bits in our case) - `log_2 (ncounters)` (5 in this case). Using something other than `uint8_t` will only be a problem if we use a hash value > 256 bits.

lib/hllcounter.cc
luizirber: sure, good idea!

lib/hllcounter.cc
luizirber: Added cleaning to `consume_fasta`, removed upper casing from `consume_string`.

luizirber:
Ready for review and merge @betatim @ctb @standage (as soon as the macOS builder works...)

luizirber:
(I'll leave the MurmurKmerIterator tests for another PR)

lib/hllcounter.cc
```diff
@@ -269,36 +181,37 @@ HLLCounter::HLLCounter(double error_rate, WordLength ksize)
...
-    this->M = M;
-
-    init_raw_estimate_data();
-    init_bias_data();
+    alpha = calc_alpha(nc);
+    ncounters_log2 = floor(log2(nc));
+    ncounters = 1 << ncounters_log2;
+    _ksize = ksize;
+    std::vector<uint8_t> newc(ncounters, 0);
+    counters = newc;
```
betatim: Why this indirection of first making a new vector and then assigning it? The goal is to have a vector of the correct size filled only with zeros?

lib/hllcounter.cc
```diff
@@ -433,16 +376,15 @@ void HLLCounter::consume_fasta(
...
                 #pragma omp task default(none) firstprivate(read) \
-                shared(counters, n_consumed_partial, total_reads_partial)
+                shared(hlls, n_consumed_partial, total_reads_partial)
                 {
-                    bool is_valid;
                     int n, t = omp_get_thread_num();
-                    n = counters[t]->check_and_process_read(read.sequence,
-                                                            is_valid);
+
+                    read.set_clean_seq();
```
betatim: I still don't like that we have to remember to call this. Can we move this to `FastxReader::get_next_read`? Would mean it is always done for everyone, pro: can't forget it, con: extra memory used for everyone.

betatim (approved):
LGTM (removing more lines than you are adding!)

betatim:
Depending on what you think about my two comments, go ahead and merge.

lib/hllcounter.cc
betatim: Maybe something for a new issue than this PR.

ctb:
+1

standage:
This didn't make it in before the 2.1 moratorium. Maybe a good time to revisit since you're all jazzed up on HLLs again? :)

PR #1321, opened 2015-10-25
fix and test for close of --output file after first input file

### LINKED ISSUE #1320: normalize-by-median fails with -o/--output when multiple input files are given
In the main loop of normalize-by-median.py, the outfile is closed after each input file - this causes problems when there are multiple input files.  See code:

https://github.com/dib-lab/khmer/blob/53484edfaa5ff0f7c088e084b00f528d781d4c10/scripts/normalize-by-median.py#L403

To reproduce:

    scripts/normalize-by-median.py data/100k-filtered.fa  data/100k-surrendered.fa -o xxx

which yields:

    ** ERROR: I/O operation on closed file
    ** Failed on data/100k-surrendered.fa: 
    ** Exiting!

While the output from the first file is correct, this then prevents the second file from being normalized.

ctb:
Fixes #1320.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2015-10-25  Titus Brown  <titus@idyll.org>
+
+   * scripts/normalize-by-median.py,tests/test_normalize_by_median.py: test
+   and fix for close of --output file after first input file.
+
 2015-10-05  Michael R. Crusoe  <crusoe@ucdavis.edu>
 
    * lib/magic: add file extensions, one media type, and a better comment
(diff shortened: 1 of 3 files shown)


ctb:
- [ ] Is it mergeable?
- [ ] Did it pass the tests?
  fixing as needed.
  changes were made?
  tested for streaming IO?)

Ready for review & merge, @luizirber @camillescott.  This is an easy one I think :)

luizirber:
LGTM

PR #1357, opened 2016-05-05
Fix: substr parameters

luizirber:
[substr()][1] second parameter should be the length of the substring, not the index of the last character to copy. There is another occurence at [lib/hashtable.cc][2] that needs to be checked.

[1]: http://www.cplusplus.com/reference/string/string/substr/
[2]: https://github.com/dib-lab/khmer/blob/ff7f047b5b0d9acb6c1eb73d54cfd39c9e3d1393/lib/hashtable.cc#L1265

- [x] Is it mergeable?
- [x] Did it pass the tests?
  fixing as needed.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2016-05-16  Luiz Irber  <khmer@luizirber.org>
+
+   * lib/hashtable.cc: fix bug with substr
+   * tests/test_countgraph.py: add test to trigger the substr bug.
+
 2016-05-11  Titus Brown  <titus@idyll.org>
 
    * scripts/trim-low-abund.py: refactor to use generators; add --diginorm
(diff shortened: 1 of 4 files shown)


ctb:
ping

luizirber:
@ctb can you take a look at [lib/hashtable.cc][1] to see if anything need to be changed there? (I'm having trouble following the logic =P)

[1]: https://github.com/dib-lab/khmer/blob/ff7f047b5b0d9acb6c1eb73d54cfd39c9e3d1393/lib/hashtable.cc#L1265

ctb:
It looks correct to me (tho quite obtuse code, sorry). The tests for the code in hashtable.cc seem to test it as well. So, :+1: I think.

ctb:
bump

luizirber:
@ctb ready for review and merge

ctb:
Thanks!

Issue #1360, opened 2016-05-10
Support paired-end reads to slice-reads-by-coverage


reedacartwright:
`slice-reads-by-coverage` does not support paired-end reads.

The program should allow two input files and two output files to be specified and only output a pair if both reads have good coverage.  If one member of a pair has good coverage, it can be optionally sent to a third output file.

PR #1366, opened 2016-05-11
Add script similar to slice-reads-by-coverage, but support paired reads

adamjorr:
- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

Includes basic tests.

Also see #1360


luizirber:
ok to test, jenkins

sandbox/slice-paired-reads-by-coverage.py
```diff
@@ -0,0 +1,134 @@
...
+                keep = False
+            if pmed > args.min_coverage:
+                pkeep = False
+
+        if keep and pkeep:
+            n_kept += 2
+            output_fp.write(output_single(record))
+            output_pfp.write(output_single(pair_record))
+
+        if args.output_singlefile:
```
reedacartwright: This could use an "else if" here.

sandbox/slice-paired-reads-by-coverage.py
```diff
@@ -0,0 +1,134 @@
...
+            if keep and not pkeep:
```
reedacartwright: if keep

sandbox/slice-paired-reads-by-coverage.py
```diff
@@ -0,0 +1,134 @@
...
+                n_kept += 1
+                output_sfp.write(output_single(record))
+            if pkeep and not keep:
+                n_kept += 1
```
reedacartwright: else if pkeep

PR #1370, opened 2016-05-12
Add a new CPython type, HashSet, to manage collections of hashes and tags.

ctb:
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
Ready for an initial review, ping @luizirber @camillescott.

ctb:
Jenkins, please test this.

ctb:
test this please

ctb:
OK, I think this is done. Reviews welcome...

standage:
Looks like we need one of these. Definitely helpful for newcomers like me to wrap my mind around the CPython stuff.

  entries
     the `ob_size` field).
     `tp_str` & `tp_repr`
  in `init_khmer()`

standage:
LGTM, although we might want @luizirber or @camillescott to take a look before merging.

ctb:
thanks & agree!

ctb:
@standage all right, it's been a week... can you push the magic green button? I'm sure we'll find any problems sooner or later anyway :)

PR #1351, opened 2016-04-18
Added MinHash C++ & CPython interface into khmer

ctb:
Closing this PR - the major remaining functionality is either not useful at the moment (the MHI stuff), is in other branches (the assembly stuff), or has been shifted to another project (MinHash -> sourmash).

Search bait: this functionality was developed for the sourmash/spacegraphcats side projects.

Ref https://github.com/dib-lab/sourmash/tree/khmer.

Includes #1370 and #1371.

- [ ] Is it mergeable?
- [ ] Did it pass the tests?
- [ ] If it introduces new functionality in scripts/ is it tested?
  Check for code coverage with `make clean diff-cover`
- [ ] Is it well formatted? Look at `make pep8`, `make diff_pylint_report`,
  `make cppcheck`, and `make doc` output. Use `make format` and manual
  fixing as needed.
- [ ] Did it change the command-line interface? Only additions are allowed
  without a major version increment. Changing file formats also requires a
  major version number increment.
- [ ] Is it documented in the ChangeLog?
  http://en.wikipedia.org/wiki/Changelog#Format
- [ ] Was a spellchecker run on the source code and documentation after
  changes were made?
- [ ] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)
- [ ] Is the Copyright year up to date?

Other TODO items:

-  [ ] test/remove 'prime'

PR #1371, opened 2016-05-14
Adds support for compact De Bruijn graph extraction

ctb:
Adds hashtable/graph functions "find_high_degree_nodes" and "traverse_linear_path", along with script "sandbox/extract-compact-dbg.py".

Extracted from MinHash branch (#1351).

Needed before review:

- [x] some tests for the extract-compact-dbg.py script.

----

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
Note to @luizirber - I copied the sandbox/extract-compact-dbg.py script over from walk-dbg.py in https://github.com/spacegraphcats/spacegraphcats/, and removed the MinHash dependencies. It should be a good place to experiment with compact DBG.  I also refactored some of the hashtable.cc high_degree_node stuff so that it no longer uses an internal C++ SeenSet.

khmer/_khmer.cc
```diff
@@ -1614,8 +1681,8 @@ hashtable_find_all_tags_list(khmer_KHashtable_Object * me, PyObject * args)
 
     Py_END_ALLOW_THREADS
 
-        PyObject * x = (PyObject *) create_HashSet_Object(tags,
-                                                          hashtable->ksize());
+    PyObject * x = (PyObject *) create_HashSet_Object(tags,
```
ctb: (this is a minor formatting fix that we should leave in)

lib/kmer_hash.cc
```diff
@@ -210,7 +211,7 @@ KmerIterator::KmerIterator(const char * seq,
     _nbits_sub_1 = (_ksize*2 - 2);
 
     index = _ksize - 1;
-    length = strlen(seq);
+    length = strlen(_seq);
```
ctb: use the internal _seq rather than the passed-in one. probably no big deal.

ctb:
Ready for review, @standage!

lib/hashtable.cc
```diff
@@ -997,4 +997,79 @@ void Hashtable::get_kmer_counts(const std::string &s,
...
+                                             SeenSet &adjacencies,
+                                             SeenSet &visited, Hashtable &bf,
+                                             SeenSet &high_degree_nodes)
+    const
+{
+    unsigned int size = 0;
+
+    auto filter = [&] (Kmer& n) -> bool {
+        return true;
+    };
```
standage: With my primarily ANSI C background I'm not sure how to parse this. Pointers? (no pun intended)

ctb: This is a lambda function that (in this case) always returns True.

However, I removed the need for it in the traverse(...) function so it
shouldn't be here anyway :).  (See traversal.cc, traverse_right(...).)

sandbox/extract-compact-dbg.py
```diff
@@ -0,0 +1,231 @@
...
+from __future__ import print_function
+import khmer
+import screed
+import argparse
+from collections import OrderedDict
+import sys
+
+# graph settings
+DEFAULT_KSIZE=31
+NODEGRAPH_SIZE=8e8 # small, big is 2e8
```
standage: I'm not sure how 8e8 corresponds to "small" while 2e8 corresponds to "big".

sandbox/extract-compact-dbg.py
```diff
@@ -0,0 +1,231 @@
...
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('seqfiles', nargs='+')
+    parser.add_argument('-o', '--output', default=None)
+    parser.add_argument('-k', '--ksize', default=DEFAULT_KSIZE, type=int)
+    parser.add_argument('-x', '--tablesize', default=NODEGRAPH_SIZE,
+                            type=float)
+    parser.add_argument('--force', action='store_true')
+    #parser.add_argument('--gml', action='store_true')
```
standage: GML is default, no? Is this option necessary?

standage:
LGTM, aside from the posted comments.

sandbox/extract-compact-dbg.py
ctb: Not any more, no. And it was commented out but should still be removed ;)

ctb:
addressed!

ctb:
w00t thanks!

standage:
👍

PR #1376, opened 2016-06-09
remove label pointer indirection from labelhash

ctb:
The LabelHash implementation has an extra level of indirection where pointers to labels are stored in addition to the labels themselves, so that labels can be reassigned en masse; this was inspired by/copied from the partitioning code, but is unnecessary for labels IMO. More to the point, it significantly increases memory demands if you have lots of labels. This PR removes it.

cc @camillescott 

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
Ready for review & merge @standage et al.

standage:
LGTM, merging.

PR #1388, opened 2016-06-21
Address cppcheck violations (mostly false positives)

standage:
- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,13 +64,15 @@ INCLUDEOPTS=$(shell gcc -E -x c++ - -v < /dev/null 2>&1 >/dev/null \
 	    | grep '^ /' | grep -v cc1plus | awk '{print "-I" $$1 " "}')
 PYINCLUDE=$(shell python -c \
 	  "import sysconfig;print(sysconfig.get_path('include'))")
+SYSPYINCLUDE=$(shell python -c \
+      "from __future__ import print_function; from distutils.sysconfig import get_python_inc; print(get_python_inc())")
 
 CPPCHECK=ls lib/*.cc khmer/_khmer.cc | grep -v test | cppcheck -DNDEBUG \
 	 -DVERSION=0.0.cppcheck -DSEQAN_HAS_BZIP2=1 -DSEQAN_HAS_ZLIB=1 \
-	 -UNO_UNIQUE_RC --enable=all --suppress='*:/usr/*' \
-	 --suppress='*:$(PYINCLUDE)/*' --file-list=- --platform=unix64 \
-	 --std=c++11 --inline-suppr --quiet -Ilib -Ithird-party/bzip2 \
-	 -Ithird-party/zlib -Ithird-party/smhasher -I$(PYINCLUDE) \
+	 -UNO_UNIQUE_RC --enable=all --suppress='*:/usr/*' --file-list=- \
+	 --platform=unix64 --std=c++11 --inline-suppr --quiet -Ilib \
+	 -Ithird-party/bzip2 -Ithird-party/zlib -Ithird-party/smhasher \
+	 -Ithird-party/seqan/core/include -I$(PYINCLUDE) -I$(SYSPYINCLUDE) \
 	 $(DEFINES) $(INCLUDEOPTS)
 
 UNAME := $(shell uname)
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -401,6 +401,7 @@ _ReadParser_iternext( PyObject * self )
     khmer_ReadParser_Object * myself  = (khmer_ReadParser_Object *)self;
     IParser *       parser  = myself->parser;
 
+    // cppcheck-suppress unreadVariable
     bool        stop_iteration  = false;
     const char *value_exception = NULL;
     const char *file_exception  = NULL;
@@ -459,6 +460,7 @@ _ReadPairIterator_iternext(khmer_ReadPairIterator_Object * myself)
     uint8_t     pair_mode = myself->pair_mode;
 
     ReadPair    the_read_pair;
+    // cppcheck-suppress unreadVariable
     bool        stop_iteration  = false;
     const char *value_exception = NULL;
     const char *file_exception  = NULL;
@@ -1019,7 +1021,7 @@ hashset_update(khmer_HashSet_Object * me, PyObject * args)
             return NULL;
         }
         me->hashes->insert(h);
-        
+
         Py_DECREF(item);
         item = PyIter_Next(iterator);
     }
@@ -1450,7 +1452,7 @@ hashtable_neighbors(khmer_KHashtable_Object * me, PyObject * args)
     }
 
     unsigned int i;
-    for (i = 0; node_q.size() > 0; i++) {
+    for (i = 0; !node_q.empty(); i++) {
         HashIntoType h = node_q.front();
         node_q.pop();
         // type K for python unsigned long long
@@ -1862,6 +1864,7 @@ hashtable_trim_on_stoptags(khmer_KHashtable_Object * me, PyObject * args)
         return NULL;
     }
 
+    // cppcheck-suppress unusedVariable
     size_t trim_at;
     Py_BEGIN_ALLOW_THREADS
 
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -401,7 +401,6 @@ _ReadParser_iternext( PyObject * self )
     khmer_ReadParser_Object * myself  = (khmer_ReadParser_Object *)self;
     IParser *       parser  = myself->parser;
 
-    // cppcheck-suppress unreadVariable
     bool        stop_iteration  = false;
     const char *value_exception = NULL;
     const char *file_exception  = NULL;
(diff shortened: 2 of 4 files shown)


standage:
Whoa! I'm using atom as my text-editor, and it's auto-trimming a lot of white space.

standage:
The `cppcheck` Makefile target relies on Bash functionality. Can we explicitly set `SHELL=bash` in the Makefile, or are there portability concerns or other targets that preclude this?

khmer/_khmer.cc
```diff
@@ -401,6 +401,7 @@ _ReadParser_iternext( PyObject * self )
     khmer_ReadParser_Object * myself  = (khmer_ReadParser_Object *)self;
     IParser *       parser  = myself->parser;
 
+    // cppcheck-suppress unreadVariable
```
standage: If we replace this with a `bool stop_iteration;` declaration, we get an unused variable warning.

khmer/_khmer.cc
```diff
@@ -1862,6 +1864,7 @@ hashtable_trim_on_stoptags(khmer_KHashtable_Object * me, PyObject * args)
         return NULL;
     }
 
+    // cppcheck-suppress unusedVariable
```
standage: No idea why this is being flagged as an unused variable.

standage:
Ah, yes...

```
# `SHELL=bash` Will break Titus's laptop, so don't use BASH-isms like
# `[[` conditional expressions.
```

It looks like `printf "%q"` isn't portable shell code, so we need another way to escape the DEFINES string.

standage:
Fixing the Makefile target so that all headers are found properly fixed the unset/unread variable warnings. Now it's reporting a syntax error in the seqan code, which is very surprising since cppcheck explicitly says it does not look for syntax errors. Any attempt to suppress `syntaxError` is unsuccessful. :(

Makefile
```diff
@@ -64,13 +64,15 @@ INCLUDEOPTS=$(shell gcc -E -x c++ - -v < /dev/null 2>&1 >/dev/null \
 	    | grep '^ /' | grep -v cc1plus | awk '{print "-I" $$1 " "}')
 PYINCLUDE=$(shell python -c \
 	  "import sysconfig;print(sysconfig.get_path('include'))")
```
betatim: Needs a `from __future__ import print_function;`

Makefile
```diff
@@ -64,13 +64,15 @@ INCLUDEOPTS=$(shell gcc -E -x c++ - -v < /dev/null 2>&1 >/dev/null \
...
+SYSPYINCLUDE=$(shell python -c \
+      "from __future__ import print_function; from distutils.sysconfig import get_python_inc; print(get_python_inc())")
```
betatim: Do you have a link that explains when `SYSPYINCLUDE` and `PYINCLUDE` would be different? On my system they are the same and it seems odd that there should be two places to include things from. "One obvious way" and all that :-/

betatim:
[OXS cppcheck output](http://ci.oxli.org/job/khmer-pullrequest/3558/label=osx/cppcheckResult/source.all/?before=5&after=5) has some "Unmatched suppression: passedByValue", is this cppcheck telling us that it doesn't know this suppression? If yes, why does this depend on the platform? 😕 

No good idea on the syntax error.

ctb:
On Mon, Jun 20, 2016 at 08:57:48PM -0700, Daniel Standage wrote:
> Ah, yes...
> 
> ```
> # `SHELL=bash` Will break Titus's laptop, so don't use BASH-isms like
> # `[[` conditional expressions.
> ```

I've since updated my laptop so maybe this is no longer a problem.  You
have the same laptop as me now, actually, so... if it works for you it'll
work for me.

Makefile
standage: If you're using virtualenv, the first one will point to the virtualenv's include directory, while the second will point to the system's include directory (where the Python dev headers are).

```
(kenv)[standage@lappy khmer]$ python -c "from __future__ import print_function; import sysconfig; print(sysconfig.get_path('include'))"
/home/standage/Software/kenv/include/python3.4m
(kenv)[standage@lappy khmer]$ python -c "from __future__ import print_function; from distutils.sysconfig import get_python_inc; print(get_python_inc())"
/usr/include/python3.4m
```

Even when I am not in a virtualenv, these give different results.

```
[standage@lappy khmer]$ python -c "from __future__ import print_function; import sysconfig; print(sysconfig.get_path('include'))"
/usr/local/include/python2.7
[standage@lappy khmer]$ python -c "from __future__ import print_function; from distutils.sysconfig import get_python_inc; print(get_python_inc())"
/usr/include/python2.7
```

That first directory doesn't even exist on my system (Debian Jessie). I'll try to confirm whether we even need the old PYINCLUDE. It may be sufficient to overwrite the value instead of creating a new SYSPYINCLUDE value.

Makefile
luizirber: On 06/20, Tim Head wrote:
>> @@ -64,13 +64,15 @@ INCLUDEOPTS=$(shell gcc -E -x c++ - -v < /dev/null 2>&1 >/dev/null \
>>  	    | grep '^ /' | grep -v cc1plus | awk '{print "-I" $$1 " "}')
>>  PYINCLUDE=$(shell python -c \
>>  	  "import sysconfig;print(sysconfig.get_path('include'))")
>
>Needs a `from __future__ import print_function;`

In this case we can abuse the print behavior in 2.7, printing a tuple unpacks
it first (so this works the same on 2.7 or 3)

Makefile
betatim: Python packaging hell 😢 

In a `conda` env both methods give the same path: the one that belongs to the `python` from the conda env. Annoying that neither of these methods gives the correct answer in both cases (inside and outside of a venv).

standage:
@timhead The unmatched suppression message means that line we flagged didn't invoke the error we intended to suppress. So the error is invoked on OSX (cppcheck v1.74) but not on Debian (cppcheck v1.6x).

I `apt-get remove`d the system cppcheck on Debian and installed version 1.74 from source to see if I could get the same behavior. This time, it fails much earlier with the following message.

```
$ make cppcheck
[third-party/seqan/core/include/seqan/basic/fundamental_concepts.h:914]: (error) Internal Error: Division by zero
```

It's frustrating ow cppcheck's behavior is changing so much between versions and platforms.

standage:
Also on OSX, warnings about the command line tools.

```
[/Library/Developer/CommandLineTools/usr/include/c++/v1/type_traits:1202]: (error) Syntax Error: AST broken, ternary operator lacks ':'.
[/Library/Developer/CommandLineTools/usr/include/c++/v1/type_traits:2463] -> [/Library/Developer/CommandLineTools/usr/include/c++/v1/type_traits:255]: (warning) The struct '__contains_void' defines member variable with name 'value' also defined in its parent struct 'integral_constant'.
[/Library/Developer/CommandLineTools/usr/include/c++/v1/type_traits:665]: (style) Struct '__any' has a constructor with 1 argument that is not explicit.
[/Library/Developer/CommandLineTools/usr/include/c++/v1/type_traits:837]: (style) Struct '_Dst' has a constructor with 1 argument that is not explicit.
```

And no warnings emitted from `/usr` as they are on Debian.

```
[/usr/*]: (information) Unmatched suppression: *
```

Makefile
betatim: You need at least one comma to make a tuple: `type((23))` -> `int` and `type((23,))` -> `tuple`. So it isn't tuple unpacking but the fact that `(23)` is a noop. Legacy python you so cray cray :)

This means we don't need the print function import in the other case either. I think.

(spot the person waiting for a plane and too much time ...)

standage:
I started this PR with hopes that making all the errors go away would be pretty straightforward. Alas, due to cppcheck's variable behavior across versions and platforms it is not. I see a few of options.

1. Add a note about possible false positives to the developer docs, refer to this in the PR checklist, and not worry too much about superfluous warnings for now.
2. Invest more time making cppcheck messages go away with platform-specific cppcheck Makefile targets
3. Invest more time making cppcheck messages go away with additional debugging, communication with cppcheck developers, in hopes of a cross-platform solution.

Comments? Suggestions?

ctb:
On Tue, Jun 21, 2016 at 12:04:59PM -0700, Daniel Standage wrote:
> I started this PR with hopes that making all the errors go away would be pretty straightforward. Alas, due to cppcheck's variable behavior across versions and platforms it is not. I see a few of options.
> 
> 1. Add a note about possible false positives to the developer docs, refer to this in the PR checklist, and not worry too much about superfluous warnings for now.
> 2. Invest more time making cppcheck messages go away with platform-specific cppcheck Makefile targets
> 3. Invest more time making cppcheck messages go away with additional debugging, communication with cppcheck developers, in hopes of a cross-platform solution.
> 
> Comments? Suggestions?

#1 sounds best to me.  Are there benefits to cppcheck that we are missing
currently?

standage:
> #1 sounds best to me.

That is my preference as well. This isn't something I had intended to spend much time on.

> Are there benefits to cppcheck that we are missing currently?

Not that I know of. My original motivation for this PR was to make all warnings disappear. I was unfamiliar with the tool and didn't anticipate the version/platform complications.

ctb:
On Wed, Jun 22, 2016 at 12:07:15PM -0700, Daniel Standage wrote:
> > #1 sounds best to me.
> 
> That is my preference as well. This isn't something I had intended to spend much time on.
> 
> > Are there benefits to cppcheck that we are missing currently?
> 
> Not that I know of. My original motivation for this PR was to make all warnings disappear. I was unfamiliar with the tool and didn't anticipate the version/platform complications.

figured as much :). let's close this with a wontfix tag.

Issue #1249, opened 2015-08-13
Determine system memory automatically and offer suggestions?


ctb:
I noticed that MEGAHIT (https://github.com/voutcn/megahit) has the following output --

```
 14.0Gb memory in total.
 Using: 13.219Gb.
```

I'm not sure we want to do this - it seems potentially problematic for HPC & queuing environments! - but would be nice to provide it as an option, maybe.

mr-c:
Now that `-M` has landed it could be a nice alternative to not running at all. There should be a warning printed, though.

standage:
@ctb I'm -0 on this one.

ctb:
why? too confusing?

standage:
Mostly for the HPC / shared environment issue you raised above. I think #1259 would be great in terms of use experience.

ctb:
Should be possible to do something sensible on default, like

* automatically grab 80% of available memory, up to 16 GB;
* ignore automatic behavior whenever -x, -N, -M etc specified.

Our memory parameter stuff is becoming a morass so ... simplification welcome :)

standage:
Embarrassingly I've never had to worry about this much, so: with memory, do we compute 16 GB as `16 x 1000 x 1000 x 1000` or `16 x 1024 x 1024 x 1024`?

ctb:
Decision in #1396 was wontfix.

PR #1396, opened 2016-06-29
Adds a new option to automagically calculate memory from available memory on system

standage:
See #1249. This adds a new dependency [psutil](https://github.com/giampaolo/psutil), which is available via pip and actively developed.

---
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -361,8 +361,7 @@ def add_loadgraph_args(parser):
                         help='load a precomputed k-mer graph from disk')
 
 
-def calculate_graphsize(args, graphtype, multiplier=1.0,
-                        sysmem=psutil.virtual_memory().available):
+def calculate_graphsize(args, graphtype, multiplier=1.0, _sysmem=None):
     """Transform the table parameters into a size."""
     if graphtype not in ('countgraph', 'nodegraph'):
         raise ValueError("unknown graph type: %s" % (graphtype,))
@@ -372,7 +371,9 @@ def calculate_graphsize(args, graphtype, multiplier=1.0,
         if graphtype == 'nodegraph':
             tablesize *= 8.0
     elif args.auto_mem:
-        avail = min(sysmem * 0.8, 16 * (1000 ** 3))
+        if _sysmem is None:
+            _sysmem = psutil.virtual_memory().available
+        avail = min(_sysmem * 0.8, 16 * (1000 ** 3))
         tablesize = avail / args.n_tables / float(multiplier)
         if graphtype == 'nodegraph':
             tablesize *= 8.0
@@ -382,8 +383,7 @@ def calculate_graphsize(args, graphtype, multiplier=1.0,
     return tablesize
 
 
-def create_nodegraph(args, ksize=None, multiplier=1.0, fp_rate=0.01,
-                     sysmem=psutil.virtual_memory().available):
+def create_nodegraph(args, ksize=None, multiplier=1.0, fp_rate=0.01):
     """Create and return a nodegraph."""
     args = _check_fp_rate(args, fp_rate)
     if ksize is None:
@@ -392,13 +392,11 @@ def create_nodegraph(args, ksize=None, multiplier=1.0, fp_rate=0.01,
         print_error("\n** ERROR: khmer only supports k-mer sizes <= 32.\n")
         sys.exit(1)
 
-    tablesize = calculate_graphsize(args, 'nodegraph', multiplier=multiplier,
-                                    sysmem=sysmem)
+    tablesize = calculate_graphsize(args, 'nodegraph', multiplier=multiplier)
     return khmer.Nodegraph(ksize, tablesize, args.n_tables)
 
 
-def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1,
-                      sysmem=psutil.virtual_memory().available):
+def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1):
     """Create and return a countgraph."""
     args = _check_fp_rate(args, fp_rate)
     if ksize is None:
@@ -407,8 +405,7 @@ def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1,
         print_error("\n** ERROR: khmer only supports k-mer sizes <= 32.\n")
         sys.exit(1)
 
-    tablesize = calculate_graphsize(args, 'countgraph', multiplier=multiplier,
-                                    sysmem=sysmem)
+    tablesize = calculate_graphsize(args, 'countgraph', multiplier=multiplier)
     return khmer.Countgraph(ksize, tablesize, args.n_tables)
 
 
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -327,11 +327,15 @@ def build_graph_args(descr=None, epilog=None, parser=None):
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--max-tablesize', '-x', type=float,
                        default=DEFAULT_MAX_TABLESIZE,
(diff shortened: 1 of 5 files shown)


khmer/khmer_args.py
```diff
@@ -356,25 +361,29 @@ def add_loadgraph_args(parser):
                         help='load a precomputed k-mer graph from disk')
 
 
-def calculate_graphsize(args, graphtype, multiplier=1.0):
+def calculate_graphsize(args, graphtype, multiplier=1.0,
+                        sysmem=psutil.virtual_memory().available):
```
ctb: I'd suggest leaving this as something default ('None'?) and then calculating it in body of the function if needed; then you can pass it in for the tests w/o making the default def that ugly.

Also, if it's an internal thing that will only be used by tests, suggest naming it _sysmem or something else.

standage: Yeah, that's a bit cleaner. And I can also avoid cluttering the `create_*graph` functions by testing `calculate_graphsize` directly. That'll also help avoid problems with Jenkins like when I, oh, tried to create a 64 GB countgraph for testing! :-/

standage:
Ready for review and merge @ctb et al.

ctb:
Hi @standage, thx -

A few questions/comments off the top:

* is psutil a binary dependency, or is it pure Python?
* I think your first comment is no longer true :)
* in #1249, we suggest having auto-mem be the default behavior; that seems OK to me, as long as we're explicit about it. ...but this might require a major version number. Hrm.

ctb:
(Note that grabbing a pile of memory is what most programs do. So I don't think we should feel guilty about it, since at least ours won't crash the system, probably.)

standage:
>  is psutil a binary dependency, or is it pure Python?

It appears psutil has some compiled C code.

> I think your first comment is no longer true :)

Correct. I thought I added a clarifying remark somewhere, but I don't see it on this thread.

> in #1249, we suggest having auto-mem be the default behavior; that seems OK to me, as long as we're explicit about it. ...but this might require a major version number. Hrm.

Yeah, making this the default would indeed be grounds for a major version bump. But what about adding the option with v2.1, and then waiting to make it the default until v3.0 release?

standage:
I've done a bit more searching, and most of the solutions I see for determining available system memory involve parsing `/proc/meminfo`, which is a Linux-specific solution. A pure Python package with this type of cross-platform OS awareness doesn't seem readily available, and (in my opinion) is beyond the scope of what we'd want to implement within this project.

ctb:
On Tue, Jul 05, 2016 at 09:16:17AM -0700, Daniel Standage wrote:
> >  is psutil a binary dependency, or is it pure Python?
> 
> It appears psutil has some compiled C code.
> 
> > I think your first comment is no longer true :)
> 
> Correct. I thought I added a clarifying remark somewhere, but I don't see it on this thread.

OK. I tend to edit the head comment matter for this kind of thing, YMMV.

> > in #1249, we suggest having auto-mem be the default behavior; that seems OK to me, as long as we're explicit about it. ...but this might require a major version number. Hrm.
> 
> Yeah, making this the default would indeed be grounds for a major version bump. But what about adding the option with v2.1, and then waiting to make it the default until v3.0 release?

:+1:

khmer/khmer_args.py
```diff
@@ -360,12 +361,15 @@ def build_graph_args(descr=None, epilog=None, parser=None):
     group = parser.add_mutually_exclusive_group()
     group.add_argument('--max-tablesize', '-x', type=float,
                        default=DEFAULT_MAX_TABLESIZE,
-                       help='upper bound on tablesize to use; overrides ' +
+                       help='upper bound on tablesize to use; overrides '
                        '--max-memory-usage/-M')
```
betatim: One thing you can do is add `(` `)` around the multi line string and indent the second line to line up with the one above. I find it helps with making it clear that these strings were meant to be joined together. Not sure what the style gods say regarding implicit/explicit joining of strings across lines.

khmer/khmer_args.py
```diff
@@ -360,12 +361,15 @@ def build_graph_args(descr=None, epilog=None, parser=None):
...
-    group.add_argument('-M', '--max-memory-usage', type=memory_setting,
-                       help='maximum amount of memory to use for data ' +
+    group.add_argument('-M', '--max-memory-usage', type=float,
+                       help='maximum amount of memory to use for data '
                        'structure')
```
betatim: same as above

khmer/khmer_args.py
```diff
@@ -360,12 +361,15 @@ def build_graph_args(descr=None, epilog=None, parser=None):
...
-
+    group.add_argument('-a', '--auto-mem', action='store_true',
+                       help='automagically allocate memory for data structure;'
+                       ' use 80%% of available memory or 16GB, whichever is '
+                       'smaller')
```
betatim: same as above

khmer/khmer_args.py
```diff
@@ -389,18 +393,22 @@ def add_loadgraph_args(parser):
                         help='load a precomputed k-mer graph from disk')
 
 
-def calculate_graphsize(args, graphtype, multiplier=1.0):
+def calculate_graphsize(args, graphtype, multiplier=1.0, _sysmem=None):
```
betatim: Why have the `_sysmem` argument? Is it solely to make the unittests work? In python you can use [mock](https://docs.python.org/3/library/unittest.mock.html) to replace the `psutil` call for the unit test. Unfortunately it isn't part of the stdlib for `python2`.

(Function arguments starting with a `_` attract attention in python land 👀

khmer/khmer_args.py
ctb: 'make pep8' checks against our chosen style standard; if it passes that,
we're happy :)

khmer/khmer_args.py
standage: Yes, it's solely for the unit tests.

By "attract attention", do you mean *bad* attention? I thought the leading underscore was a pretty common convention for saying "this is internal, you shouldn't mess with this unless you know what you're doing."

betatim: Attracts attention in the sense that I'd have to think about it before I could show you another example of it. `_` in python does indeed mean "don't mess" though usually apply to member variables or functions/methods.

TL;DR: There should be a better way of doing this (but I can't think of one right now short of introducing `mock` as a dependency for `python2`). Maybe an executive decision is needed re adding dependencies for developers.

ctb: If it causes practical problems, we'll eliminate it.  Hard to know at this
point. But memory settings are one of the major causes of user pain,
at least anecdotally.

betatim: The downside to having the argument there is that one day a user will try and use it. I don't think it will cause technical problems. The pythonista in me thinks it is ugly, in the end it is like all code smells...

ctb: Sure; the tradeoff is that the workaround would add technical complexity
to our testing framework :)

standage:
Reflecting on this several months later: we have minor syntax/implementation issues that IMO aren't blockers for merging. If we still feel strongly about them we can quickly fix, if not we can open separate issues to continue (not) discussing.

The big question here is whether we want to add psutil as a dependency. It's not pure Python, so it will complication installation, especially on legacy systems. The counter argument is that now we're requiring C++11 support anyway, and since this is a robust and actively developed library it should compile on the vast majority of "modern" systems.

What I DON'T think we should do is try to implement OS-dependent memory awareness ourselves. That'll be messy and buggy.

Honestly, now that khmer supports human-friendly memory requests (`-M 8G`), much of the user headache should be gone. I'm in favor of closing this PR for now, or punting.

ctb:
punt with #wontfix then. thx!

PR #1401, opened 2016-07-03
add get_kmer_hashes_as_hashset function

ctb:
the existing function get_kmer_hashes returns all the hashes of all the k-mers in a string, in order; get_kmer_hashes_as_hashset returns them in a HashSet instead, which is much faster for certain use cases.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
Ready for review!

standage:
LGTM!

PR #1377, opened 2016-06-09
Various tools for assembling directly from the graph, including streaming RNAseq assembly prototype.

ctb:
(Further work on this is occurring in #1412.)

See especially `sandbox/link-compact-dbg.py`, end of `tests/test_nodegraph.py`, and `sandbox/assemble-on-the-go.py`

You need the fix/reverse_complement latest branch of screed: `pip install git+https://github.com/dib-lab/screed.git@fix/reverse_complement`

Note: incorporates linear-path contig assembly (https://github.com/dib-lab/khmer/pull/1372).

- [ ] Is it mergeable?
- [ ] `make test` Did it pass the tests?
- [ ] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?
- [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
  formatted?
- [ ] Did it change the command-line interface? Only additions are allowed
  without a major version increment. Changing file formats also requires a
  major version number increment.
- [ ] Is it documented in the `ChangeLog`?
  http://en.wikipedia.org/wiki/Changelog#Format
- [ ] Was a spellchecker run on the source code and documentation after
  changes were made?
- [ ] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)
- [ ] Is the Copyright year up to date?

PR #1412, opened 2016-07-25
Various tools for assembling directly from the graph, including using labels to track paths.

camillescott:
(Forked from #1377, which is now closed.)

See especially sandbox/link-compact-dbg.py, end of tests/test_nodegraph.py, and sandbox/assemble-on-the-go.py

You need the fix/reverse_complement latest branch of screed: pip install git+https://github.com/dib-lab/screed.git@fix/reverse_complement

Note: incorporates linear-path contig assembly (#1372).
## Camille's train of thought:
- [x] Move assemble-related stuff into a new `Assembler` class.
  
  **Reasoning**: Fits in line with overarching goal of stripping stuff out of `Hashtable`, such as was done with the `Traverser` class. `Hashtable` is already large and unwieldy. 
- [x] Fix the segfaut in labelled assembly.
  
  **Reasoning**: It doesn't work! More precisely, the labelled assembly function appears to get stuck when confronted with cycles of a certain variety.
- [x] Subclass `Assembler` for `LabeledAssembler`
  
  **Reasoning**: See reasoning for `Assembler` class.
- [x] Add an `_assemble_left` function.
  
  **Reasoning**: Obvious extension of having an assemble right function.
- [x] Let `assemble_linear_path` take a `std::function` for filtering.
## Changes that Might be Made
- [x] Use `Traverser` for `Assembler`
  
  Unfortunately, after trying to implement this myself, it's clear why @ctb didn't use it for this in the first place -- it's very geared toward agnostically walking around integer k-mers, and difficult / inefficient to extend to string traversal. This should probably change in the future, but not before tip detection is in. **Update**: this is now working. 
## Obligatory checklist (necessary for a PR to a PR?)
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
I endorse these changes thus far :)

camillescott:
Jenkins, test this please

camillescott:
Jenkins, test this please

tests/test_nodegraph.py
```diff
@@ -1089,3 +1090,466 @@ def test_traverse_linear_path_3_stopgraph():
...
+    len_path = len(paths)
+
+    print('len path:', len_path)
+
+    found = False
+    for path in paths:
+        if utils._equals_rc(path, contig):
+            found = True
+            break
+    assert found
```
betatim: -> `assert any(utils._equals_rc(path, contig) for path in paths)`? and same below

lib/assembler.cc
```diff
@@ -0,0 +1,300 @@
...
+Contact: khmer-project@idyll.org
+*/
+
+#include "assembler.hh"
+
+#include <algorithm>
+#include <iostream>
+
+#define DEBUG 0
+#define DEBUG_AT 0
```
betatim: What is the normal thing to do around here regarding these kind of `#define`s? I'd remove them to let someone `-DDEBUG` with a compiler option if they want to.

lib/khmer.hh
```diff
@@ -154,6 +161,11 @@ class Kmer;
 typedef std::queue<Kmer> KmerQueue;
 typedef std::set<Kmer> KmerSet;
 
+// A function which takes a Kmer and returns true if it
+// is to be filtered / ignored
+typedef std::function<bool (Kmer&)> KmerFilter;
+typedef std::list<KmerFilter> KmerFilterList;
```
betatim: Why `std::list`? This should be a `vector` unless there is a good reason to use `list` which is nearly never a good idea ;)

If there is a reason for `list` it is worth leaving a comment here to inform archeologists from the future who will ask the same question.

lib/traversal.cc
```diff
@@ -141,7 +160,147 @@ unsigned int Traverser::degree_right(Kmer& node)
...
+template<bool direction>
+char AssemblerTraverser<direction>::next_symbol()
+{
+    char * symbol_ptr = alphabets::DNA_SIMPLE;
+    char base = *symbol_ptr; // kill -Wmaybe-uninitialized warning
+    short found = 0;
+    Kmer neighbor;
+    Kmer cursor_next;
+
+    while(*symbol_ptr != '\0') {
```
betatim: Why not `for (auto base : alphabets::DNA_SIMPLE) {...}`? For me `while () {}` is more obscure/harder to reason about.

lib/traversal.cc
```diff
@@ -141,7 +160,147 @@ unsigned int Traverser::degree_right(Kmer& node)
...
+        neighbor = get_neighbor(cursor, *symbol_ptr);
```
betatim: Can you add a comment saying that this is a potential neighbour? It took me a while to understand why you fetch a neighbour and then check if it is in the graph etc

ctb:
Ready for review! @betatim could you take this one on?

camillescott:
@betatim thanks for the comments -- for some reason I didn't get any sort of message from github saying that a review had been triggered.

For both @betatim and Titus -- going to make the suggested changes; also have one more fix to add in today that I think really needs to be in before merge. Will ping when finished.

betatim:
I'll take another look when you pushed more changes. Most of my comments so far go in the direction of "can we make the code easier to read without any loss of performance". I ❤️ easy to read code.

lib/traversal.cc
camillescott: done. had to change the alphabets to type `std::string` to make it work, but that's probably reasonable anyhow.

lib/subset.cc
```diff
@@ -236,12 +236,11 @@ void SubsetPartition::find_all_tags(
     unsigned int total = 0;
     unsigned int nfound = 0;
 
-    Traverser traverser(_ht);
     KmerSet keeper;		// keep track of traversed kmers
-
-    auto filter = [&] (Kmer& n) -> bool {
-        return !set_contains(keeper, n);
+    KmerFilter filter = [&] (const Kmer& n) -> bool {
+        return set_contains(keeper, n);
```
betatim: did you mean to remove the ! in front of `set_contains`?

lib/Makefile
```diff
@@ -1,4 +1,4 @@
-# This file is part of khmer, https://github.com/dib-lab/khmer/, and is
+alphabets# This file is part of khmer, https://github.com/dib-lab/khmer/, and is
```
camillescott: oh dear

lib/subset.cc
camillescott: Yeah -- turns out the behavior between the previous `Traverser` class and the new `AssemblerTraverser` with respect to `KmerFilter`s was inverted. For `Traverser`, a filter returned `true` to pass a filter, and for `AssemblerTraverser`, to fail. This was obviously somewhat confusing, and I decided that a k-mer _filter_ should universally return `true` if it aims to, well filter, a k-mer.

lib/traversal.cc
camillescott: Added a comment along with doxygen comments for all the classes. Somewhat annoying that the UI for this new review system doesn't seem to reflect new changes...

tests/test_nodegraph.py
```diff
@@ -1089,3 +1090,466 @@ def test_traverse_linear_path_3_stopgraph():
...
+    path = nodegraph.assemble_linear_path(contig[101:101 + K])
+    len_path = len(path)
+
+    print('len path:', len_path)
+
+    # this is actually bad behavior; we're ignoring a branch.
+    # FIXME at some point.
+    assert utils._equals_rc(path, contig)
+    # should? be:
+    #    assert utils._equals_rc(path, contig[101:])
```
camillescott: So, it seems that the FIXME is not necessary. I'll remove it.

lib/traversal.cc
betatim: Yes, that has been annoying ever since. In the old system at least comments that were on stuff that changed got folded away, so I could see if people had done stuff (assuming they'd only ever change it towards the suggestion if they didn't also comment) ... alas the world is not perfect :)

betatim:
Do you understand why travis is not doing its thing with this? Could you push something to see if that wakes it up? My guess is that this is somehow a remnant of the GH troubles??

betatim:
The various sequence generation and manipulation stuff at the top of `test_assembly.py`, could that move to `khmer_tst_utils` or some other shared place? Seems like that might be useful for other tests.

khmer/_khmer.cc
```diff
@@ -1584,8 +1585,11 @@ hashtable_assemble_linear_path(khmer_KHashtable_Object * me, PyObject * args)
     if (nodegraph_o) {
         stop_bf = nodegraph_o->hashbits;
     }
+    LinearAssembler assembler(hashtable);
+    KmerFilterList lst;
+    AssemblerTraverser<RIGHT> test(hashtable, start_kmer, lst);
```
betatim: Is this left over from a previous incarnation, otherwise I think we need a comment to explain what it does.

khmer/_khmer.cc
```diff
@@ -4329,6 +4332,63 @@ labelhash_n_labels(khmer_KGraphLabels_Object * me, PyObject * args)
...
 static
 PyObject *
+labelhash_label_across_high_degree_nodes(khmer_KGraphLabels_Object * me,
+        PyObject * args)
+{
+    LabelHash * labelhash = me->labelhash;
+
+    const char * long_str;
+    khmer_HashSet_Object * hdn_o = NULL;
+    Label label;
```
betatim: What is a label? Type wise it is an `uint64` which makes me think that in the future it also needs to be a `BigHashType`. But I don't actually really know what the concept of a label is (from a science POV).

ctb:
On Wed, Oct 05, 2016 at 08:05:18AM -0700, Tim Head wrote:
> The various sequence generation and manipulation stuff at the top of `test_assembly.py`, could that move to `khmer_tst_utils` or some other shared place? Seems like that might be useful for other tests.

When we find tests that could use it, I think we can move it then...

tests/test_nodegraph.py
```diff
@@ -1089,3 +1090,466 @@ def test_traverse_linear_path_3_stopgraph():
...
+
+    K = 21
+
+    nodegraph = khmer.Nodegraph(K, 1e5, 4)
+
+    nodegraph.consume(contig)
+    nodegraph.count(contig[101:121] + 'G')  # will add another neighbor
+
+    path = nodegraph.assemble_linear_path(contig[0:K])
+    len_path = len(path)
```
camillescott: Done.

khmer/_khmer.cc
camillescott: A Label is just an annotation for a node. For now it's just a uint64 for simplicity, though it could be defined as a base class and subclassed for various types in the future.

camillescott:
@ctb i'm sure the compact dbg stuff could use these fixtures, but that's out of scope of this PR

camillescott:
TODO for me: add a test for repeat sequences.

In particular, to cover a bug discovered in the NonLoopingAT, where the cursor was only added the seen-set when `next_symbol` was called. When neighbors were found with the `neighbors` function, the HDN junction was not added, allowing a "leak".

betatim:
To help getting this reviewed and merged, can we take parts of this PR (starting from the lowest layer working upwards), split it out to a separate PR, review and merge it? Github's tools for seeing what we discussed already and what not etc are a bit crappy. It would also structure things a bit in terms of what is done, what is still in flux, etc. I would like that but obviously up to @camillescott.

(There must be some law of software dev somewhere stating that quality of code review declines exponentially with size of code to be reviewed)

ctb:
I'm a bit confused here -- @camillescott is this PR fully baked at this point, other than changes requested under review? If so, @betatim I didn't think it was that big & obnoxious but I can try to pull out bits and pieces. If it's still being updated by Camille tho I am not that thrilled about that approach.

camillescott:
@ctb i'd call it baked and ready to be frosted -- further work would be best served by having this merged.

I see little point in pulling it apart into smaller pieces -- it's all quite tightly coupled, and I think, well tested. Perhaps the refactoring of the `Traverser` class and its use elsewhere in the code, along with the introduction of `alphabets`, could be separated, but I'm -1 on being burdened by more logistics when I've got a clock ticking for getting actual research done. @ctb will have to make the final call there.

ctb:
On Thu, Oct 13, 2016 at 10:53:13AM -0700, Camille Scott wrote:
> @ctb i'd call it baked and ready to be frosted -- further work would be best served by having this merged.

lol

> I see little point in pulling it apart into smaller pieces -- it's all quite tightly coupled, and I think, well tested. Perhaps the refactoring of the `Traverser` class and its use elsewhere in the code, along with the introduction of `alphabets`, could be separated, but I'm -1 on being burdened by more logistics when I've got a clock ticking for getting actual research done. @ctb will have to make the final call there.

Right, I would be pulling it apart as part of the code review :).

camillescott:
In that case, carry on :)

ctb:
yes ma'am! :)

betatim:
If its done then that is fine, I just wasn't sure if there was more to be added in which case we could have started merging stuff.

lib/assembler.cc
```diff
@@ -0,0 +1,393 @@
...
+#endif
+            paths.push_back(segment);
+            continue;
+        }
+    }
+}
+
+
+
+/*
```
betatim: delete me?

betatim:
I've read the c++ a few times now and we could probably discuss for ages on rearranging things and why stuff is done the way it is but given that the tests pass and that there are quite a few comments in the c++ part (yay!) we could merge this and if someone feels strongly enough about it move things around later.

For the tests, what do you think about my comment about using a separate RNG instance everywhere? Mainly tedious leg work to add it :-/

I've not looked at the `sandbox/` scripts.

ctb:
I am +1 on merging it.  @betatim can you extract your RNG comment into an issue?

ctb:
I guess since I made the last commit, I can't review. So, @betatim or @camillescott, go ahead and review & merge.

betatim:
One more merge of master into this branch, then a second commit that only updates the formatting according to `make format`.

betatim:
🍻 good work everyone! Thanks for the patience with my complaining about code volume @camillescott .

PR #1431, opened 2016-09-03
Removal of all of the partitioning code, for better/safer hash function refactoring.

ctb:
The partitioning and labelhash code was getting in the way of trying out new hash functions, so I removed it all.  Please don't merge :).

It subtracted pretty cleanly, so we might explore ways to split off and isolate this code in the main branch so that the rest of the codebase becomes more labile.

cc @betatim @camillescott @standage


ctb:
p.s. the tests do all pass :)

ctb:
Note to self: it should be possible to implement the reversibility without using _revhash, by using a map.

ctb:

PR #1432, opened 2016-09-03
Simplify the hashing code by removing sliding window optimization &c.

ctb:
Starting from #1431 (removal of all partitioning code), simplify the hashing code by removing various confounding optimizations.  Eliminates code that requires reversibility, which also breaks the `_revhash` function.

Almost all of the tests pass, including the Traverser and KmerIterator code.  The only tests that were removed were the ones that started traversing the graph from a hash value rather than from a string.


ctb:

Issue #1426, opened 2016-08-26
Support for k>32


standage:
Supporting k values > 32 has been on the table for a while (since #27 and #60 in 2013). Some progress was made on this with #624, but this PR has stalled.

I'd like to suggest that we prioritize this feature.

From my reading of the relevant threads, there seem to be a couple of possible approaches and concerns.
- If we quadruple the size of the hash function output (256 bits instead of the current 64), we can maintain the hashing properties we currently enjoy with k <= 32 (a perfect reversible hash). However, 256 bit unsigned integers are not a native data type, and so we'd have have to find a portable drop-in library with a compatible license. We'd also be increasing memory usage, although for some (all?) applications this won't be a big deal since the hashes simply index into a table.
- If we compile khmer with support for multiple hash functions, we can choose the hash function dynamically. For applications that require reversibility (sparse hash), we will continue to use the 2bit representation and require k <= 32.

Am I reading things correctly? Anything I've missed?

ctb:
Yep, seems about right.  The reversibility is the biggest problem; for probably 95% of the code that we actually currently use, we could go straight for a cyclic hash of some sort and it wouldn't matter, because all we're doing is indexing into hash tables.  But there is a lot of code that relies on reversibility and we have to figure out what to do with it, even if it's not necessarily highly used code at the moment.  Unfortunately I am also probably the primary expert on this code (the partitioning code)...

My primary concern with support for multiple hash functions is around performance, note; virtual functions and the like can give significant performance hits.

I suspect there are some relatively small refactorings that could be done to make this all much easier, but I don't know enough C++ to be confident in the right approach here.

camillescott:
* A simple struct holding 2-4 long longs would be easy to implement. We
rarely do any math on the actual hash values, so there's no need to use an
external library with integer types.
* Virtual functions do introduce a small performance hit which usually
washes out; I think we call our hash functions enough to justify avoiding
them, However, templating a function parameter into the hash function would
shift the performance hit to compile-time and not be overly complicated.

standage:
Regarding the first point, I think I grok the idea of hashing in pieces (similar to the `ARCH_32` approach taken in [Mash's hash function](https://github.com/marbl/Mash/blob/master/src/mash/hash.cpp)?). What I'm *not* so clear on is how to then cast these multiple values as a single value to use as a table index. Probably very simple, but not clear to me at the moment.

[...]

PR #1433, opened 2016-09-03
Swap murmurhash in as primary hash function.

ctb:
Starting from #1432 (simplification of hash function code), swap in murmurhash as the default hash function for khmer.

This will support k > 32.  cc #1426 

The tests that are broken are either checks on the exact hash values, OR are sensitive to collisions (a lot of the occupancy numbers are like this), OR are broken for unknown reasons - I'm not sure why the graph alignment code is broken, in particular.  Still, we're down to 15 broken tests which is pretty good...

### DIFF
--- a/lib/kmer_hash.cc
+++ b/lib/kmer_hash.cc
@@ -66,14 +66,10 @@ HashIntoType _hash_forward(const char * kmer, WordLength k)
         throw khmer_exception("Supplied kmer string doesn't match the underlying k-size.");
     }
 
-    HashIntoType h = 0;
-
-    h |= twobit_repr(kmer[0]);
-
-    for (WordLength i = 1; i < k; i++) {
-        h = h << 2;
-        h |= twobit_repr(kmer[i]);
-    }
+    HashIntoType out[2];
+    uint32_t seed = 0;
+    MurmurHash3_x64_128((void *)kmer, k, seed, &out);
+    HashIntoType h = out[0];
 
     return h;
 }


ctb:
Note to self: it should be possible to implement the reversibility without using _revhash, by using a map.

betatim:
Does it make sense to investigate a general purpose string/byte hashing algorithm? MH is pretty speedy as far as hash functions go but it is still ~100x slower than `master`. Considering that `khmer` doesn't need a hash function that can handle all 256 values of a byte but only maybe 10, does it make sense to brew a custom hash function? It seems like you should be able to capitalise on this limitation. *thinkingoutloud*

ctb:
Good question ;). Poor man's solution could be this:

* hash DNA into two bit form;
* use the bytes emerging from that two bit form as the input to a
  normal hash function.

Should be much faster, too!

I do think we should avoid developing our own hash function because it's
an area with a lot of already developed expertise.

betatim:
> I do think we should avoid developing our own hash function because it's
an area with a lot of already developed expertise.

... and it probably is a bit like crypto, you think it will be a one afternoon adventure and two years later you are still developing and discovering what others learnt many years ago.

betatim:
I was thinking that the way you'd design your hash function would be different if you only have 10 values or if you have 256. I'd be surprised if MH runs faster on byte arrays that only contain 10 different values than on one that contains 256 different values.

ctb:
On Mon, Sep 05, 2016 at 08:09:51AM -0700, Tim Head wrote:
> I was thinking that the way you'd design your hash function would be different if you only have 10 values or if you have 256. I'd be surprised if MH runs faster on byte arrays that only contain 10 different values than on one that contains 256 different values.

Right - but if you hash four DNA characters at a time in an 8-bit string
you're effectively hashing the entire range of 256 characters.

ctb:

Issue #1434, opened 2016-09-03
screed doesn't uppercase DNA in loaded records


ctb:
In the course of mucking about with hash functions, I discovered that there are multiple places in the code where bad data is fed to our hashing code.

More specifically,
* our hashing code (`twobit_repr` etc in kmer_hash.hh) intentionally does not handle lower case DNA characters (it treats them as 'G');
* screed does not upper case DNA sequence when it loads it;
* so, any place where screed is used to load sequences has the potential to allow through lowercase DNA, which are then treated as 'G'. 

This is an excellent combination of optimization (avoiding too much input DNA sanitization 'cause it makes hashing slow) with poorly specified library behavior by screed.

This affects a number of tests, which is how I found it.  For example, tests/test_filter_abund.py::test_filter_abund_6_trim_high_abund_Z will fail if you uppercase `record.sequence` in `khmer/thread_utils.py::verbose_loader`.

There are lots of possible solutions -

* amend all the relevant scripts to use uppercase;
* make sure we periodically recompile and run the tests with `KHMER_EXTRA_SANITY_CHECKS` turned on;
* change the default behavior of screed;

but this kind of thing has happened enough in the past that we should probably think about how to check for it more systematically. Thoughts welcome.

ctb:
Hey, this breaks all of our diginorm and trim-low-abund consistency checks. Sweet :)

```
tests/test_script_output.py::test_normalize_by_median_k21_C20_M1e7 FAILED
tests/test_script_output.py::test_normalize_by_median_k21_C15_M1e7 FAILED
tests/test_script_output.py::test_trim_low_abund_k21_C0_M1e7_diginorm FAILED
tests/test_script_output.py::test_trim_low_abund_k21_C0_M1e7_diginorm_dn15 FAILED
tests/test_script_output.py::test_trim_low_abund_k21_C2_M1e7_diginorm_dn15 FAILED
tests/test_script_output.py::test_trim_low_abund_k21_M1e7_C2 FAILED
tests/test_script_output.py::test_trim_low_abund_k21_M1e7_C3 FAILED
tests/test_script_output.py::test_trim_low_abund_k21_M1e7_C4 FAILED
tests/test_script_output.py::test_trim_low_abund_k21_M1e7_C4_variable FAILED
tests/test_script_output.py::test_trim_low_abund_k21_M1e7_C4_variable_Z25 FAILED
[...]

Issue #1262, opened 2015-08-17
Extract processing classes from scripts and write tests for them


ctb:
Several scripts -- normalize-by-median and trim-low-abund (as of #1138), and perhaps others -- have classes that do interesting processing functions but are only tested at the script level. We should extract these classes into the khmer or oxli packages and test the heck out of 'em.

Issue #1036, opened 2015-05-27
how should we handle 'N's in DNA, in our scripts and our codebase?


ctb:
see #1001 for inspiration.

Currently, we turn them into As (explicitly) or Gs (by default).

@drtamermansour suggests that we ignore reads containing Ns (see https://github.com/dib-lab/khmer/pull/1001#issuecomment-104034103)

We could also ignore k-mers containing Ns but count the rest of the read.

Whatever we do, I think we must output number of skipped k-mers and/or reads; otherwise this behavior will be opaque to the user.

drtamermansour:
I would like to highlight the importance of this problem. The current exome dataset I am working with for variant calling has 22% of its reads counting Ns. Different aligners and variant calling software would behave differently with such ambiguous reads.

ctb:
My thinking has been that, going forward, we should pass Ns on unchanged as much as possible; it doesn't seem like it should be our job to "fix" them.  @drtamermansour do you have a specific recommendation for what to do about Ns in the output, as opposed to counting them/not counting them internally?

drtamermansour:
Totally agree. This is not the job of diginorm. Subsequent trimming or
error correction modules should handle this.

ctb:
On Tue, Jun 02, 2015 at 04:50:41AM -0700, Tamer Mansour wrote:
> Totally agree. This is not the job of diginorm. Subsequent trimming or
> error correction modules should handle this.

I'm skeptical that ANYTHING should be done automatically in the case of
filtering before variant calling!

drtamermansour:
Yes. This is what I mean. I am saying that khmer as a package could provide
other solutions for further pre-processing.

ctb:
Just to follow through on this a bit more: there are three issues.

1. How do we handle reads containing Ns internally?
2. How do we transform reads containing Ns by default?
3. What options do we provide for transforming Ns in reads for those who want to do so?

The answers so far seem to be:

(1) is what we're discussing in this issue;

(2) is "we do as little as possible";

(3) is open to discussion; so far the main idea is to provide error correction of some sort that replaces Ns with a corrected base.

ctb:
The proper way to deal with this might be to make it depend on the hash function. Some hash functions will easily support more characters than others...

ctb:
This is something we could handle differently by providing different hash functions (#27).

standage:
[...]

PR #1435, opened 2016-09-03
fix lowercase actgn input handling.

ctb:
This is a fix for #1434.

Adds a 'cleaned_seq' attribute to screed records that should be used whenever k-mer operations are performed, and propagates this attribute through the codebase with associated refactorings of ReadBundle, trim-low-abund, and normalize-by-median.

The key bugfix commit is 7b40857, which makes the key change to normalize-by-median to uppercase sequences before adding them to the graph.  Prior to this, all k-mers containing lower-case characters would simply be ignored.  Note that this changes output formats, so the md5 hashes in test_script_output.py have been updated.  Since this is a bug and not a new feature, I think semantic versioning will permit a 2.x release.  Also note that the output of trim-low-abund.py matches the output of normalize-by-median.py despite quite different implementations, suggesting that both are correct :).

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/utils.py
+++ b/khmer/utils.py
@@ -183,9 +183,7 @@ def broken_paired_reader(screed_iter, min_length=None,
         raise ValueError("force_single and require_paired cannot both be set!")
 
     # handle the majority of the stream.
-    for record in screed_iter:
-        record.cleaned_seq = record.sequence.upper().replace('N', 'A')
-
+    for record in clean_input_reads(screed_iter):
         if prev_record:
             if check_is_pair(prev_record, record) and not force_single:
                 if min_length and (len(prev_record.sequence) < min_length or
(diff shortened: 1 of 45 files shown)


ctb:
- [ ] TODO: add explicit tests of appropriate behavior into broken_paired_reader, other functions.

ctb:
ready for skeptical review - I think this is for @standage :)

scripts/normalize-by-median.py
```diff
@@ -168,24 +168,13 @@ def __call__(self, is_paired, read0, read1):
...
-                passed_filter = True
-
-        if passed_filter:
-            for record in batch:
-                seq = record.sequence.replace('N', 'A')
-                self.countgraph.consume(seq)
+        # if any in batch have coverage below desired coverage, consume &yield
+        if not batch.coverages_at_least(self.countgraph, desired_coverage):
+            for record in batch.reads:
+                self.countgraph.consume(record.cleaned_seq)
```
standage: Cleaner and more concise. I like it!

khmer/utils.py
```diff
@@ -249,24 +248,30 @@ def write_record_pair(read1, read2, fileobj):
...
+def clean_input_reads(screed_iter):
+    for record in screed_iter:
+        record.cleaned_seq = record.sequence.upper().replace('N', 'A')
+        yield record
+
+
 class ReadBundle(object):
     def __init__(self, *raw_records):
         self.reads = [i for i in raw_records if i]
-        self.cleaned_seqs = [r.sequence.replace('N', 'A') for r in self.reads]
```
standage: I guess I'm fine with moving the read cleaning code to a function rather than a class. (Having the cleaned seq as part of the original record is better organization anyway, IMO.) But then that leaves the question of what the `ReadBundle` class is really for. Just aggregation? If so, we need to update the docs from my last PR to make sure we're clear about what code is doing what.

standage:
s/clean/clear/. Won't let me edit review comment. :-/

khmer/utils.py
ctb: Agreed!

Yes, the `ReadBundle` class is about aggregation (pairs/singletons of reads).  I'll go update the docs.

scripts/normalize-by-median.py
ctb: :) yes, a little complicated on the side of double and triple negatives when you dig into it, but nice and concise now that it's done!

ctb:
Ready for review again @standage.  (I wasn't 100% sure of the details of the new github review approach, so ... all you wanted me to do was update the documentation, right?)

standage:
Yep al teh wrods is gud,

ctb:
Note, this refactored code from #1458; just want to link in :). Also bears on #1262 and #1036.

Issue #1436, opened 2016-09-04
Simple benchmarking for the hash function


ctb:
Here's a simple benchmarking script that just creates hashes for a bunch of k-mers in a string:

https://gist.github.com/ctb/ce69abe77a40e4936f42652dfd41cb38

On master currently, it reports 0.17234091000136686.

On the deoptimized version in #1432, it reports 15.948195557000872.  So about a factor of 100x, which seems reasonable.

On the re-optimized version in #1437, it reports 0.17987623100634664, so back in business.

ctb:
I think @betatim's got this.

betatim:
What do you think of making this something that lives in `tests/` so that future-us can find it again? I wouldn't run it on travis and the like as there is probably too much variation.

ctb:
On Tue, Oct 04, 2016 at 06:58:59AM -0700, Tim Head wrote:
> What do you think of making this something that lives in `tests/` so that future-us can find it again? I wouldn't run it on travis and the like as there is probably too much variation.

What about `benchmarks/`?

`tests/` is really for stuff that is run very frequently.

PR #1437, opened 2016-09-04
Add sliding window optimization back in w/irreversible hashing

ctb:
This PR is on top of #1432, and isn't going to be suitable for merging; it's a test PR.

Here, I add back in the sliding window optimization for calculating Kmer and HashIntoType from strings.  I split the use cases so that KmerIterator calculations are distinct from HashIntoType calculations; the former need to have full Kmer information for things like traversal, while the latter can lose information as they are simply used for one-way hashing.

On the benchmark in #1436, this PR performs equivalently to (or a bit better than) the current master code base.  On the flip side, the KmerIterator is about 100 times slower, which is largely because of the string manipulations around tracking exact k-mers.  See discussion in #1426 for more on next steps.

### DIFF
--- a/lib/counting.cc
+++ b/lib/counting.cc
@@ -55,7 +55,7 @@ using namespace khmer:: read_parsers;
 
 BoundedCounterType CountingHash::get_min_count(const std::string &s)
 {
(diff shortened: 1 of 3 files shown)


ctb:

PR #1438, opened 2016-09-05
twobit_repr as lookup table

betatim:
Exploratory branch for #1434. Use a lookup table and a function instead of a macro.

This is a bit naughty (it generate compiler warnings) but using a lookup table to turn
characters into integers has some advantages as well. You don't have to upper case
things anymore and you can have as many comparisons as you want. Not checked
but I assume that the compiler figures out how to inline this kind of function (or do other magic). 

Another unchecked thing is how this goes in terms of SIMD instructions. Having a long sequence of characters which you iterate over with a `for` loop should be a prime target for the compiler (but sometimes it needs (a lot of) help to figure that out).

The bench from #1436 suggests that this is marginally (sometimes I see 0.16 or even 0.15 c.f. 0.17) faster than the macro.

Opinions?

---

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/lib/kmer_hash.cc
+++ b/lib/kmer_hash.cc
@@ -61,6 +61,18 @@ khmer::HashIntoType twobit_repr(const char ch)
     throw khmer::khmer_exception("Sequence contains invalid DNA symbol.");
   }
 }
+const khmer::HashIntoType
+twobit_values_comp[256] = {['A'] = 1+1, ['T'] = 0+1, ['C'] = 3+1, ['G'] = 2+1,
+                           ['a'] = 1+1, ['t'] = 0+1, ['c'] = 3+1, ['g'] = 2+1};
+khmer::HashIntoType twobit_comp(const char ch)
+{
+  const khmer::HashIntoType v(twobit_values_comp[ch]);
+  if (v > 0) {
+    return v - 1;
+  } else {
+    throw khmer::khmer_exception("Sequence contains invalid DNA symbol.");
+  }
+}
 
 
 //
--- a/lib/kmer_hash.cc
+++ b/lib/kmer_hash.cc
@@ -49,12 +49,17 @@ Contact: khmer-project@idyll.org
 using namespace std;
 
 
-const unsigned long long
-twobit_values[256] = {['A'] = 0, ['T'] = 1, ['C'] = 2, ['G'] = 3,
-                      ['a'] = 0, ['t'] = 1, ['c'] = 2, ['g'] = 3};
-khmer::HashIntoType twobit_repr(char ch)
+const khmer::HashIntoType
+twobit_values[256] = {['A'] = 0+1, ['T'] = 1+1, ['C'] = 2+1, ['G'] = 3+1,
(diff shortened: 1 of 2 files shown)


betatim:
Another thing where my `g++` fu is too weak: you get a bunch of these:

```
lib/kmer_hash.cc:53:23: warning: designated initializers are a C99 feature
      [-Wc99-extensions]
twobit_values[256] = {['A'] = 0, ['T'] = 1, ['C'] = 2, ['G'] = 3,
                      ^~~~~~~~~
```

Why do I get C99 warnings? `g++ -Wall` doesn't tell me off for doing this. #educate me please :)

ctb:
On Mon, Sep 05, 2016 at 05:31:01AM -0700, Tim Head wrote:
> The bench from #1436 suggests that this is marginally (sometimes I see 0.16 or even 0.15 c.f. 0.17) faster than the macro.

Very cool - speed was my major concern with this approach, TBH, but I never
looked into it!

How much is it slowed down if you add in a check for an invalid character?

betatim:
After reading your comment about moving to 3bit representation I was wondering if it is possible to make `0` the value of all invalid characters? This would come for free as the array (seems to be, not sure if this depends on the compiler, to be checked) set to zero everywhere. Not sure if it helps having zero as a special value that only occurs for invalid characters. Or is this trying to be too smart in an effort to avoid raising an exception? I'll add the `raise` and benchmark it.

betatim:
Actually, the easiest way to do the check would be to make 0 a special value:

```c++
const unsigned long long
twobit_values[256] = {['A'] = 0, ['T'] = 1, ['C'] = 2, ['G'] = 3,
                      ['a'] = 0, ['t'] = 1, ['c'] = 2, ['g'] = 3};
khmer::HashIntoType twobit_repr(char ch)
{
  HashIntoType v(twobit_values[ch]);
  if (v) { return v; }
  else { raise ... }
}
```

WDYT?

ctb:
LGTM!

betatim:
Until we switch to 3bit representation we could also have a mask and check that.

ctb:
making 0 the "bad" sounds good to me!

betatim:
Added a check which brings the performance down to 0.17 again.

For the moment this maintains the same 2bit representation as previously used.

One thing I am mystified by is this:

```python
dna = "acgtacgtacgtacgttgcagtgactacgtt"
import khmer
def hash(seq, N=31):
     seq = seq[:N].lower()
     print(khmer.forward_hash(seq, N))
     print(khmer.forward_hash(seq.upper(), N))
     print(khmer.forward_hash_no_rc(seq, N))
     print(khmer.forward_hash_no_rc(seq.upper(), N))

hash(dna, N=31)
# prints:
#813826944745809077
#203614676455337261
#813826944745809077
#813826944745809077
```

For some other random string it works as expected.

betatim:
After bd22d3b performance gets worse: 0.19 now 😢

betatim:
Going to close this. Seems less relevant after #1435

Issue #1442, opened 2016-09-09
Storing reversible hashes for k>32


betatim:
Place for thoughts and progress on representing reversible hash values of (arbitrarily) large k-mers.

This is one of the things needed for #1441.

First idea from @camillescott is to switch from using a single `int64` to a struct of integers.

standage:
Another idea discussed was using the 128 bit integer implementation (`__int128`) discussed [here](http://stackoverflow.com/a/33828720/459780).

Pros:
- sticking with a "native" data type
- could support quick, reversible hashing up to k=64 without perturbing the codebase too much

Cons:
- cannot support arbitrary k
- there may be good reasons to support different hash functions anyway, so this would just be kicking the can down the road

Probably not the best idea in the long term, but worth bringing up.

betatim:
Some experiments in [this gist](https://gist.github.com/betatim/f311eeea15e713927ac64773c8fdd9e5). This uses a int128 and int256 implementation I found by [googling](https://github.com/calccrypto/uint256_t). They were useful to quickly get started. Having 33-mers seems to "just work" if you have a bigger int type available.

The second thing I investigated is how to transport these big integers into python land. Python's integer type has no limit and you can create them from an array of bytes `_PyLong_FromByteArray` (which despite underscore seems to be fairly stable). The reverse [`_PyLong_AsByteArray`](https://github.com/python/cpython/blob/40afe6e4de7bf66d7af4cd763e251762033f48d9/Objects/longobject.c#L5215) also exists. On the pure python side there is [`int.from_bytes` and `int.to_bytes`](https://docs.python.org/3.5/library/stdtypes.html#int.to_bytes) which use these.

After this digging around I think implementing the big-ints with a byte array instead of a pair of smaller integers is the way to go given we need to make those arrays for going to/from python.

betatim:
My only worry about using a byte array is that it might be slower?? I will make a small benchmark pitting the uint128_t from the gist against one based on an array to answer that.

Other remark: `int.{to,from}_bytes` is not available in legacy python. The C API however has been there since way back when the tim bot was still active.

ctb:
*blinks* Well, this looks pretty awesome... What's the performance hit?

betatim:
I don't know yet ;) but I might have just achieved one of the tests passing after switching `_hash` to use a class that uses a byte array to store stuff.

ctb:
in #1519 and friends => too slow. #wontfix

ctb:
[...]

PR #1444, opened 2016-09-12
A type to store large hash values (>64bit)

betatim:
Exploratory surgery for #1442 

This uses a byte array to store the hash value. We can template or similar to support "arbitrarily" large hash values.

At the moment trying to make it work and get some tests to pass. After that speeding up the code. it currently also contains various stuff that is useful for development but you might want to remove later.
- [ ] benchmark `bits` vs `bytes` vs `uint64` as type of the array
- [ ] clean up spurious `#include`s
- [ ] asserts or other protection when users try and `|` or `&` values larger than one byte
- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -1671,7 +1671,9 @@ hashtable_get_tags_and_positions(khmer_KHashtable_Object * me, PyObject * args)
 
     PyObject * posns_list = PyList_New(posns.size());
     for (size_t i = 0; i < posns.size(); i++) {
-        PyObject * tup = Py_BuildValue("IK", posns[i], tags[i]);
+        PyObject * tag = _PyLong_FromByteArray(tags[i].bytes.data(),
+                                               tags[i].bytes.size(), 0, 0);
+        PyObject * tup = Py_BuildValue("IO", posns[i], tag);
         PyList_SET_ITEM(posns_list, i, tup);
     }
 
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -1506,7 +1506,8 @@ hashtable_neighbors(khmer_KHashtable_Object * me, PyObject * args)
         HashIntoType h = node_q.front();
         node_q.pop();
         // type K for python unsigned long long
-        PyList_SET_ITEM(x, i, Py_BuildValue("K", h));
+        PyList_SET_ITEM(x, i, _PyLong_FromByteArray(h.bytes.data(),
+                                                    h.bytes.size(), 0, 0));
     }
 
     return x;
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -5090,19 +5090,24 @@ static PyObject * forward_hash_no_rc(PyObject * self, PyObject * args)
 
 static PyObject * reverse_hash(PyObject * self, PyObject * args)
 {
-    HashIntoType val;
+    PyObject * val;
+    HashIntoType hash;
     WordLength ksize;
 
-    if (!PyArg_ParseTuple(args, "Kb", &val, &ksize)) {
+    if (!PyArg_ParseTuple(args, "Ob", &val, &ksize)) {
         return NULL;
     }
+    _PyLong_AsByteArray((PyLongObject *)val,
+                        hash.bytes.data(),
+                        hash.bytes.size(), 0, 0);
+    std::cout << hash.as_ull() << std::endl;
 
     if (ksize > KSIZE_MAX) {
         PyErr_Format(PyExc_ValueError, "k-mer size must be <= %u", KSIZE_MAX);
         return NULL;
     }
 
-    return PyUnicode_FromString(_revhash(val, ksize).c_str());
+    return PyUnicode_FromString(_revhash(hash, ksize).c_str());
 }
 
 static PyObject * murmur3_forward_hash(PyObject * self, PyObject * args)
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -1269,8 +1269,12 @@ hashtable_reverse_hash(khmer_KHashtable_Object * me, PyObject * args)
 {
     Hashtable * hashtable = me->hashtable;
 
+    PyObject * val_o;
     HashIntoType val;
-    if (!PyArg_ParseTuple(args, "K", &val)) {
+    if (!PyArg_ParseTuple(args, "O", &val_o)) {
+        return NULL;
+    }
+    if (!convert_PyObject_to_HashIntoType(val_o, val, 0)) {
         return NULL;
     }
 
@@ -4228,9 +4232,13 @@ labelhash_get_tag_labels(khmer_KGraphLabels_Object * me, PyObject * args)
 {
     LabelHash * labelhash = me->labelhash;
 
+    PyObject * tag_o;
     HashIntoType tag;
 
-    if (!PyArg_ParseTuple(args, "K", &tag)) {
+    if (!PyArg_ParseTuple(args, "O", &tag_o)) {
+        return NULL;
+    }
+    if (!convert_PyObject_to_HashIntoType(tag_o, tag, 0)) {
         return NULL;
     }
 
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -858,10 +858,9 @@ static PyObject * _HashSet_iternext(PyObject * self)
     _HashSet_iterobj * iter_obj = (_HashSet_iterobj *) self;
     SeenSet * hashes = iter_obj->parent->hashes;
     if (*iter_obj->it != hashes->end()) {
-        //PyObject * ret = PyLong_FromUnsignedLongLong(**iter_obj->it);
-
-        PyObject * ret = _PyLong_FromByteArray((unsigned char *)&(**iter_obj->it).bytes,
-                            (**iter_obj->it).bytes.size(), 0, 0);
+        PyObject * ret = _PyLong_FromByteArray((**iter_obj->it).bytes.data(),
+                                               (**iter_obj->it).bytes.size(),
+                                                0, 0);
         (*(iter_obj->it))++;
         return ret;
     }
@@ -1259,10 +1258,7 @@ hashtable_hash(khmer_KHashtable_Object * me, PyObject * args)
     try {
         PyObject * hash;
         HashIntoType h(_hash(kmer, hashtable->ksize()));
-        hash = _PyLong_FromByteArray((unsigned char *)&h.bytes,
-                            h.bytes.size(), 0, 0);
-        //hash = PyLong_FromUnsignedLongLong(_hash(kmer, hashtable->ksize()));
-
+        hash = _PyLong_FromByteArray(h.bytes.data(), h.bytes.size(), 0, 0);
         return hash;
     } catch (khmer_exception &e) {
         PyErr_SetString(PyExc_RuntimeError, e.what());
@@ -2842,9 +2838,8 @@ hashtable_get_kmer_hashes(khmer_KHashtable_Object * me, PyObject * args)
 
     PyObject * x = PyList_New(hashes.size());
     for (unsigned int i = 0; i < hashes.size(); i++) {
-        //PyObject * obj = PyLong_FromUnsignedLongLong(hashes[i]);
-        PyObject * obj = _PyLong_FromByteArray((unsigned char *)&hashes[i].bytes,
-                            hashes[i].bytes.size(), 0, 0);
+        PyObject * obj = _PyLong_FromByteArray(hashes[i].bytes.data(),
+                                               hashes[i].bytes.size(), 0, 0);
         PyList_SET_ITEM(x, i, obj);
     }
 
@@ -5057,10 +5052,8 @@ static PyObject * forward_hash(PyObject * self, PyObject * args)
 
     try {
         PyObject * hash;
-        //hash = PyLong_FromUnsignedLongLong(_hash(kmer, ksize));
         HashIntoType h(_hash(kmer, ksize));
-        hash = _PyLong_FromByteArray((unsigned char *)&h.bytes,
-                            h.bytes.size(), 0, 0);
+        hash = _PyLong_FromByteArray(h.bytes.data(), h.bytes.size(), 0, 0);
         return hash;
     } catch (khmer_exception &e) {
         PyErr_SetString(PyExc_RuntimeError, e.what());
@@ -5090,8 +5083,8 @@ static PyObject * forward_hash_no_rc(PyObject * self, PyObject * args)
     }
 
     HashIntoType h(_hash_forward(kmer, ksize));
-    PyObject * hash = _PyLong_FromByteArray((unsigned char *)&h.bytes,
-                                            h.bytes.size(), 0, 0);
+    PyObject * hash = _PyLong_FromByteArray(h.bytes.data(), h.bytes.size(),
+                                            0, 0);
     return hash;
 }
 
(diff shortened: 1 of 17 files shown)


betatim:
Also unsure where the class should live. `kmer_hash.hh`?

betatim:
I know @ctb suggests ignoring the partitioning stuff, but it would still be interesting to try and comprehend why some of those tests fail.

ctb:
Is now a good time to look?

--



--

ctb:
Wow, that is an esoteric place for things to fail!  But it's something minor and detailed, rather than something big. I can't dig into it right now, but I suggest you go about your business...

betatim:
I think the failure is because there is a bug in the new class I added. Happens for long~ish k-mer sizes that are less than 32. For some reason it diverges from the value you get with a uint64. #test #test #test

ctb:
Nice work! The tests pass for me on Mac OS X.

What's the plan moving forward? Benchmarking, for sure... if there's no major slowdown, it seems to me that the code is clean enough to merge, and then we can start working on supporting k > 32, yah?

betatim:
Yes, I'd like to tidy it up a bit (make sure all the methods I defined make sense, and put in an assert or two for when you pass in arguments that are too big) and work out what we would need to add to make
```
typedef BigHashType HashIntoType;
```
work. I think that will be a whole new PR but doing a quick survey of what is missing might point to fundamental flaws in this idea (or not).

ctb:
The only thing I worry about (other than performance) is suitably generalizing the save-to-disk code. But we have lots of tests for that so we just need to duplicate them for k > 32.

sjackman:
You may find that [`std:bitset`](http://www.cplusplus.com/reference/bitset/bitset/) is more convenient that `std::array` when you need bitwise operators, particularly the bit shift operators (`<<` and `>>`). Most of the operators in `class BigHashType` is implemented in `std::bitset`.
Note that `std::bitsets` is a misnomer. It's an array of bits, not a set of bits.

sjackman:
I just noticed that you're using `std::array<uint8_t, 8>` which is 64 bits. Just to check, the plan is to to eventually increase `8` to a larger value, in a second PR?

betatim:
The overall idea of using an array of bytes is based around `_PyLong_FromByteArray`. This would make it easy to transport large integers from c++ to python land (and back again). Hence bytes and not `uint64` (or single bits) in the array. Not benchmarked yet what the performance difference/trade off is as a function of the size of the type in the array. This is on the to do list.

In the long term you want to make the size of the array configurable (what you really care about is some efficient way to store large integers). Not sure yet if you want to provide just one (fixed at compile time) or several or even one that chooses the size at run time (unlikely to perform well?). First though, stick with 8 bytes and make sure we can reproduce the `uint64` behaviour 🚧 👷 🚧.

sjackman:
Fair enough. It's unfortunate there's no `PyLong_FromUnsignedLongLongArray`

betatim:
[Some benchmarks](https://gist.github.com/betatim/f1351ad06fbb3029a1a5de4029463bf9):

`hashing_basic` is the current implementation with a `uint64`. `bighash` uses an array of `N` integers of type `T`. The last two numbers indicate the length of the kmer (8, 16, 24, 32) and which one of two strings was used (both random sequences). My conclusions from this wall of numbers:

* single `uint64` wrapped in a `BigHashType` is nearly as fast as just a `uint64`
* as you increase the size of the basic type (and reduce `N`) performance improves
* `hashing_bighash<uint8_t, 8>` is a bad idea?!
* (not shown here but in the benchmark `bitset` is slower than `hashing_bighash<uint8_t, 8>`)
* converting from a `BigHashType` to a `uint64` is pretty slow, we should avoid doing that when ever possible (-> compute the modulus directly)

If you have some experience with `google-benchmark` I'd like to chat with you. With compilers being super sneaky these days I worry that these benchmarks don't actually benchmark what I want to test.

```
Benchmark                                  Time           CPU Iterations
------------------------------------------------------------------------
hashing_basic/8/1                         20 ns         40 ns   17541837
hashing_basic/8/2                         20 ns         23 ns  178492200
hashing_basic/16/1                        44 ns         52 ns   86685160
hashing_basic/16/2                        43 ns         49 ns   84504320
hashing_basic/24/1                        63 ns         72 ns   54933840
hashing_basic/24/2                        64 ns         71 ns   52610220
hashing_basic/32/1                        87 ns         98 ns   41023000
hashing_basic/32/2                        90 ns        177 ns    3775009
hashing_bighash<uint8_t, 8>/8/1          104 ns        107 ns   10000000
hashing_bighash<uint8_t, 8>/8/2          102 ns        203 ns    3449941
hashing_bighash<uint8_t, 8>/16/1         213 ns        228 ns   15286310
hashing_bighash<uint8_t, 8>/16/2         215 ns        430 ns    1642105
hashing_bighash<uint8_t, 8>/24/1         321 ns        641 ns    1000000
hashing_bighash<uint8_t, 8>/24/2         319 ns        638 ns    1113550
hashing_bighash<uint8_t, 8>/32/1         450 ns        523 ns    8079410
hashing_bighash<uint8_t, 8>/32/2         453 ns        499 ns    7537900
hashing_bighash<uint16_t, 4>/8/1          66 ns        131 ns    5386107
hashing_bighash<uint16_t, 4>/8/2          71 ns        139 ns   10000000
hashing_bighash<uint16_t, 4>/16/1        141 ns        155 ns   24197510
hashing_bighash<uint16_t, 4>/16/2        135 ns        150 ns   25038090
hashing_bighash<uint16_t, 4>/24/1        202 ns        227 ns   16956210
hashing_bighash<uint16_t, 4>/24/2        203 ns        227 ns   16856020
hashing_bighash<uint16_t, 4>/32/1        272 ns        543 ns    1314850
hashing_bighash<uint16_t, 4>/32/2        271 ns        313 ns   13142580
hashing_bighash<uint32_t, 2>/8/1          30 ns         34 ns  116678330
hashing_bighash<uint32_t, 2>/8/2          29 ns         31 ns  108268630
hashing_bighash<uint32_t, 2>/16/1         61 ns        121 ns    5961404
hashing_bighash<uint32_t, 2>/16/2         62 ns         71 ns   58215930
hashing_bighash<uint32_t, 2>/24/1         93 ns        102 ns   36724590
hashing_bighash<uint32_t, 2>/24/2         90 ns        180 ns    3883926
hashing_bighash<uint32_t, 2>/32/1        122 ns        139 ns   29083780
hashing_bighash<uint32_t, 2>/32/2        123 ns        140 ns   28921320
hashing_bighash<uint64_t, 1>/8/1          22 ns         23 ns  100000000
hashing_bighash<uint64_t, 1>/8/2          20 ns         40 ns   17014326
hashing_bighash<uint64_t, 1>/16/1         42 ns         48 ns   82092180
hashing_bighash<uint64_t, 1>/16/2         42 ns         49 ns   84606460
hashing_bighash<uint64_t, 1>/24/1         64 ns         74 ns   55392890
hashing_bighash<uint64_t, 1>/24/2         70 ns         81 ns   52367770
hashing_bighash<uint64_t, 1>/32/1         88 ns         98 ns   39777700
hashing_bighash<uint64_t, 1>/32/2         86 ns        170 ns    4043905
hashing_bighash<uint64_t, 2>/8/1          37 ns         72 ns   10000000
hashing_bighash<uint64_t, 2>/8/2          34 ns         68 ns   10315659
hashing_bighash<uint64_t, 2>/16/1         69 ns         80 ns   52080230
hashing_bighash<uint64_t, 2>/16/2         70 ns         80 ns   50454090
hashing_bighash<uint64_t, 2>/24/1        109 ns        129 ns   34405140
hashing_bighash<uint64_t, 2>/24/2        109 ns        116 ns   10000000
hashing_bighash<uint64_t, 2>/32/1        148 ns        296 ns    2319986
hashing_bighash<uint64_t, 2>/32/2        146 ns        163 ns   23393220
hashing_bighash<uint64_t, 2>/40/1        182 ns        209 ns   19664030
hashing_bighash<uint64_t, 2>/40/2        181 ns        201 ns   18983150
```

sjackman:
> If you have some experience with google-benchmark I'd like to chat with you.

I don't, sorry.

> With compilers being super sneaky these days I worry that these benchmarks don't actually benchmark what I want to test.

Add the `-save-temps` option to GCC, which will save the assembler code in a `.s` file. You can then check the function you're testing to see which instructions it uses. If for example you're testing the `>>` operator, there should be a `shr` instruction in the function that you're testing. It can also give insight into why one implementation is faster than another. For example, `hashing_bighash<uint64_t, 2>` would almost certainly be loop unrolled, whereas `hashing_bighash<uint8_t, 16>` might be implemented with a loop.

betatim:
Some kind of progress. Most of the last commit is disentangling the use of `HashIntoType` to mean "I want a hash" and "I want a big integer to use as index/for counting stuff". (I think)

Tempted to make this part a separate PR as it touches a lot of different parts of the code. The hope is to make it easier to review.

Right now this is still a bit of a construction site. Especially the c <-> python interface is a hack.

standage:
I agree that using `HashIntoType` as a large index is unnecessarily confusing. I also agree this might be better submitted as a separate PR, to make it easier to review the changes.

ctb:
Shall I look at this commit or wait for you to cherry pick it out to a new
PR? :)

betatim:
Wait please.

lib/hashtable.hh
```diff
@@ -123,11 +123,13 @@ protected:
 
     void _init_bitstuff()
     {
-        bitmask = 0;
+      #if 1
+        //bitmask = 0;
```
betatim: commenting this out is a bug?

lib/kmer_hash.cc
```diff
@@ -40,6 +40,7 @@ Contact: khmer-project@idyll.org
 #include <string.h>
 #include <algorithm>
 #include <string>
+#include <iostream>
```
betatim: remove me

lib/kmer_hash.cc
```diff
@@ -203,16 +204,13 @@ KmerIterator::KmerIterator(const char * seq,
                            unsigned char k) :
     KmerFactory(k), _seq(seq)
 {
-    bitmask = 0;
```
betatim: bug?

lib/kmer_hash.hh
```diff
@@ -38,11 +38,14 @@ Contact: khmer-project@idyll.org
 #ifndef KMER_HASH_HH
 #define KMER_HASH_HH
 
+#include <array>
+#include <iostream>
```
betatim: remove me

lib/kmer_hash.hh
```diff
@@ -113,6 +116,117 @@ HashIntoType _hash_murmur(const std::string& kmer,
                           HashIntoType& h, HashIntoType& r);
 HashIntoType _hash_murmur_forward(const std::string& kmer);
 
+
+template <typename T, std::size_t N>
```
betatim: given that this can't work with anything but `unsigned char` at the moment, should we remove the type template argument?

lib/kmer_hash.hh
```diff
@@ -163,7 +277,7 @@ public:
     /// @warning The default constructor builds an invalid k-mer.
     Kmer()
     {
-        kmer_f = kmer_r = kmer_u = 0;
+        //kmer_f = kmer_r = kmer_u = 0;
```
betatim: initialiser list instead? Needs some kind of init.

lib/subset.cc
```diff
@@ -515,17 +515,18 @@ void SubsetPartition::do_partition(
     CallbackFn		callback,
     void *		callback_data)
 {
+    HashIntoType empty;
```
betatim: add `operator bool()` instead of this?

lib/subset.cc
```diff
@@ -546,7 +547,7 @@ void SubsetPartition::do_partition(
 
         // run callback, if specified
         if (total_reads % CALLBACK_PERIOD == 0 && callback) {
-            cout << "...subset-part " << first_kmer << "-" << last_kmer << ": "
+            cout << "...subset-part " << first_kmer.as_ull() << "-" << last_kmer.as_ull() << ": "
```
betatim: @ctb what is the intention of this code? What if the hash does not fit in 64bit, what should this print?

lib/traversal.cc
```diff
@@ -43,7 +43,7 @@ using namespace std;
 Traverser::Traverser(const Hashtable * ht) :
     KmerFactory(ht->ksize()), graph(ht)
 {
-    bitmask = 0;
+    bitmask;
```
betatim: bug?

betatim:
The two tests that are failing are related to storing tags as binary format on disk. The test checks that the size of what was written to disk is "as expected", which now fails. That kinda makes sense as the `BigHashType` is bigger?? I think it'll be faster to get some advice from someone who knows the code than for me to figure this out.

Otherwise things compile and the tests pass! Wohoo! There are a few bits that are quite a mess, so I will tidy those up and address the comments I made above. If you want to start reviewing and help point out the messy parts, please do so.

---

The [benchmark](https://gist.github.com/ctb/ce69abe77a40e4936f42652dfd41cb38) runs in ~0.58, so several times slower than master. Nothing that can't be fixed, hopefully. Probably worth investigating the switch to `uint64`s: https://github.com/dib-lab/khmer/pull/1444#issuecomment-247038018

khmer/_khmer.cc
```diff
@@ -5064,15 +5090,32 @@ static PyObject * forward_hash_no_rc(PyObject * self, PyObject * args)
...
+    PyObject * val;
+    HashIntoType hash;
     WordLength ksize;
 
-    if (!PyArg_ParseTuple(args, "Kb", &val, &ksize)) {
+    if (!PyArg_ParseTuple(args, "Ob", &val, &ksize)) {
+        return NULL;
+    }
+    if (PyLong_Check(val)) {
+        _PyLong_AsByteArray((PyLongObject *)val,
```
ctb: This seems like a nice place to write a macro - replace _PyLong_AsByteArray(...) with a standard function/macro whatever, that could work for whatever we end up doing.

lib/hashtable.cc
```diff
@@ -163,7 +163,6 @@ unsigned int Hashtable::consume_string(const std::string &s)
 
     while(!kmers.done()) {
         HashIntoType kmer = kmers.next();
-
```
ctb: ?

lib/hashtable.cc
```diff
@@ -1080,4 +1079,3 @@ const
 }
 
 // vim: set sts=2 sw=2:
-
```
ctb: ?

lib/hashtable.hh
ctb: I don't understand.

lib/subset.cc
ctb: doesn't *really* matter. It's just diagnostic output that can be removed w/o a problem.

ctb:
Overall, this is *really* nice - thanks!

I think with virtually no effort you could backport most of these patches into the codebase immediately, and eliminate all but the hash-function specific stuff from consideration. What do you think?

lib/hashtable.hh
betatim: this was more a note-to-self than anything else.

betatim:
Not sure I get what you mean 😕 what is the hash function specific stuff?

Contrary to what I wrote before I am now not so sure anymore about lifting bits out of this PR into smaller ones.

---

could you take a look at the failing tests? Can I safely increase the size it compares to or ...?

ctb:
On Thu, Sep 22, 2016 at 06:13:40AM -0700, Tim Head wrote:
> Not sure I get what you mean ???? what is the hash function specific stuff?
> 
> Contrary to what I wrote before I am now not so sure anymore about lifting bits out of this PR into smaller ones.

The HashIntoType => uint64 stuff is easy to extract, no?

> could you take a look at the failing tests? Can I safely increase the size it compares to or ...?

yes.

betatim:
This branch is becoming a place to try things. All the useful things are being extracted into smaller PRs:

* #1454
* #1455

ctb:
@betatim could you write up a brief summary of the > 64 bit hash value stuff on
a new issue, and link in the various trial efforts you made (and just closed
:)?

documentation FTW :)

Issue #1441, opened 2016-09-09
Plan for hash function refactoring


ctb:
Problem under discussion:
- We want to support multiple hash functions in khmer.
- Some uses of the hash function require reversibility, others do not:
- Partitioning requires exact storage of some k-mers
- Traversal requires reversibility of some k-mers
- We are concerned about performance so things like virtual functions are being approached with care.
- We want some of the hash functions to support k > 32. It is not clear if the reversible ones need to support k > 32, at least in the short term.

The two proximal use cases are Daniel’s need for k > 32 for human genomics work, and Camille’s work on assembly and traversal. Titus is also interested in supporting k > 32 for the compact De Bruijn graph work.

See: https://github.com/dib-lab/khmer/issues/1426

A few notes:
- The partitioning code is a big, unwieldy mess but it is fairly easy to isolate. Titus would like to avoid major refactoring of this code, and would also like to keep it working.  But it can do so at k <= 32.
- Quite a bit of the current code doesn’t need reversible hashing, and there may be performance reasons to add irreversible hashing as a thing. Maybe there is an opportunity to split the use cases up and align our software architecture appropriately. Or, maybe there is no performance reason - needs to be checked.
  - Do you need to be able to use k>32 and assembly/traversal at the same time? (Yes, I think so, in the medium term. Need a hash function that is reversible for k>32?)
  - Any rolling hash function can satisfy the performance needs for hashing, Titus believes.
- There’s a public relations reason to go with k up to 100 or 150; this might not be reasonable from a performance reason.

Current proposal:
- Develop a bitstring representation that is reversible and permits “rolling” updates - (adding a base for traversal). Provisionally, we’d like to see if this can support k > 64 without disastrous performance degradation. Compile time flags that let you fix your upper k size are possible, but add to maintenance burden. (struct of ints to store the hash)
- Figure out the appropriate C++ incantations to support multiple hash functions. The current guess is that templating the hashtable class with a functor will be highest performance and sufficiently flexible.
  If we can get the C++ architecture right, it seems like we can quickly evaluate performance questions at that point.
- Titus believes that it is probably ok to trial code in #1431, until we figure out the right architecture.
- Is there something we can do in the architecture to more cleanly split out the partitioning code to ease the eventual merge into master of the new hashing code?  Note that performance of the partitioning code is not so important at the moment. 
- Perhaps introduce a new class between hashtable and countgraph?

TODO items:

[...]

PR #1450, opened 2016-09-20
Hash Function Arch Refactoring

camillescott:
This is the working branch for my part of the hash function refactor re: #1441. It will eventually need to be resolved with #1442. The main goal of this branch will be to remove the use of the global `_hash` function and its siblings, and instead replace them with a Functor templated into the classes that require hashing. In pursuit of this, there will be some reorganization of the class hierarchy for clarity and modularity.
---
- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
I'm going to close this in favor of #1495.

PR #1454, opened 2016-09-23
Refactor uint64_t vs HashIntoType

betatim:
In some places `HashIntoType` was used instead of a larger integer. This PR fixes that. While it makes no difference in the current stage as the two are synonyms in the future we will have a special type for hash values which is not equal to `uint64_t` anymore.

Prep work for #1444 

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -954,17 +954,11 @@ static PyObject * khmer_HashSet_concat_inplace(khmer_HashSet_Object * o,
 
 static int khmer_HashSet_contains(khmer_HashSet_Object * o, PyObject * val)
 {
-    if (PyInt_Check(val) || PyLong_Check(val)) {
-        HashIntoType v;
-        if (PyLong_Check(val)) {
-            v = PyLong_AsUnsignedLongLong(val);
-        } else {
-            v = PyInt_AsLong(val);
-        }
-
-        if (set_contains(*o->hashes, v)) {
-            return 1;
-        }
+    HashIntoType v;
+    if (convert_PyObject_to_HashIntoType(val, v, 0)) {
+      if (set_contains(*o->hashes, v)) {
+          return 1;
+      }
     }
     return 0;
 }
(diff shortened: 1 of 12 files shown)


betatim:

khmer/_khmer.cc
```diff
@@ -3588,11 +3578,12 @@ static PyObject* _new_counting_hash(PyTypeObject * type, PyObject * args,
...
             if (PyLong_Check(size_o)) {
-                sizes.push_back((HashIntoType) PyLong_AsUnsignedLongLong(size_o));
+                sizes.push_back(PyLong_AsUnsignedLongLong(size_o));
             } else if (PyInt_Check(size_o)) {
-                sizes.push_back((HashIntoType) PyInt_AsLong(size_o));
+                sizes.push_back(PyInt_AsLong(size_o));
             } else if (PyFloat_Check(size_o)) {
-                sizes.push_back((HashIntoType) PyFloat_AS_DOUBLE(size_o));
+                // XXX really? should be Float_AS_INT?
+                sizes.push_back(PyFloat_AS_DOUBLE(size_o));
```
standage: Do we lose range with float --> int conversion? We certainly don't gain any precision by converting float to double. Not a bit deal though, we can discuss in a separate issue thread if we feel it's necessary.

tests/test_functions.py
```diff
@@ -98,6 +102,33 @@ def test_reverse_hash():
...
+
+    s = khmer.reverse_hash(long(255), 4)
+    assert s == "GGGG"
+
+
+def test_reverse_hash_raises():
+    with pytest.raises(TypeError) as excinfo:
+        khmer.reverse_hash('2345', 4)
+
+    assert 'must be int' in str(excinfo.value)
```
standage: Looks like this test is passing on Travis Linux builds, but not Mac. Not sure why on first look.

khmer/_khmer.cc
betatim: What I was thinking of was that `vector<uint64_t> sizes` and that I wasn't sure if that turns 135.64 into 135 or 136.

tests/test_functions.py
betatim: Looks like this is due to slightly different error messages. Made the check less stringent.

khmer/_khmer.cc
betatim: Removed the comment, we can create a new issue if we want to discuss this.

ctb:
LGTM.

standage:
LGTM

betatim:
🔢 all the numbers all the time ;)

PR #1455, opened 2016-09-26
Beyond 64bit hash values

betatim:
This is another extract from #1444. This implements a type that can be used with hash values >64bits.
- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/lib/hllcounter.cc
+++ b/lib/hllcounter.cc
@@ -259,7 +259,7 @@ double ep_sum(double acc, int b)
 
 int get_rho(HashIntoType w, int max_width)
 {
-    return max_width - floor(log2(w));
+    return max_width - floor(log2(w.as_ull()));
 }
 
 HLLCounter::HLLCounter(double error_rate, WordLength ksize)
@@ -348,7 +348,7 @@ uint64_t HLLCounter::estimate_cardinality()
 void HLLCounter::add(const std::string &value)
 {
     HashIntoType x = khmer::_hash_murmur(value);
-    uint64_t j = x & (this->m - 1);
+    uint64_t j = x.as_ull() & (this->m - 1);
     this->M[j] = std::max(this->M[j], get_rho(x >> this->p, 64 - this->p));
 }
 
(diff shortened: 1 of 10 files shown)


lib/hllcounter.cc
```diff
@@ -348,7 +348,7 @@ uint64_t HLLCounter::estimate_cardinality()
 void HLLCounter::add(const std::string &value)
 {
     HashIntoType x = khmer::_hash_murmur(value);
-    uint64_t j = x & (this->m - 1);
+    uint64_t j = x.as_ull() & (this->m - 1);
```
betatim: @camillescott what is the right thing to do here? `x` can now be >64bit but we want to store it in a `uint64` 😔

lib/hllcounter.cc
```diff
@@ -259,7 +259,7 @@ double ep_sum(double acc, int b)
 
 int get_rho(HashIntoType w, int max_width)
 {
-    return max_width - floor(log2(w));
+    return max_width - floor(log2(w.as_ull()));
```
betatim: @camillescott same as below, is converting down from >64bit to 64bit the right thing to do?

lib/kmer_hash.hh
```diff
@@ -113,6 +116,142 @@ HashIntoType _hash_murmur(const std::string& kmer,
                           HashIntoType& h, HashIntoType& r);
 HashIntoType _hash_murmur_forward(const std::string& kmer);
 
+
+class BigHashType {
+public:
+  std::array<uint8_t, 10> bytes{{0}};
```
betatim: make me private?!

lib/hllcounter.cc
luizirber: `x` is still 64bit because `_hash_murmur` returns 64bit values, but I can backtrack on the calculations here and avoid hardcoding for 64 bits.

betatim:
Waiting for #1462

PR #1458, opened 2016-09-27
Extract ReadBundler to utils

standage:
See #1036. The `clean_up_reads` function was unnecessarily verbose, so I integrated it with the ReadBundler constructor (the only place it was called) and added a couple of property methods.

I added a simple test, which should provide coverage of the entire ReadBundler class sans the `coverages` method.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
- [ ] <strike>Did it change the command-line interface? Only additions are allowed
  changes were made?
- [ ] <strike>Do the changes respect streaming IO? (Are they


standage:
Ready for review @betatim @camillescott @ctb @luizirber

betatim:
I vote for docs + edge case testing. Low hanging fruit issue for the further refactor.

standage:
Added tests, documentation, and new issue for further refactoring (#1466). Should be ready for merge now.

tests/test_read_handling.py
```diff
@@ -777,3 +777,46 @@ def test_extract_paired_reads_5_stdin_error():
...
+    assert bundle.num_reads == 2
+    assert bundle.agg_length == 200
+    assert bundle.cleaned_reads[0] == cleaned_reads[0]
+    assert bundle.cleaned_reads[1] == cleaned_reads[1]
+
+    for (rd, cln), raw, tstcln in zip(bundle.both(), raw_reads, cleaned_reads):
+        assert rd.sequence == raw
+        assert cln == tstcln
+
+    infile = utils.get_test_data('single-read.fq')
```
ctb: In general each test should look at one kind of behavior, so that you don't fix one thing and still have the test failing. So, please break this up into 4 functions.

khmer/utils.py
```diff
@@ -249,5 +249,25 @@ def write_record_pair(read1, read2, fileobj):
...
+
+    def both(self):
+        return zip(self.reads, self.cleaned_reads)
+
+    @property
+    def num_reads(self):
+        return len(self.cleaned_reads)
+
+    @property
+    def agg_length(self):
```
ctb: we don't use 'agg' (presumably for 'aggregate') anywhere else in the code base - how about 'sum_length' or 'total_length'?

standage:
Same!

PR #1462, opened 2016-09-29
Abstract hash <-> pyobject conversion

betatim:
Add a layer of abstraction for converting hash values into python
objects (and back). This is in preparation for supporting our
own hash type that is not a plain integer.

---

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -109,6 +109,32 @@ extern "C" {
 
 /***********************************************************************/
 
+// Convert a hash to a python long object.
+static bool convert_HashIntoType_to_PyObject(const HashIntoType &hashval,
+                                             PyObject **value) {
+      *value = PyLong_FromUnsignedLongLong(hashval);
+      return true;
+}
+
+
+// Convert a python long to a hash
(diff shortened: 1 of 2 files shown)


khmer/_khmer.cc
```diff
@@ -109,6 +109,32 @@ extern "C" {
 
 /***********************************************************************/
 
+// Convert a hash to a python long object.
+static bool convert_HashIntoType_to_PyObject(const HashIntoType &hashval,
+                                             PyObject **value) {
```
betatim: Dilemma: would like to keep the interface similar to the other `convert_*` functions but doing this `PyObject **value` and `convert_HashIntoType_to_PyObject(h, &ptr_to_pyobj)` double take seems extremely ugly. Any better ideas?

betatim:

khmer/_khmer.cc
standage: I actually think it's fine.

standage:
LGTM, but I'd feel more comfortable if someone with more experience with the CPython code gave the official sign off.

ctb:
Looks good - update & merge!

PR #1485, opened 2016-10-06
use `make pep8` target in travis and rebel against fascist whitespace tyranny

camillescott:
Whitespace in arithmetic expressions is hard to read in slices. Fight the power.
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

…whitespace

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -173,10 +173,10 @@ cppcheck-long: FORCE
 
 ## pep8        : check Python code style
 pep8: $(PYSOURCES) $(wildcard tests/*.py)
-	pep8 --ignore E309,E226 setup.py khmer/*.py scripts/*.py tests/*.py oxli/*.py
+	pep8 setup.py khmer/*.py scripts/*.py tests/*.py oxli/*.py
 
 pep8_report.txt: $(PYSOURCES) $(wildcard tests/*.py)
-	pep8 --exclude=_version.py setup.py khmer/ scripts/ tests/ oxli/ \
+	pep8 setup.py khmer/ scripts/ tests/ oxli/ \
 		> pep8_report.txt || true
 
 diff_pep8_report: pep8_report.txt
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -172,8 +172,7 @@ cppcheck-long: FORCE
 
 ## pep8        : check Python code style
 pep8: $(PYSOURCES) $(wildcard tests/*.py)
-	pep8 --exclude=_version.py  --show-source setup.py khmer/ \
-		scripts/ tests/ oxli/ || true
+	pep8 --ignore E309,E226 setup.py khmer/*.py scripts/*.py tests/*.py oxli/*.py
```
standage: Barring a good reason, we should not get rid of `--exclude=_version.py`.

luizirber: This is automatically generated by `versioneer`, so probably a good idea do exclude it from pep8.

mr-c: You may want to add pep8 settings to `setup.cfg` or `tox
ini` instead.

betatim:
+1 on adding (as many as possible of) the command-line arguments to a config file. This would also reduce the amount of out-of-sync-ness between `make pep8` and `make autopep8`. If we don't move things to a config that target should be updated as well.

Is there a way to disable the whitespace around arithmetic check only for `tests/`? In general it is a good rule of thumb (to be disabled on a line by line basis if you must).

On the fight the power front: after this PR builds will still fail when `pep8` is unhappy. I am Ok with that but we should discuss this once and for all now. For example I would be -1 on adding more exceptions to `pep8` and rather not have it enforced by travis.

betatim:
Do you want to polish off the minor comments above @camillescott or can I take this PR?

ctb:
executive decision: => @betatim please handle

betatim:
@standage I think we need you to push some kind of button if your review comments have been addressed.

standage:
Original comment still stands: `--exclude=_version.py` should not have been removed.

betatim:
It is here now: https://github.com/dib-lab/khmer/pull/1485/files#diff-380c6a8ebbbce17d55d50ef17d3cf906R33

standage:
@betatim Thanks, missed that!

betatim:
Null problemo.

PR #1488, opened 2016-10-11
Beyond 64bit hash values - alternate implementation

betatim:
Alternative way of storing hash values >64bit. Using a `bitset` should be as fast or faster than the `BigHashType` in #1455.

Converting between `PyLong` and a `bitset` is a bit cumbersome as we have to go via a string. Feedback on a better way of doing this welcome.

---

Any sane `bitset` implementation packs the bits into the largest integer available so it should be the same as the homebrew `BigHashType` for large number of bits. With free gains because the STL wizards do good things with vector instructions etc
- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -113,8 +113,7 @@ extern "C" {
 static bool convert_HashIntoType_to_PyObject(const HashIntoType &hashval,
         PyObject **value)
 {
-    *value = _PyLong_FromByteArray(hashval.bytes.data(), hashval.bytes.size(),
-                                   0, 0);
+    *value = PyLong_FromString(hashval.to_string().c_str(), NULL, 2);
     return true;
 }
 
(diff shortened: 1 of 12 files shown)


ctb:
On Tue, Oct 11, 2016 at 05:57:36AM -0700, Tim Head wrote:
> Alternative way of storing hash values >64bit. Using a `bitset` should be as fast or faster than the `BigHashType` in #1455.
> 
> Converting between `PyLong` and a `bitset` is a bit cumbersome as we have to go via a string. Feedback on a better way of doing this welcome.

I can't think of a good reason to make the number available in Python, note.
I think we just need an opaque object that we can stringify for debugging
if need be.

betatim:
You don't want to have the result of `khmer.forward_hash` and similar available in python land?

ctb:
On Tue, Oct 11, 2016 at 06:04:22AM -0700, Tim Head wrote:
> You don't want to have the result of `khmer.forward_hash` and similar available in python land?

I'm not sure we use it anywhere except for testing and debugging.

betatim:
https://giphy.com/gifs/jonathan-pryce-narcopolis-8plTxLcdnBse4

betatim:
On a more serious note: if it isn't used much I would stick with this approach of going via strings (and exclude that from the benchmark).

ctb:
agree

Issue #1368, opened 2016-05-11
C++ I/O exception on Fedora 23


standage:
I get the following issue when running `make test` with the current master on Fedora 23.

```
tests.test_countgraph.test_load_truncated ... terminate called after throwing an instance of 'std::ios_base::failure'
  what():  basic_ios::clear
```

luizirber:
I'm seeing similar errors with Arch Linux, seems to be a change on GCC 5
stdlib

standage:
Hmm, I thought this would have been fixed by #1407, but now on Fedora 23 I'm getting:

```
tests/test_countgraph.py::test_load_truncated Makefile:295: recipe for target 'test' failed
make: *** [test] Aborted (core dumped)
```

PR #1489, opened 2016-10-13
Address the silly gcc bug

camillescott:
Address #1368. There is a mismatch between gcc and system libs which makes it impossible to specifically catch an exception from ifstream::read; the problem is described on the gcc bugtacker:
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=69877
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66145

This was previously only affecting a few users, but now affects the core devs :) the temporary measure is to just catch all exceptions in those places and hope for the best; a few times, we have to rethrow khmer exceptions to propagate them up.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2016-10-13  Camille Scott  <camille.scott.w@gmail.com>
+
+   * Wrap IO functions in with an extra catch to handle the gcc
+     basic_ios_failure bug.
+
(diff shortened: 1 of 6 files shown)


camillescott:
@luizirber @ctb @standage @betatim ready for review and merge, relatively urgent

Issue #1483, opened 2016-10-06
make C++ code that mimics the utils sequence handling code


ctb:
I feel like the code in utils.py introduced  #1435 is headed in the right direction. It'd be nice to have C++ versions of the ReadBundle and broken_paired_iter functions, for speed.

ctb:
#769 is relevant.

betatim:
I'll take a look at this.

betatim:
Regarding doing things "for speed": using `khmer.ReadParser` instead of screed speeds up `broken_pair_reader` by a bit. Next slowest thing is `check_is_pair`.

Not sure there is any benefit in moving `ReadBundle` to c++, at least not for speed.

ctb:
can broken_paired_reader be converted into C++, do you think?

betatim:
You could. What is the motivation? Do you want to use it from other c++ code?

---

If you run `kernprof -l -v scripts/extract-paired-reads.py /tmp/paired-mixed.fq` on the output of
`for i in {1..2000}; do cat tests/test-data/paired-mixed.fq >> /tmp/paired-mixed.fq ;done` you go from 1.3s to 0.5s by switching the read parser. After applying that switch you spend about 44% of your time in `broken_paired_reader` on the for loop, and 35% calling `check_is_pair`.

ctb:
yes, speed.

here's an evaluation file that's a bit bigger for you to play with :)

https://s3.amazonaws.com/public.ged.msu.edu/ecoli_ref-5m.fastq.gz

betatim:
http://gph.is/1TKQFaC

because I liked it so much. I'll take a look.

While poking around the other day I found remnants like https://github.com/dib-lab/khmer/blob/master/lib/read_parsers.hh#L111-L115 and https://github.com/dib-lab/khmer/blob/master/lib/read_parsers.cc#L204-L208 is there any memory on that still around? It looks like it once was the start of something similar to `broken_paired_reader`

ctb:
On Wed, Nov 16, 2016 at 06:31:50AM -0800, Tim Head wrote:
> http://gph.is/1TKQFaC
> 
> because I liked it so much. I'll take a look.
> 
[...]

Issue #1466, opened 2016-09-30
Migrate read handling code to ReadBundler


standage:
In #1458, `ReadBundle` was taken out of trim-low-abund and put into utils so as to be accessible to all scripts and modules. We should:

- try to centralize read handling code into `ReadBundle` as much as possible
- replace *ad hoc* read handling code throughout the codebase with `ReadBundle` objects

ctb:
Here, scripts/filter-abund.py and scripts/filter-abund-single.py would be
good targets for a next round of refactoring.

ctb:
(filter-abund* done in #1469.)

betatim:
You can now use

```python
from khmer import ReadParser

for index, is_pair, read1, read2 in broken_paired_reader(ReadParser(infile)):
  ...
```

instead of using `screed` and benefit from faster reading.

Issue #791, opened 2015-02-20
screed 1.0 API thoughts


ctb:
Things that should be dealt with before screed 1.0 -- please add to this issue.

- [x] Slicing of screed records (#768)
- [x] A plan for synching of quality attribute details between khmer and screed (#769)
- [x] Putting in place tests for and fixes for many broken FASTQ formats (#760, #108)
- [x] len(record) should work (or fail obviously) #705 - screed#18 needs review.
- [x] Make a nicer way to create new screed Record objects (#767)

- [x] Is #625 still relevant or should it be closed?
- [x] Also: remove description, maybe. #820

ctb:
Also: remove description, maybe. #820

mr-c:
- [ ] delete the redefinition of `open` in `screed/openscreed.py`

ctb:
Disagree; it's only a problem if someones does 'from screed import *' which is UnPythonic. I like 'open'.

mr-c:
It proved to be a stumbling block to Camille and I during a `pdb` debugging session. *shrug*

ctb:
All done.

Issue #769, opened 2015-02-11
Synchronize screed and khmer read record details


ctb:
khmer.ReadParser returns read objects that differ from screed-created read objects in silly ways; this should be fixed.

ctb:
One question - is it OK for a legitimate read to have an _empty_ quality attribute?

mr-c:
Note: ReadParser reads still refer to 'accuracy'

ctb:
See also #645

ctb:
Needs to be verified.

mr-c:
ReadParser reads do correctly refer to 'quality' instead of 'accuracy'

Reads from FASTA files still have an 'accuracy' attribute (an empty UTF string).

```
In [6]: o = khmer.ReadParser('tests/test-data/paired.fa')

In [7]: n = o.next()

In [8]: n.
n.annotations  n.name         n.quality      n.sequence  

In [9]: n.quality
Out[9]: u''
```

betatim:
Has the opinion on exposing the `Read` object "fully" changed since #645? Motivation for doing so would be to make it easy to test that `screed.Record` and `khmer.Read` look and feel the same.

ctb:
On Mon, Oct 24, 2016 at 07:02:46AM -0700, Tim Head wrote:
> Has the opinion on exposing the `Read` object "fully" changed since #645? Motivation for doing so would be to make it easy to test that `screed.Record` and `khmer.Read` look and feel the same.

Yes, I think it's a good idea for that reason.

PR #1491, opened 2016-10-24
expose khmer.Read to python and use ReadParser in scripts

### LINKED ISSUE #1098: can't make new ReadParser read object from Python
See the conversation at https://github.com/dib-lab/khmer/issues/645#issuecomment-111739214 and onwards

betatim:
Fix #1098
and related to #1483 

After getting started on this I realised I wasn't sure how we wanted to use this. So far I am assuming the use case is to replace `utils.ReadBundle` as used in for example `trim-low-abund.py`.

Is it really worth it? We will end up turning all the screed `Records` into `read_parser::Read`s via various unicode -> bytes stuff, to then do some computation on them. In the current setup we pass the cleaned sequence to `get_median_count` and friends. Maybe there is something I am missing.
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

ReadBundles group several Reads together.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,13 @@
+2016-11-14  Tim Head  <betatim@gmail.com>
+
+   * khmer/_khmer.cc,lib/read_parsers.hh: Expose khmer's Read type fully to
+   the python side.
+   * scripts/{extract-paired-reads.py,filter-abund-single.py,filter-abund.py,
+   sample-reads-randomly.py,split-paired-reads.py,trim-low-abund.py}: Switch to
+   using ReadParser to speed up scripts.
+   * tests/test_read_handling.py,tests/test_scripts.py: Remove streaming tests
+   that use a FIFO.
+
 2016-11-14  Shannon EK Joslin	<sejoslin@ucdavis.edu>
    *khmer/tests/test_countgraph.py: Refactor test code to use load_countgraph #1474
    *khmer/tests/test_nodegraph.py: Refactor test code to use load_nodegraph and load_countgraph #1474
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -426,7 +426,7 @@ Read_set_cleaned_seq(khmer_Read_Object *obj, PyObject *value, void *closure)
         return 0;
     }
 
-    if (! PyUnicode_Check(value)) {
+    if (! (PyUnicode_Check(value) | PyBytes_Check(value))) {
         PyErr_SetString(PyExc_TypeError,
                         "The 'cleaned_seq' attribute value must be a string");
         return -1;
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -339,6 +339,17 @@ khmer_Read_dealloc(khmer_Read_Object * obj)
 }
 
 
+static Py_ssize_t
+khmer_Read_len(khmer_Read_Object* obj)
+{
+    return obj->read->sequence.size();
+}
+
+static PySequenceMethods khmer_Read_sequence_methods = {
+    (lenfunc)khmer_Read_len,                  /* sq_length */
+};
+
+
 static
 PyObject *
 Read_get_name(khmer_Read_Object * obj, void * closure )
@@ -483,7 +494,7 @@ static PyTypeObject khmer_Read_Type = {
     0,                                    /* tp_compare */
     0,                                    /* tp_repr */
     0,                                    /* tp_as_number */
-    0,                                    /* tp_as_sequence */
+    &khmer_Read_sequence_methods,         /* tp_as_sequence */
     0,                                    /* tp_as_mapping */
     0,                                    /* tp_hash */
     0,                                    /* tp_call */
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -395,6 +395,48 @@ Read_get_annotations(khmer_Read_Object * obj, void * closure)
 }
 
 
+static
+PyObject *
+Read_get_cleaned_seq(khmer_Read_Object * obj, void * closure)
+{
+    if (obj->read->cleaned_seq.size() > 0) {
+        return PyUnicode_FromString(obj->read->cleaned_seq.c_str());
+    } else {
+        PyErr_SetString(PyExc_AttributeError,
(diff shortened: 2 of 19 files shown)


ctb:
Some more background/thoughts/components of this --

* we don't have broken-paired reading functionality in the C++ layer, and the Python version is kinda slow I think; or, at least, screed loading of sequences seems way faster when you don't use broken_paired_reader; benchmarking could be a place to start to see if I'm right :).
* we also need identical read cleaning behavior in C++ and Python levels, so e.g. the clean_input_reads function could be replicated over to or simply moved into the C++ layer;
* the C++ Read / screed.Record functionality should be synchronized (#769)

So it's half yard cleaning and half optimization, I think.

Make sense?

betatim:
Ok, so as a simple minded person I will take that to mean: we want a faster way of doing what `broken_pair_reader` does :)

Read cleaning in C++ vs python: did some exploring and came across [`Hashtable::check_and_normalize_read`](https://github.com/dib-lab/khmer/blob/master/lib/hashtable.cc#L83) which we should keep track of for the consolidation effort.

betatim:
Having re-read what `Hashtable::check_and_normalize_read` does it doesn't improve your read (by upper casing) it just tells you that it isn't a good read.

ctb:
On Mon, Oct 24, 2016 at 07:04:36AM -0700, Tim Head wrote:
> Ok, so as a simple minded person I will take that to mean: we want a faster way of doing what `broken_pair_reader` does :)
> 
> Read cleaning in C++ vs python: did some exploring and came across [`Hashtable::check_and_normalize_read`](https://github.com/dib-lab/khmer/blob/master/lib/hashtable.cc#L83) which we should keep track of for the consolidation effort.

yep, and making this code faster and more sensible is also a +1.

betatim:
Two notes to myself:

* test_streaming_io mostly works, except where it fails because of a missing attribute (expected). 

* `seqan` internally does something different if the input pathname is `-` or a file name. This means testing with a fifo which we pass as a pathname doesn't exercise the right code. In particular for paths it opens the file, reads a few bytes, closes it again, and re-opens later for the real reading. For a FIFO that shouldn't do the right thing.

* the test that is failing looks to be some deadlock issue with opening the fifo in the right order or so.

betatim:
However, replacing `screed.open` with `khmer.ReadParser` does speed up by ~2 the `broken_pair_reader` with a one line change 😀

ctb:
On Wed, Nov 09, 2016 at 04:11:59AM -0800, Tim Head wrote:
> Two notes to myself:
> 
> * test_streaming_io mostly works, except where it fails because of a missing attribute (expected). 

exellent!

> * `seqan` internally does something different if the input pathname is `-` or a file name. This means testing with a fifo which we pass as a pathname doesn't exercise the right code. In particular for paths it opens the file, reads a few bytes, closes it again, and re-opens later for the real reading. For a FIFO that shouldn't do the right thing.
> 
> * the test that is failing looks to be some deadlock issue with opening the fifo in the right order or so.

Do you think the test should be discarded? We have the test_streaming_io tests
which actullay test *live* streaming (which we didn't have at the time of the
FIFO test writing).

betatim:
>> the test that is failing looks to be some deadlock issue with opening the fifo in the right order or so.

> Do you think the test should be discarded? We have the test_streaming_io tests
> which actullay test *live* streaming (which we didn't have at the time of the
> FIFO test writing).

Probably yes. It is super hard to debug because not only is the deadlock happening in another thread, but in some c++ code that is being called by another script being `exec`'ed from another thread ... so if the streaming IO tests cover this/can be made to test this I am 👍 for that. Will check tomorrow morning.

ctb:
Looking good on a quick skim.

One comment - does eliminating the FIFO-based streaming tests drop coverage at all?

betatim:
I removed the stuff that attempted to add the `ReadBundle` type. It was half-baked and the PR ended up focussing on exposing the `Read` type and speeding up a bunch of scripts by using `khmer.ReadParser`.

We can go back to #1483 if we really want to add the `ReadBundle`. Not sure it will do much in terms of speed though. The next slowest thing in `broken_pair_reader` is `check_is_pair` so I'd C++'ify that or speed it up otherwise.

betatim:
No drop in coverage.

```
$ make diff-cover
diff-cover coverage-gcovr.xml coverage.xml
-------------
Diff Coverage
Diff: origin/master...HEAD, staged, and unstaged changes
-------------
khmer/_khmer.cc (100%)
scripts/extract-paired-reads.py (100%)
scripts/filter-abund-single.py (100%)
scripts/filter-abund.py (100%)
scripts/sample-reads-randomly.py (75.0%): Missing lines 154
scripts/split-paired-reads.py (100%)
scripts/trim-low-abund.py (100%)
-------------
Total:   37 lines
Missing: 1 line
Coverage: 97%
-------------
```

ctb:
On Mon, Nov 14, 2016 at 03:05:32AM -0800, Tim Head wrote:
> I removed the stuff that attempted to add the `ReadBundle` type. It was half-baked and the PR ended up focussing on exposing the `Read` type and speeding up a bunch of scripts by using `khmer.ReadParser`.
> 
> We can go back to #1483 if we really want to add the `ReadBundle`. Not sure it will do much in terms of speed though. The next slowest thing in `broken_pair_reader` is `check_is_pair` so I'd C++'ify that or speed it up otherwise.

agreed.

betatim:
This should pass on both legacy python and python any minute now.

betatim:

ctb:
Let's go ahead and merge when the tests pass!

@betatim do you think you could go through the various Read-related issues and update them with this new functionality?

betatim:
> @betatim do you think you could go through the various Read-related issues and update them with this new functionality?

I'll give it a go. Do you have something particular in mind? Otherwise I'll locate read related issues by searching, etc

ctb:
On Mon, Nov 14, 2016 at 05:30:35AM -0800, Tim Head wrote:
> > @betatim do you think you could go through the various Read-related issues and update them with this new functionality?
> 
> I'll give it a go. Do you have something particular in mind? Otherwise I'll locate read related issues by searching, etc

they're all linked issues, if only transitively :)

#1483
#791
#1098
#1466

Issue #1248, opened 2015-08-13
test_load_graph_multithread() in test_scripts.py finds different unique k-mers


ljcohen:
Working on https://github.com/dib-lab/khmer/issues/998, using this command:

> ./setup.py nosetests --tests tests/test_scripts.py:test_load_graph_multithread

default args:

> args = ['-N', '4', '-x', '1e7', '-T', '8', outfile, infile]

(ran ~ 20 x) displayed 4 different unique k-mers in stdout:

> Total number of unique k-mers: 1124051
> Total number of unique k-mers: 1124050
> Total number of unique k-mers: 1124049
> Total number of unique k-mers: 1124052

When change args to -T 1,only one unique k-mer (ran ~20x):

> Total number of unique k-mers: 1124049

OS info:

> ProductName:	Mac OS X
> ProductVersion:	10.9.5
> BuildVersion:	13F34

mr-c:
Right, this sort of thing has been brewing for a while. It *could* be due to the probabilistic nature of the Count-Min Sketch or it could be a threading bug. @luizirber @camillescott may have insights to share

ctb:
On Thu, Aug 13, 2015 at 02:08:58PM -0700, Michael R. Crusoe wrote:
> Right, this sort of thing has been brewing for a while. It *could* be due to the probabilistic nature of the Count-Min Sketch or it could be a threading bug. @luizirber @camillescott may have insights to share

Having dug around in that code recently (and now simplified it), it shouldn't
be happening if the code was working the way I understood it to work.  So
there's almost certainly a bug... Thanks, @ljcohen!

Let's take a look through the issue tracker to find you a new bug
to chase while we figure this out :)

mr-c:
@ctb Is this "known-issues" worthy?

ctb:
On Fri, Sep 04, 2015 at 09:49:16AM -0700, Michael R. Crusoe wrote:
> @ctb Is this "known-issues" worthy?

yep.

camillescott:
This should be resolved by #876 (and serve as extra incentive for me to get that merged)

betatim:
[Another occurrence of a Heisenbug](https://github.com/dib-lab/khmer/pull/1492#issuecomment-257877530) in `test_abundance_dist_threaded` this time.

betatim:
I'm looking into both of these.

For reference:
```
[...]

PR #1492, opened 2016-10-26
Make `Read` accessible from python

### LINKED ISSUE #769: Synchronize screed and khmer read record details

betatim:
Fixes #769 

This makes `khmer.Read` accessible from python land. If you do not set one of the attributes `hasattr` will return `False`.
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,7 @@
+2016-11-04  Tim Head  <betatim@gmail.com>
+
+   * Make khmer's Read type accesible from the python API.
+
 2016-11-03  Titus Brown  <titus@idyll.org>
 
    * lib/{counting.cc,counting.hh,hashbits.cc,hashbits.hh,hashtable.cc,
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,6 @@
 2016-11-04  Tim Head  <betatim@gmail.com>
 
-   * Make khmer's Read type accesible from the python API.
+   * Make khmer's Read type accessible from the python API.
 
 2016-11-03  Titus Brown  <titus@idyll.org>
 
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -299,7 +299,8 @@ khmer_Read_init(khmer_Read_Object *self, PyObject *args, PyObject *kwds)
     const char * quality{};
     char *kwlist[5] = {
         const_cast<char *>("name"), const_cast<char *>("sequence"),
-        const_cast<char *>("quality"), const_cast<char *>("annotations"), NULL};
+        const_cast<char *>("quality"), const_cast<char *>("annotations"), NULL
+    };
 
     if (!PyArg_ParseTupleAndKeywords(args, kwds, "|zzzz", kwlist,
                                      &name, &sequence, &quality, &annotations)) {
@@ -337,8 +338,7 @@ Read_get_name(khmer_Read_Object * obj, void * closure )
 {
     if (obj->read->name.size() > 0) {
         return PyUnicode_FromString(obj->read->name.c_str());
-    }
-    else {
+    } else {
         PyErr_SetString(PyExc_AttributeError,
                         "'Read' object has no attribute 'name'.");
         return NULL;
@@ -352,8 +352,7 @@ Read_get_sequence(khmer_Read_Object * obj, void * closure)
 {
     if (obj->read->sequence.size() > 0) {
         return PyUnicode_FromString(obj->read->sequence.c_str());
-    }
-    else {
+    } else {
         PyErr_SetString(PyExc_AttributeError,
                         "'Read' object has no attribute 'sequence'.");
         return NULL;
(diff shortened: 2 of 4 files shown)


khmer/_khmer.cc
```diff
@@ -308,9 +352,66 @@ static PyGetSetDef khmer_Read_accessors [ ] = {
 };
 
 
+static
+PyObject *
+khmer_Read_getattr(khmer_Read_Object *self, PyObject *attr_name)
```
luizirber: Is this necessary, since we have a `PyGetSetDef` defined?
https://github.com/dib-lab/khmer/pull/1492/files#diff-944b784821ddf1048fc73488ee2ee675R329

betatim: I will try if raising the excepting in the getter works. Good idea!

betatim: You were right :)

betatim:

ChangeLog
```diff
@@ -1,3 +1,7 @@
+2016-10-27  Tim Head  <betatim@gmail.com>
+
+   * Make khmer's Read type accesible from the python API.
```
ctb: accessible - two ses :)

khmer/_khmer.cc
```diff
@@ -238,6 +238,51 @@ typedef struct {
...
+    const char * name{};
+    const char * annotations{};
+    const char * sequence{};
+    const char * quality{};
+    char *kwlist[5] = {
+        const_cast<char *>("name"), const_cast<char *>("sequence"),
+        const_cast<char *>("quality"), const_cast<char *>("annotations"), NULL
+    };
+
+    if (!PyArg_ParseTupleAndKeywords(args, kwds, "|zzzz", kwlist,
```
ctb: Name and sequence shouldn't be optional, I think?

betatim: Agreed. Now we have `khmer.Read(name, seq)` vs `screed.Record(dict(name=name, sequence=sequence))`. Actually you can pass any kind of `dict` to `Record`.

Should we move `Read` to this "pass a dict as only arg" interface? I prefer the named parameters, but if you frequently turn dicts into `Read`s it will get tedious typing `Read(*my_dict)`.

betatim:
@luizirber do you have an opinion on the following:

We have `khmer.Read(name, seq)` vs `screed.Record(dict(name=name, sequence=sequence))`. Does that difference in interface bother you?

Should we move `Read` to this "pass a dict as only arg" interface? I prefer the named parameters, but if you frequently turn dicts into `Reads` it will get tedious typing `Read(*my_dict)`.

ctb:
On Mon, Oct 31, 2016 at 01:16:58AM -0700, Tim Head wrote:
> @luizirber do you have an opinion on the following:
> 
> We have `khmer.Read(name, seq)` vs `screed.Record(dict(name=name, sequence=sequence))`. Does that difference in interface bother you?

screed.Record should be changed to support that IMO.

betatim:
Merge once travis is happy?

ctb:
Agreed.

ctb:
Build failed. Update from latest master, fix build, => merge?

betatim:
This is the error message: https://travis-ci.org/dib-lab/khmer/jobs/171994487#L1841-L1870

betatim:
This failing test is super weird. `make clean && make` -> test fails, edit test case to use `-x 10e7`, rerun tests -> passes, undo edit, test still passes. `make clean && make` and it fails again.

ctb:
could be related to https://github.com/dib-lab/khmer/issues/1248

My suggestion: make it '98 or 99' and commit; we'll deal with thread safety
better later.

--titus

ctb:
I would prefer that the uncertainty be addressed explicitly so that we don't face this in every test going forward ;).

betatim:
Running the test in a loop 40 times also surfaced this failure mode

```
>           assert line == '1,96,96,0.98', line
E           AssertionError: 1,96,96,0.96
E           assert '1,96,96,0.96' == '1,96,96,0.98'
E             - 1,96,96,0.96
E             ?            ^
E             + 1,96,96,0.98
E             ?            ^
```

ctb:
alas... well, perhaps fix it in #1495?

betatim:
I would make a new PR to fix this. At least I have no good idea where to start looking for the race condition so. Empirically this test doesn't seem to happen that often: "a few times in 100 runs"

---

There are indeed 98 kmers in the file:
```python
import  tests.khmer_tst_utils as utils
def kmer(seq, k=17):
  for i in range(0, len(seq)-k+1):
    yield seq[i:i+k]

kmers = []
for rec in screed.open(utils.copy_test_data('test-abund-read-2.fa')):
  kmers += list(kmer(rec.sequence))
```
in case we need this snippet again.

ctb:
On Wed, Nov 02, 2016 at 08:24:26AM -0700, Tim Head wrote:
> I would make a new PR to fix this. At least I have no good idea where to start looking for the race condition so. Empirically this test doesn't seem to happen that often: "a few times in 100 runs"

It's clearly in the 'add' function in Hashtable; that's the only place
the relevant things are incremented.

betatim:
merge?

ctb:
I'm fine with merging...

ctb:
LGTM - if the thread issue (re)surfaces on CI, let's fix it then.

PR #1493, opened 2016-10-28
Fix linear path

standage:
Currently calling `assemble_linear_path` with a seed _k_-mer that is not in the DBG will return the seed _k_-mer (or its reverse complement). I'm not sure if this is a feature or a bug, but I'd argue it's misleading. This PR treats it as a bug and returns an empty string if the seed _k_-mer is not in the DBG.
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
- <strike>[ ] Did it change the command-line interface? Only additions are allowed
  changes were made?
- <strike>[ ] Do the changes respect streaming IO? (Are they


standage:
Ready for review @ctb @luizirber @camillescott @betatim!

Issue #1490, opened 2016-10-20
Irreversible hashing for k>32


betatim:
(This is related to #1441)

What would it take to support `k`>32 without insisting on reversibility? Thats the question.

We need to use a hash function that "mixes" things to avoid `XXXACGT` and `YYYYACGT` giving the same value (for a 8bit hash value). This means we need to somehow detect that we are in "irreversible mode".

We tell `_hash()` what `k` is, so you could make it use a different hash function/scheme for `k>32` that still returns a `uint64`. `_revhash` would either raise an exception or return “something” when called with k>32. Currently `StringToHashIterator` is annoying because it has to know about what `_hash` does in order to do its clever thing with the sliding window. [Here](https://github.com/lemire/rollinghashcpp/blob/master/example3.cpp) this is solved by having a hash function object that allows you to "roll" one character left/right. We could add a similar abstraction so that the knowledge of the hash function isn't inside `StringToHashIterator` but instead where the hash function is defined (that would make me happy). Could be as simple as providing a `roll_hash_left/_right(current_hash_value, remove, add)` function.
#1437 shows how to split things into "needs string" and "only needs hashes" iteration and contains `StringToHashIterator`.

PR #1494, opened 2016-10-30
Make hash function a Hashtable method rather than a global function.

ctb:
This PR moves the _hash and _revhash calls into object-specific methods.  This paves the way to
more radical refactoring of hash functions down the road to support e.g. #1490, ref #1426.

Briefly,
- create new methods hash_dna and unhash_dna on Hashtable objects;
- refactor code in subset.cc and read_aligner.cc appropriately;
- refactor _khmer.cc extension module to be more explicit about hashing.

No slowdown was observed on my hash function benchmarks (using #1436); if anything it
seemed about 5% faster _shrug_.

Note that Hashset still uses the global _hash function, as does KmerIterator; this doesn't
provide a one-stop shopping solution for replacing the global hash function :).
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
Ready for review @luizirber @betatim.

betatim:
Is there a reason for having the various hash methods duplicated in `counting.hh` and `hashbits.hh` instead of putting them in `hashtable.hh`? Can we add a comment in `hashtable.hh` so we don't forget the reason?

I like the new name for `revhash`.

The left over direct calls to `_hash` could be redirected through `Hashtable::hash_dna(kmer)` as well if we make it a class method (`static` in c++ lingo). Would allow us to not have `convert_PyObject_to_HashIntoType` and `ht_convert_PyObject_to_HashIntoType` at the price of having to find a way to compute the k-size for each call. WDYT?

betatim:
Nitpick: if we do keep `convert_PyObject_to_HashIntoType` and `ht_convert_PyObject_to_HashIntoType` can we put them next to each other in `_khmer.cc`. Would make it easier to compare them when you are wondering why there are two apparently equal functions ;)

lib/read_aligner.cc
```diff
@@ -532,12 +532,13 @@ Alignment* ReadAligner::Align(const std::string& read)
     }
 
     HashIntoType fhash = 0, rhash = 0;
-    _hash(start.kmer, k, fhash, rhash);
+    fhash = m_ch->hash_dna_top_strand(start.kmer.c_str());
+    rhash = m_ch->hash_dna_bottom_strand(start.kmer.c_str());
```
betatim: Take a look at http://cpp.sh/5fce6 for something akin to "tuple unpacking" for return values. Might not be strictly necessary here but IMO makes for nice code especially compared to (ab)using arguments for return values 😃

ctb:
Thanks for the comments, @betatim. I moved `convert_PyObject_to_HashIntoType` and `ht_convert_PyObject_to_HashIntoType` next to each other, and pulled `hash_dna` and `unhash_dna` back to `Hashtable`.

I'm punting on removing the last `_hash` calls from _khmer.cc and moving `hash_dna` to be class methods; I don't have a reason to go either way right now, and I think we will be revisiting the hash functions very soon in any case.

ctb:
Actually, in further response to @betatim on the hash functions: in the not-so-distant future when we have multiple Hashtable-derived classes with different hash functions, how would we choose which hash function to use?  This is going to be a more general problem as we mix and match hashtables and hash functions... <sigh>

One solution might be to only get Hashsets from Hashtables, so they swipe the hash functions from there. An alternative might be to have two basic hash functions, one for irreversible and one for reversible, and just clearly distinguish their use in the code; then e.g. Hashset, Nodetable and Counttable would use irreversible, while Nodegraph and Countgraph would use reversible.

Both approaches sound like they'll get confusing... hmm.

Another idea is to have a global list of available hash functions and make sure they're specified explicitly when each object is created.  That could get ugly down the road but it works well with our standard modus operandi of primarily working through Python scripts.

ctb:
Ready for review & merge @betatim

PR #1495, opened 2016-10-30
Extract storage into separate class; split Hashtable into Hashtable and Hashgraph.

ctb:
This PR separates the storage details of hashes from the actual hash function. It should support both more types of storage (4-bit and two byte CountMin sketches? maybe even exact storage?) and multiple hash functions.

It also separates out table and graph operations, making it easier to split class defn's between functionality that needs reversible hashing (graph operations) and functions that don't (table functions). Ultimately this can lead to new CPython objects (maybe Nodetable and Counttable?) that support k > 32 while the Nodegraph and Countgraph CPython objects would remain limited to k <= 32.

 I tried to use templates instead of composition but couldn't get the various friend & inheritance declarations to work.  Suggestions welcome.

No discernable performance impact was observed in some basic benchmarking.

Built on #1494. Might be a way to do #1490 (irreversible hashing for k > 32).
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
@betatim @luizirber I am having linker symbol resolution issues. Any suggestions?

ctb:
Got it to compile!

Comments would be appreciated.  Is it worth following through on this and fixing up the load and save code?

ctb:
All the tests pass!

@betatim @luizirber @standage ready for an initial review!

ctb:
Ready for review! @betatim @standage

betatim:
Mostly good 😄 

One pedantic comment: all the variables that used to be of type `Hashtable` but now are `Hashgraphs` are still called `hashtable` :-/ It'll start getting confusing when `Hashtable`s are used more why the variable called `hashtable` actually contains a graph...

I only spotted one used of `Hashtable` directly, everyone else goes through `Hashgraph`. Makes me wonder if the other `stop_bf` related things should continue to be `Hashtable`s as well. This is where I'll resume the reviewing.

lib/storage.hh
```diff
@@ -0,0 +1,456 @@
...
+ */
+
+class BitStorage : public Storage
+{
+friend class Hashbits;
+protected:
+    std::vector<uint64_t> _tablesizes;
+    size_t _n_tables;
+    uint64_t _occupied_bins;
+    uint64_t _n_unique_kmers;
```
standage: Why duplicate these class members? Isn't the idea of inheritance that the base class stores the common members and functions?

lib/storage.hh
```diff
@@ -0,0 +1,456 @@
...
+ * tracking statistics, as well as save/load, and not much else.
+ *
+ */
+
+class BitStorage : public Storage
+{
+friend class Hashbits;
+protected:
+    std::vector<uint64_t> _tablesizes;
+    size_t _n_tables;
```
standage: Isn't this redundant with `_tablesizes.size()`? Not a big deal, but why risk the two variables getting out of sync?

ctb:
@standage, I'll fix your points on the next PR, which is forthcoming.

@betatim, the inheritance hierarchy will expand and I will do a more thorough review then.

Thanks all :).

ctb:
On Thu, Nov 03, 2016 at 11:22:01AM -0700, Tim Head wrote:
> One pedantic comment: all the variables that used to be of type `Hashtable` but now are `Hashgraphs` are still called `hashtable` :-/ It'll start getting confusing when `Hashtable`s are used more why the variable called `hashtable` actually contains a graph...

this should all have been fixed in #1504.

PR #1504, opened 2016-11-03
More consolidation of Hashtable derived types.

ctb:
Briefly, this PR:

- [x] interposes a new Hashgraph CPython class that inherits from Hashtable and becomes a new base for Countgraph and Nodegraph (in the CPython interface).
- [x] articulates the C++ internals so that they use Hashtable and Hashgraph appropriately (as well as Countgraph etc. where absolutely necessary).
- [x] renames the C++ and CPython API to match: Counttable, Nodetable, Countgraph, Nodegraph.

These changes make a clear distinction between 'tables' and 'graphs' - tables have all of
the basic functionality needed for counting, while graphs support various traversal methods.
This paves the way for:

(a) adding new hash functions, including irreversible ones supporting k > 32 for the *table objects; and
(b) building out a new Counttable CPython object that will support the non-graph operations for k > 32.

(I don't like the 'Counttable' name that much but it fits with Countgraph. I suppose we could do Countstable. But that might engender confusion. Thoughts?)

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
Ready for initial review, y'all. @betatim @luizirber @camillescott @standage.  Might be easier to go commit by commit :(.

Still have to check out my modifications to the abundance dist functions, and I'm not sure if I should add tests for Counttable and Nodetable on this PR, but I think most of it is done.

betatim:
Prefer `Counttable` over `Countstable`. More substantial comments later

ctb:
- [ ] look into using our saner inheritance hierarchy to simplify hashtable extraction in the CPython code.

khmer/_cpy_hashgraph.hh
```diff
@@ -0,0 +1,1606 @@
...
+    0,              /*tp_as_sequence*/
+    0,              /*tp_as_mapping*/
+    0,              /*tp_hash */
+    0,              /*tp_call*/
+    0,              /*tp_str*/
+    0,              /*tp_getattro*/
+    0,              /*tp_setattro*/
+    0,              /*tp_as_buffer*/
+    Py_TPFLAGS_DEFAULT | Py_TPFLAGS_BASETYPE,       /*tp_flags*/
+    "hashgraph object",           /* tp_doc */
```
betatim: Delete all the following lines

khmer/_cpy_hashgraph.hh
```diff
@@ -0,0 +1,1606 @@
...
+    0,                       /* tp_dict */
+    0,                       /* tp_descr_get */
+    0,                       /* tp_descr_set */
+    0,                       /* tp_dictoffset */
+    0,                       /* tp_init */
+    0,                       /* tp_alloc */
+    0,                       /* tp_new */
+};
+
+#define is_hashgraph_obj(v)  (Py_TYPE(v) == &khmer_KHashgraph_Type)
```
betatim: delete me

lib/hashgraph.cc
```diff
@@ -0,0 +1,899 @@
...
+            if (kmer_tagged) {
+                since = 1;
+                if (found_tags) {
+                    found_tags->insert(kmer);
+                }
+            } else {
+                ++since;
+            }
+        }
+#else
```
betatim: Should we keep this around or delete?

lib/hashtable.cc
```diff
@@ -242,327 +241,115 @@ bool Hashtable::median_at_least(const std::string &s,
...
-    Read			  read;
-
-    // TODO? Delete the following assignments.
-    total_reads = 0;
-    n_consumed = 0;
+    KmerIterator kmers(s.c_str(), _ksize);
 
-    // Iterate through the reads and consume their k-mers.
-    while (!parser->is_complete( )) {
+    BoundedCounterType max_count = 0;
```
betatim: If you move this up to be the first definition in the method you (probably) don't have to pay for a copy when returning it.

betatim:
I assume that the substantial bits of code that were moved didn't get modified when you moved them because the tests still pass.

khmer/_cpy_counttable.hh
```diff
@@ -0,0 +1,132 @@
...
+static PyObject* khmer_counttable_new(PyTypeObject * type, PyObject * args,
+                                      PyObject * kwds)
+{
+    khmer_KCounttable_Object * self;
+
+    self = (khmer_KCounttable_Object *)type->tp_alloc(type, 0);
+
+    if (self != NULL) {
+        WordLength k = 0;
+        PyListObject * sizes_list_o = NULL;
```
betatim: Do you know who owns the reference to the list that comes from PyArg_ParseTuple?

betatim: PyArg_ParseTuple doesn't increase the reference count, and presumably the ref count can't decrease to zero while we are using an argument to this function.

khmer/_cpy_counttable.hh
```diff
@@ -0,0 +1,132 @@
...
+
+Contact: khmer-project@idyll.org
+*/
+
+typedef struct {
+    khmer_KHashtable_Object khashtable;
+    Counttable * counttable;
+} khmer_KCounttable_Object;
+
+static PyMethodDef khmer_counttable_methods[] = {
```
ctb: I went both ways on this - it's nice to have it there for if/when we add methods... but yeah, I guess no need.

khmer/_cpy_counttable.hh
ctb: Yep (or at least not with the GIL held).

lib/hashtable.cc
betatim: https://en.wikipedia.org/wiki/Return_value_optimization it probably isn't that exciting in this case but for other/bigger/expensive to copy types it might be more interesting

ctb:
Let's ahead and push the magic merge button when the build ends! (but please don't squash commits :)

camillescott:
Doh, I was too late :(

ctb:
wassup @camillescott?

PR #1505, opened 2016-11-07
Fix two installation quirks

luizirber:
- Quirk 1: `doc/requirements.txt` forced `setuptools==3.4.1`, which is super old and breaks installations inside Anaconda (sometimes)
- Quirk 2: `pip` sometimes fails because it can't find a `easy_install.pth` inside the Anaconda `site-packages` (makes sense, no conda package should be installing it), and apparently the fix is to use the `--ignore-installed` flag. [Source](https://github.com/pypa/pip/issues/2751#issuecomment-165390180)

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2016-11-07  Luiz Irber  <khmer@luizirber.org>
+
+   * Makefile: add --ignore-installed flag to pip install (fix installation
+   on conda environments.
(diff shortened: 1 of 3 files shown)


luizirber:
Ready for review and merge @standage @betatim

PR #1509, opened 2016-11-11
fix documentation build failure on ReadTheDocs from bad autoprogram version

ctb:
Doc builds have been failing for a while (see e.g. [this])(https://readthedocs.org/projects/khmer/builds/4650703/) since we allowed newer version of autoprogram in doc/requirements.txt with

```
sphinxcontrib-autoprogram>=0.1.2
```

but the 0.1.3 release of autoprogram [breaks on python 2.7](https://bitbucket.org/birkenfeld/sphinx-contrib/issues/168/autoprogram-013-fails-on-python-27-due-to).  This PR fixes it by nailing autoprogram down to 0.1.2 which works ([see here](https://khmer.readthedocs.io/en/fix-rtd/)).

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2016-11-11  Titus Brown  <titus@idyll.org>
+
+  * doc/requirements.txt: set autoprogram version requirements to something
+  that works on python 2.7, in order to fix ReadTheDocs builds.
+
 2016-11-10  Camille Scott  <camille.scott.w@gmail.com>
 
   * lib/assembler.{hh,cc}: add JunctionCountAssembler implementation.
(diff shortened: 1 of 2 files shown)


ctb:
Ready for review!

betatim:
LGTM. Restarted the OSX travis build to see if this was just a temporary failure (apparently the build exceeded the time limit)

Is there a way we can automate noticing that the RTD build has started failing? Is it worth it? IIRC travis does build the docs but any failures don't fail the build. This would be easy to change.

ctb:
I receive e-mails about failed RTD builds, so that's all sorted...

PR #1511, opened 2016-11-12
Add hash functions supporting k > 32 to Counttable

ctb:
This follows from #1504 and adds support for irreversible hash functions (supporting k > 32).

Specifically, in this PR we:

* update `Counttable` to use murmurhash and thus supported k > 32;
* introduce a MurmurHashKmerIterator for hashing strings;
* use parameterized tests to test the CPython API for `_Countgraph`, _`Counttable`, `_SmallCountgraph`, `_SmallCounttable` `_Nodegraph`, and `_Nodetable`;
* provide some minimal documentation for the Python API for tables;

Basically this is many small changes that are now reasonably well tested and understood, taking advantage of the refactoring done in #1504.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
w00t!

This code:
```
import khmer

a = khmer.Counttable(5, 1e6, 3)
b = khmer.Countgraph(5, 1e6, 3)

print(a.get_kmer_hashes("ATGGAGGACAAGTAT"))
print(b.get_kmer_hashes("ATGGAGGACAAGTAT"))

print(a.hash('ATGGA'))
print(b.hash('ATGGA'))
```

now yields:
```
[2898244513215582206, 7782325612895303798, 17716873001369615262, 5266100104194697999, 16527671766578686784, 1954197895523434265, 9372126835751090531, 2912821416709855779, 8698352977978348149, 13735319396097939869]
[124, 499, 666, 422, 242, 474, 374, 131, 151, 52, 73]
2898244513215582206
124
```

so we're finally able to swap out hash functions on Hashtable classes!

ctb:
(and reversibility is no longer a requirement on Hashtable derived classes, only on Hashgraph derivatives.)

ctb:
So this now works, too:

```
import khmer

a = khmer.Counttable(33, 1e6, 3)

print(a.get_kmer_counts("ATGGATATGGAGGACAAGTATATGGAGGACAAGTATATGGAGGACAAGTAT"))
print(a.consume("ATGGATATGGAGGACAAGTATATGGAGGACAAGTATATGGAGGACAAGTAT"))
print(a.get_kmer_counts("ATGGATATGGAGGACAAGTATATGGAGGACAAGTATATGGAGGACAAGTAT"))
```

betatim:
What am I missing here:
```python
s = "ATGGATATGGAGGACAAGTATATGGAGGACAAGTATATGGAGGACAAGTAT"
a =khmer.Counttable(33, 1e6, 3)
a.get_kmer_hashes(s[:33]) # -> []
a.get_kmer_hashes(s[:34]) # -> [48584371158645721]
```

ctb:
looks like an off by one error to me (i.e. a bug). will investigate.

lib/hashtable.cc
```diff
@@ -257,72 +260,77 @@ void Hashtable::get_kmers(const std::string &s,
...
     }
+    delete kmers;
 }
 
 
 void Hashtable::get_kmer_counts(const std::string &s,
                                 std::vector<BoundedCounterType> &counts) const
 {
-    KmerIterator kmers(s.c_str(), _ksize);
+    KmerHashIterator * kmers = new_kmer_iterator(s);
```
betatim: I would wrap it in `std::unique_ptr< KmerHashIterator > kmers(new_kmer_iterator(s))` and then get rid of the `delete`.

khmer/_khmer.cc
```diff
@@ -148,29 +148,9 @@ static bool convert_PyObject_to_HashIntoType(PyObject * value,
 {
```
betatim: the comment for this function needs updating (delete last two lines)

ctb:
perfect, thanks!

ctb:
done, much cleaner, thanks!

ctb:
Question for all you foolish mortals: how should we handle testing of CPython class hierarchies?

Do we just test most of the methods on the base types and then test only the derived methods on derived types, and rely on code coverage to tell us when we're missing something?

Or do we massively duplicate tests?

Or what?

betatim:
I'd probably go for option (a).

Another option I like is to devise a set of tests/checks for things that are graph-ish and apply them (via a for loop) to all things that want to be graph-ish (same for table-ish, X-ish). And then you add tests for each type that test what is special about that type.

betatim:
I can't work out why travis is ignoring this. According to the logs it is ignoring this "as per configuration" which isn't super helpful for debugging. One thing that might wake travis up is if you change the target of this PR to `master` (which it should be anyway no?)

standage:
@betatim Consider the following (from our .travis.yml file).

```
branches:
  only: 
    - master
```

ctb:
yes! that is almost certainly the problem. willfix.

ctb:
As this functionality emerges, I'm thinking about adding an option to all of our Countgraph-using code so that people can replace Countgraph with Counttable (swapping out hash functions and allowing k > 32).  This would permit its inclusion in a 2.1 release (no changed default behavior) but also support new use cases. Thoughts?

standage:
@ctb +100 on that suggestion. I can definitely help with the testing side of things to make sure new use cases are well-handled.

standage:
Per our in-person discussion: `-H|--hashing-behavior` with `perfect` (default), `bigK` or `irreversible` and `ss` or `strand-specific` as the initially supported options.

betatim:
Isn't it kind of annoying to have to have `Counttable` and the `KmerIterator` know about which hash function is being used (and duplicate the code)? Can we get away with a generic iterator to which you pass a `std::function` when you construct it from Counttable? Then there is only one place where you need to know about the hash function.

ctb:
On Tue, Nov 15, 2016 at 01:56:06PM -0800, Tim Head wrote:
> Isn't it kind of annoying to have to have `Counttable` and the `KmerIterator` know about which hash function is being used (and duplicate the code)? Can we get away with a generic iterator to which you pass a `std::function` when you construct it from Counttable? Then there is only one place where you need to know about the hash function.

that may be where we end up. but I started down that path and found it
annoying.  I'm trying to feel my way through the code thickets and use
cases as I go.

small moves, ellie. :)

betatim:
Another thing to ponder is: is there a need to allow for a choice of non-reversible hash functions? Not sure I can think of many reasons but maybe there are tradeoffs and so users might want to choose??

ctb:
I think there may well be - for example, we may want to choose a single or
a double stranded irreversible hash function.

betatim:
Had a brief think about why `gcc` is choking on `unique_ptr` but no idea.

betatim:
Compiles again 😀

ctb:
you rock!

betatim:
> Question for all you foolish mortals: how should we handle testing of CPython class hierarchies?

I would create tests for the base type and then apply them also to all derived types via something like
```python
GRAPHISH= [ ... list of graph derived types ... ]

@pytest.mark.parametrize("klass", GRAPHISH)
def test_sth_all_graphs_should_do(klass):
  graph = klass(23, 1e6, 4)
  ...
```

then test the stuff added in the derived classes on the derived classes only.

lib/hashtable.cc
```diff
@@ -516,4 +516,45 @@ const
...
+    }
+
+    unsigned int get_start_pos() const { return index; }
+    unsigned int get_end_pos() const { return index + _ksize; }
+};
+
+unique_ptr<KmerHashIterator> Counttable::new_kmer_iterator(const char * sp) const {
+    KmerHashIterator * ki = new MurmurKmerHashIterator(sp, _ksize);
+    return unique_ptr<KmerHashIterator>(ki);
+}
```
standage: Ok, so `(Node|Count)graph` use 2-bit hash, and `(Node|Count)table` use the murmur hash. Simple enough.

lib/hashtable.hh
```diff
@@ -110,6 +113,17 @@ protected:
     explicit Hashtable(const Hashtable&);
     Hashtable& operator=(const Hashtable&);
 
+    virtual unique_ptr<KmerHashIterator>
```
standage: I wonder if code clarity would benefit from typedef'ing `std::unique_ptr<KmerHashIterator>` to something more concise. The counterargument would be that it's not as explicit, but concise code would be a bigger win IMO.

tests/test_cpython_hierarchy.py
```diff
@@ -62,3 +62,10 @@ def test_nodegraph_vs_table():
 
     assert not hasattr(x, 'consume_and_tag')
     assert hasattr(y, 'consume_and_tag')
+
+
+def test_counttable_no_unhash():
```
standage: I second @betatim's suggestion: `@pytest.mark.parametrize` is the way to go for testing all base and derived types.

ctb:
From discussion with @standage - the main remaining thing to be done (other than addressing comments) is to test the bejeezus out of it.  Test code added in #1551 may be relevant.

lib/kmer_hash.hh
```diff
@@ -364,6 +364,46 @@ public:
     }
 }; // class KmerIterator
 
+//
+// KmerHashIterator - analogous to KmerIterator classes, but returns only
+// only HashIntoType hashes, not full Kmer objects.  This supports irreversible
```
ctb: fixed in 9b214f5

khmer/_khmer.cc
ctb: done.

betatim:
Do you have time to add the tests to this? Otherwise I could make a PR to the PR with some testing.

ctb:
I'd like to - plan to by early next week.  Want to get a chance to think about
the arch.

standage:
I've always relied on k-mers hashing to the same value as their reverse complement. It appears this is not the case in this branch, which may explain some of the weirdness I've seen recently with kevlar. khmer should continue to make this guarantee (seq and revcom hash to same value) for different hash functions, should it not? Unless the user specifically requests a strand-specific hash function.

```python
>>> import khmer
>>> import screed
>>> ct = khmer.Counttable(25, 1e9, 4)
>>> ct.consume('CGTTTTGCCTTGAACGAAAAGAATCATCCTCATCCTCATCCTCATCCTCATCCTGCAATGGCATATAGAGAACTAGAGACATC')
59
>>> ct.consume('TGAATCAGATAAAAATGATCATTCCATATCAGAAGATTCATCCTCATCCTCATCCTCATCCTCATCTTCATCAAGGATATAAA')
59
>>> ct.consume('AGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGATGAGGAT')
59
>>> ct.consume('AAGATGAGGATGAGGATGAGGATGAGGATGAATCTTCTGATATGGAATGATCATTTTTATCTGATTCACAACAGCTTGCAGCT')
59
>>> ct.consume('ATCCCTTTTTAGTACCTCGTTTTGCCTTGAACGAAAAGAATCATCCTCATCCTCATCCTCATCCTCATCCTGCAATGGCATAT')
59
>>> ct.get('CCTCATCCTCATCCTCATCCTCATC')
3
>>> ct.get(screed.dna.reverse_complement('CCTCATCCTCATCCTCATCCTCATC'))
11
>>>
```

ctb:
This is indeed unexpected behavior; my apologies.  #fixme

standage:
Sooo...I swapped out `khmer::_hash_murmur()` for the internals of `MurmurKmerHashIterator`. Now trying to figure out why new test in `tests/test_counttable.py` doesn't work. cc @ctb @luizirber

standage:
Ok, fixed `Hashtable::hash_dna()` function and it's looking good now!

ctb:
@standage see 7a7a51d - I think that addresses this issue for all table types, ne c'est pas?

standage:
Ça me paraît bien.

standage:
Another issue: [reads containing `[^ACGT]`](https://github.com/dib-lab/khmer/issues/1541). The `_hash_murmur()` function is the precise point of failure. It would be easy to extend that function to handle Ns. That doesn't seem like the best way forward with this, but fixing the problem gracefully is going to touch a lot of code, probably require a refactoring of reader/consume code, and in any case is orthogonal to the changes in this PR. Thoughts?

standage:
How about this for a compromise: I fixed the `MurmurKmerHashIterator` to ignore any sequence containing non-ACGT. I did not fix `Counttable::hash`, so it will crash with a C++ exception on any read containing non-ACGT. Perhaps we add notes to the code *in situ* and comments to relevant issues and call it good for this PR?

standage:
Calling `Hashtable::get` with a k-mer string relies on `Hashtable::hash`. This can be avoided by using something like `get_kmer_hashes()` instead of `get_kmers()` to query abundances, but this is a big pain when the k-mer sequence is needed as well as the hash. Sigh...

betatim:
My preference would be to move all the input validation code to as early as possible (the scripts?) and assume that people using the python/C API directly are consenting adults. It feels like we have 101 different places where we (repeatedly) check for lower case, non ACGT, etc.

ctb:
point of order: please add changes like this to a separate PR, so they can be reviewed independently... and also add tests so that we can see some of the intended consequences :).

ctb:
A couple of comments:

* agree with @betatim that the checking-the-valid-DNA code shouldn't be in the CPython interface except for bulk loading functions.
* the current solution violates *both* principles espoused in #1541 ;)
* the other KmerIterator functionality doesn't do this either, so now we're inconsistent;
* "It feels like we have 101 different places where we (repeatedly) check for lower case, non ACGT, etc." is emotive but incorrect. See comment in #1541 ;).
* I'm constantly worried about how to understand and ensure the sequence cleaning that is going on.

At the end of the day I'm wondering why this particular problem needs to be fixed in KmerIterator rather than being addressed in the kevlar code. ??

See soon-to-be-comment in #1541.

standage:
> point of order: please add changes like this to a separate PR, so they can be reviewed independently... and also add tests so that we can see some of the intended consequences :).

Mea culpa, I've done PRs on PRs before. In this case I ended up making more changes than I originally anticipated.

standage:
Sooo...any reason we can't add tests for k > 32 to this PR? Is there still groundwork to be laid for that?

ctb:
On Fri, Jan 27, 2017 at 01:33:46PM -0800, Daniel Standage wrote:
> Sooo...any reason we can't add tests for k > 32 to this PR? Is there still groundwork to be laid for that?

hah, good point!  no, it should Just Work at this point.

ctb:
Tests for k > 32 are already in `tests/test_counttable.py`. Yay!

standage:
Is there anything in particular blocking this PR? Of course the recent problems with kevlar were troubling, but evidence suggests these were problems with linking rather than code. Anything else we need to consider for this PR?

ctb:
That was the last big blocker. I'd like to continue expanding the tests and
the python-api documentation a bit, but it doesn't seem like that's in the
near term cards. Take a look at doc/dev/python-api.rst and see what you think.

ctb:
Added code to fix #1599 (bigcount behavior) in a342b4d95.

ctb:
This is now ready for review! The only thing that I have not checked is code coverage.

ctb:
(Is there a simple way to ask @codecov-io to add coverage summary to the comments? Or should it do so automatically if build passes?)

ctb:
@codecov-io claims that coverage is unchanged: https://codecov.io/gh/dib-lab/khmer/pulls/1511

betatim:
when I run the coverage tools locally `coverage-gcovr.xml` is empty. At some point we stopped generating gcda files when the tests run which is what `gcovr` needs (I think). Trying to find a "good" commit to `git bisect`.

doc/dev/python-api.rst
```diff
@@ -0,0 +1,155 @@
...
+
+Counting types
+--------------
+
+Countgraph:
+
+* get_raw_tables
+* do_subset_partition_with_abundance
+
+Nodegraph:
```
standage: Not sure why this is under the "counting types" heading.

khmer/_khmer.cc
```diff
@@ -2102,7 +1995,12 @@ hashtable_get_kmer_counts(khmer_KHashtable_Object * me, PyObject * args)
     }
 
     std::vector<BoundedCounterType> counts;
-    hashtable->get_kmer_counts(sequence, counts);
+    try {
+        hashtable->get_kmer_counts(sequence, counts);
+    } catch (khmer_exception &e) {
+        PyErr_SetString(PyExc_ValueError, e.what());
+        return NULL;
+    }
```
standage: w00t!

tests/test_tabletype.py
```diff
@@ -0,0 +1,442 @@
+"""
```
standage: Tests are all pretty simple and cover the basic features well.

lib/hashtable.cc
```diff
@@ -509,6 +507,53 @@ const
     return posns;
 }
 
+class MurmurKmerHashIterator : public KmerHashIterator
```
luizirber: shouldn't this be in `lib/kmer_hash.cc` (or `lib/kmer_hash.hh`)?

tests/test_tabletype.py
```diff
@@ -0,0 +1,442 @@
...
+    if tabletype in (_Countgraph, _Counttable):
+        loaded = khmer.load_countgraph(savefile)
+    elif tabletype in (_SmallCountgraph, _SmallCounttable):
+        loaded = khmer.load_countgraph(savefile, small=True)
+    elif tabletype in (_Nodegraph, _Nodetable):
+        loaded = khmer.load_nodegraph(savefile)
+    else:
+        raise Exception("unknown tabletype")
+
+    z = kh.get('ATGGC')
```
standage: Oops, this should be `loaded.get` shouldn't it?

ctb:
On Mon, Feb 13, 2017 at 03:04:36PM -0800, Daniel Standage wrote:
> +* get_raw_tables
> +* do_subset_partition_with_abundance
> +
> +Nodegraph:
> 
> Not sure why this is under the "counting types" heading.

Fixed 8f9e8b0

ctb:
On Mon, Feb 13, 2017 at 07:09:47PM -0800, Luiz Irber wrote:
> luizirber commented on this pull request.
> 
> 
> 
> > @@ -509,6 +507,53 @@ const
>      return posns;
>  }
>  
> +class MurmurKmerHashIterator : public KmerHashIterator
> 
> shouldn't this be in `lib/kmer_hash.cc` (or `lib/kmer_hash.hh`)?

it should never be used publicly, and I think we should actually be
hiding the TwoBit iterator.  -0 on moving it for now.

Presumably as part of Cythonization it will need to be exposed?

tests/test_tabletype.py
ctb: Fixed dc6b287 - thanks!

luizirber:
On February 14, 2017 9:03:56 AM PST, "C. Titus Brown" <notifications@github.com> wrote:
>On Mon, Feb 13, 2017 at 07:09:47PM -0800, Luiz Irber wrote:
>> luizirber commented on this pull request.
>> 
>> 
>> 
>> > @@ -509,6 +507,53 @@ const
>>      return posns;
>>  }
>>  
>> +class MurmurKmerHashIterator : public KmerHashIterator
>> 
>> shouldn't this be in `lib/kmer_hash.cc` (or `lib/kmer_hash.hh`)?
>
>it should never be used publicly, and I think we should actually be
>hiding the TwoBit iterator.  -0 on moving it for now.
>
>Presumably as part of Cythonization it will need to be exposed?
>
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/dib-lab/khmer/pull/1511#issuecomment-279768963

Not about exposing it externally, I would use that for HLL

ctb:
> Not about exposing it externally, I would use that for HLL

oh! cool.  I'd rather get this merged first tho :). is ok?

doc/dev/python-api.rst
ctb: fixed 8f9e8b0d5

luizirber:
On February 14, 2017 9:30:10 AM PST, "C. Titus Brown" <notifications@github.com> wrote:
>> Not about exposing it externally, I would use that for HLL
>
>oh! cool.  I'd rather get this merged first tho :). is ok?
>
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/dib-lab/khmer/pull/1511#issuecomment-279776469

That's the idea =]

ctb:
The Mac OS X build is holding this up. Can we merge, pretty please? :)

Issue #1296, opened 2015-09-07
README.rst & doc/index.rst duplicate a lot of content


mr-c:
Often some piece of information will get updated in one location but not the other. Would be nice if there was less duplication.

Ideas:
- Have sphinx include the contents of README.rst when rendering doc/index.rst
- Have README.rst include not much more than a link to the docs

betatim:
Another PR related to this #1529

betatim:
The README is pretty minimal now, closing.

standage:
I'm doing a bit of cleanup right now :)

PR #1518, opened 2016-11-15
Clean up README

betatim:
A proposal for #1296 with opinions.

Basically it is all about reducing the amount of text. Don't know what it is about the dev instructions but I found them confusing (given I was trying to get a quick gist to then translate to my `conda` world).

I would also reduce `doc/dev/getting-started.rst`:
* fork the repo on github, clone locally, setup dib remote. Instead of expanding on this link to the github docs for details. I think today most people either know their way around github or are super new in which case they will need a human to help them,
* virtualenvs only,
* one section per OS instead of by tool,
* remove instructions for more obscure tools (ccache, cppformat) instead explain how to find the output of travis that runs this for you

My ideal would be to get it down to in terms of steps and verbosity of instructions:
```
conda create -n khmer python=2 # or 3
source activate khmer
git clone https://github.com/dib-lab/khmer
cd khmer
make install-dependencies
make
```

Please provide some counter-balance to this viewpoint which is based on spending every day on github etc.

- [x] Is it mergeable?
  changes were made?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -62,12 +62,13 @@ them, please visit the following URLs:
 
 -  Announcements: http://lists.idyll.org/listinfo/khmer-announce
 
-We chat at https://gitter.im/dib-lab/khmer and the maintainers can be
-contacted at khmer-project@idyll.org.
+We have a slack chat that [you can join](LINK/EMAIL/JOIN) and the maintainers
+can be contacted at khmer-project@idyll.org.
 
 For getting help please see this guide: http://khmer.readthedocs.io/en/stable/user/getting-help.html
 
-IMPORTANT NOTE: CITE US!
+
+Important note: cite us!
 ------------------------
 
 khmer is *research software*, so you should cite us when you use it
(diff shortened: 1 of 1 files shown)


betatim:
I thought #1505 solved the "Cannot remove entries from nonexistent file /lib/python2.7/site-packages/easy-install.pth" problem when typing `make install-dep` but it just happened to me again.

betatim:
Running `pip install --ignore-installed setuptools` and then `make install-dep` seems to fix it. Any ideas @luizirber or @standage ?

README.rst
```diff
@@ -62,38 +62,47 @@ them, please visit the following URLs:
 
 -  Announcements: http://lists.idyll.org/listinfo/khmer-announce
 
-We chat at https://gitter.im/dib-lab/khmer and the maintainers can be
-contacted at khmer-project@idyll.org.
+We have a slack chat that [you can join](LINK/EMAIL/JOIN) and the maintainers
```
standage: If I understand correctly, the slack chat is for "internal" lab communications. Of course we're pretty lax about how we define "internal" -- collaborators, colleagues, friends all welcome -- but giving access to the khmer channel requires membership in the dib-lab slack group, which I don't think we want to advertise as public.

standage: We could  make a new Slack group dedicated to khmer outreach, or we could simply point folks to our issue tracker.

ctb:
I think we should point folks at our issue tracker...

README.rst
ctb: I would prefer not to proliferate slack membership beyond people I know.
However, we could pay for slack and then set up a channel to whom we can invite
specific people.  But unfortunately you need an invitation to be sent out.
There are ways around that but it's all just very annoying.

I think we should

* primarily point people at khmer issue tracker
* advertise khmer@ mailing list

standage: Yeah, see Titus' comment. Issue tracker for support, with a brief mention of the mailing list.

betatim:
No more slacking off. Issues are mentioned at the bottom of the `getting-help.rst` that is linked. You want it more prominently? As long as the volume is low I think issues are the way to go.

standage:
Yeah, I would say we should re-word the "getting help" docs to point at the issue tracker as the primary point of contact.

ctb:
@betatim, the only disagreement I have is here:

"fork the repo on github, clone locally, setup dib remote. Instead of expanding on this link to the github docs for details. I think today most people either know their way around github or are super new in which case they will need a human to help them,"

We ran a few different contributathons and found that the instructions were actually pretty good for novices (see [our WSSSPE 2014 paper](https://github.com/dib-lab/2014-paper-wssspe14/blob/master/2014-wssspe14.pdf) for a discussion). We are still asking rotation students in the lab to go through them.  So that's exactly our benchmark - novices :). And no, most scientists do not know their way around github... but in our experience they don't all need a human to help them either!

betatim:
Ok :) Those remarks were meant for the dev guide and thats now being taken care of in #1529 so from my side there is nothing more to add to this PR

PR #1526, opened 2016-11-15
More reasonable change log format

### LINKED ISSUE #1457: ChangeLog thoughts - why do we have it?
Discussion from Sep 26 hangout --
- most people don't use or like the ChangeLog;
- CTB uses the ChangeLog to track down files that changed and may have broken something, but now that he knows how to use git bisect he doesn't do that any more;
- LI points out that it's the Debian format and may be useful for Debian releases;
- ChangeLog is useful for API breakage and tracking, but we don't have any.

Other issues:
- the current ChangeLog format breaks every merge without special tooling;
- the format itself is not nec standard (at least in CTB's emacs)

Should we ditch the ChangeLog?

Should we tool up and make it easy to regenerate the ChangeLog from the git commits?

Is there a way to autogenerate it from GitHub PRs for each release?

standage:
This PR moves the existing ChangeLog into a legacy directory and replaces it with a new CHANGELOG.md file that is much more reasonable for what we need. It will be much less verbose, and include only substantial changes to khmer. Anything relevant to semantic versioning MUST be included, but under-the-hood changes can also be mentioned if they are suitably large and interesting.

We should still include this in the PR checklist, but maybe replace the message with something like "Are substantial changes documented in CHANGELOG.md?"

My prediction: merge conflicts will drop by > 75%.

Fixes #1457.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)


ctb:
looks like you're including another PR's branch in this one - suggest either setting the base to that PR, or ..., well, doing that. it's nice to see the diffs that are specific to this PR.

standage:
Yeah, just saw that. Unintentional. :(

CHANGELOG.md
```diff
@@ -0,0 +1,22 @@
+# Change Log
+All notable changes to the khmer project will be documented in this file.
+See [this](http://keepachangelog.com/) for more info.
```
ctb: [keepachangelog](...) so people don't have to hover or click to see

standage:
Closing due to inclusion of an extraneous commit. Will open a new PR momentarily.

PR #1527, opened 2016-11-15
More reasonable change log format

### LINKED ISSUE #1457: ChangeLog thoughts - why do we have it?

standage:
This PR moves the existing ChangeLog into a legacy directory and replaces it with a new CHANGELOG.md file that is much more reasonable for what we need. It will be much less verbose, and include only substantial changes to khmer. Anything relevant to semantic versioning MUST be included, but under-the-hood changes can also be mentioned if they are suitably large and interesting.

We should still include this in the PR checklist, but maybe replace the message with something like "Are substantial changes documented in CHANGELOG.md?"

My prediction: merge conflicts will drop by > 75%.

Fixes #1457.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
<strike>- [ ] Did it change the command-line interface? Only additions are allowed
  changes were made?
<strike>- [ ] Do the changes respect streaming IO? (Are they

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -4,11 +4,12 @@
   `scripts/` is it tested?
 - [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
   formatted?
-- [ ] Did it change the command-line interface? Only additions are allowed
-  without a major version increment. Changing file formats also requires a
-  major version number increment.
-- [ ] Is it documented in the `ChangeLog`?
-  http://en.wikipedia.org/wiki/Changelog#Format
+- [ ] Did it change the command-line interface? Only backwards-compatible
+  additions are allowed without a major version increment. Changing file
+  formats also requires a major version number increment.
+- [ ] For substantial changes or changes to the command-line interface, is it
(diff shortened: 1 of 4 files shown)


standage:
The latest commit fills out the new CHANGELOG a bit. It doesn't include an entry for every single commit or PR, a lot of refactoring and test-facing PRs were excluded. With each release we'll change the [Unreleased] header to the version number (2.1 coming up soon), and provide an even more curated summary in release notes (which IMO don't need to be nearly as verbose as with previous versions).

standage:
Done.

betatim:
I know everyone loved crafting those detailed change log entries, from now on it'll be more terse.

PR #1530, opened 2016-11-16
Fix streaming gzip output

### LINKED ISSUE #1363: `trim-low-abund.py ... --gzip -o -` fails on Python 3
Witness:

```
% scripts/trim-low-abund.py data/100k-filtered.fa --gzip -o - | gunzip -c > /dev/null
...
created temporary directory ./tmp_6iiq1ovkhmer; use -T to change location
  File "scripts/trim-low-abund.py", line 530, in <module>
    main()
  File "scripts/trim-low-abund.py", line 403, in main
    trimfp = get_file_writer(outfp, args.gzip, args.bzip)
  File "/Users/t/dev/khmer/khmer/kfile.py", line 236, in get_file_writer
    ofile = gzip.GzipFile(fileobj=file_handle, mode='w')
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/gzip.py", line 192, in __init__
    self._write_gzip_header()
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/gzip.py", line 220, in _write_gzip_header
    self.fileobj.write(b'\037\213')             # magic header
TypeError: write() argument must be str, not bytes
```

betatim:
Fixes #1363 

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ under semantic versioning, but will be in future versions of khmer.
   Hashbits --> Nodegraph.
 
 ### Fixed
+- Bug in compressed(gzip) streaming output from scripts
 - The hashbits `update_from` function to correctly track occupied bins for
   calculating FPR.
 - Bug in the `filter-abund.py` script when `--gzip` and `-o` flags are used
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -94,6 +94,18 @@ class ComboFormatter(argparse.ArgumentDefaultsHelpFormatter,
     pass
 
 
+# XXX Temporary fix to argparse's FileType which ignores the
+# binary mode flag. Upstream bug tracked in https://bugs.python.org/issue14156
+class FileType(argparse.FileType):
+    def __call__(self, fname):
+        if fname == '-' and sys.version_info.major == 3:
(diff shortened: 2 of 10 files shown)


betatim:
(`<_io.StringIO object at 0x105497168>` (test) vs `<_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>` (shell))

betatim:
There are some disagreements with `make format` about how to format python code and how much stuff I should add to the temporary hack class, but otherwise:

nb. at some point we should investigate how to modify the unit-tests that monkey about with std{in,out} so that we don't need to special case them in the code. Feels dirty.

standage:
The OS X build on the last commit timed out. Just restarted.

standage:
Actually, cancelling since we need to merge in master anyway.

betatim:
🔄🎸 rebased.

(I think the OSX failure was the heisenbug from earlier related to thread safety (or lack thereof). Bumped into this in #1492, related #1248 )

betatim:
Changed the imports to `from foo import FileType as khFileType` to make it more obvious that we are using our own `FileType` on purpose

PR #1539, opened 2016-11-23
Add info flag to command line

### LINKED ISSUE #1395: Refactor citation checking code
The function `check_version_and_basic_citation(...)` in `tests/test_scripts.py` looks at the output of running the scripts with "--version" to see if the `info(...)` command is present and reporting at least some required citations.  However, when argument processing needs to happen before 'info' as is the case for scripts with `--quiet` enabled, this doesn't work.  How can we fix this generically?

One option is to decorate the argument parser with the citation information & add an `--info` flag that produces it and then exits. This seems like the right way to me...

See #1380 for the motivational PR.

betatim:
Fixes #1395

This introduces the `--info` command line flag. `scripts/filter-abund-single.py --info` now prints the citation information and then exits. This still prints the citation information when you run the script. This is a bit ugly because you have to specify what to cite twice (DRY). We can fix that by either getting rid of printing the citation info on every run or by moving things around. WDYT?

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

New flag that prints citation information

### DIFF
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -77,6 +77,16 @@ ALGORITHMS = {
 }
 
 
+class CitationAction(argparse.Action):
+    def __init__(self, *args, **kwargs):
+        self.citations = kwargs.pop('citations')
+        super(CitationAction, self).__init__(*args, nargs=0, **kwargs)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        info(parser.prog, self.citations)
+        parser.exit()
+
+
 class _VersionStdErrAction(_VersionAction):
     # pylint: disable=too-few-public-methods, protected-access
     """Force output to StdErr."""
@@ -372,7 +382,7 @@ def _check_fp_rate(args, desired_max_fp):
     return args
 
 
-def build_graph_args(descr=None, epilog=None, parser=None):
+def build_graph_args(descr=None, epilog=None, parser=None, citations=None):
     """Build an ArgumentParser with args for bloom filter based scripts."""
     if parser is None:
         parser = argparse.ArgumentParser(description=descr, epilog=epilog,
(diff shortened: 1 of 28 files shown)


ctb:
IMO citation info should be printed unless -q is specified.

betatim:
Have a look how you like this. Should print `info()` for everything except `--help` and adds a `--info` flag. Also removes a lot of the duplicated `--version` definitions.

betatim:
Instead of subclassing we could create a `build_basic_args()`, I'm +-0 on either

standage:
@betatim If we're going to print info by default and suppress only when `-q|--quiet` is selected, perhaps it makes sense to add `-q|--quiet` as a standard flag to all scripts (via the base `KhmerArgumentParser` class) and remove the per-script definitions.

And I think we want to print info even when `--help` is selected.

betatim:
Not sure about adding `-q` to everything via `KhmerArgParse`. The only thing it would do for some scripts is disable the citation info, but it wouldn't actually make the script anymore quiet. Should we make a new issue/PR that adds real quietness to every script and adds `-q` to `KhmerArgParse`?

`-h` now prints citation info.

ctb:
The only reason we haven't added -q automatically is that adding -q properly to
all the scripts is annoying and time consuming and it didn't seem time
critical.

betatim:
So I propose we don't add `-q` to the new base class and also don't make a issue to add it in the future :)

khmer/khmer_args.py
```diff
@@ -77,11 +77,34 @@
 }
 
 
+class CitationAction(argparse.Action):
```
standage: Nice, clean and concise implementation.

khmer/khmer_args.py
```diff
@@ -98,6 +121,36 @@ class ComboFormatter(argparse.ArgumentDefaultsHelpFormatter,
     pass
 
 
+class KhmerArgumentParser(argparse.ArgumentParser):
```
standage: 👍

betatim:
Merci!

Issue #1548, opened 2016-12-01
load-graph seems to be computing memory sizes incorrectly


ctb:
```
load-graph.py -M 24e9 -k 21 NB16-combined.abundtrim.fq.gz.cg NB16-combined.abundtrim.fq.gz

|| This is the script load-graph.py in khmer.
|| You are running khmer version 2.0+569.g1388f8b
|| You are also using screed version 0.9
||
|| If you use this script in a publication, please cite EACH of the following:
||
||   * MR Crusoe et al., 2015. http://dx.doi.org/10.12688/f1000research.6924.1
||   * J Pell et al., http://dx.doi.org/10.1073/pnas.1121464109
||   * A. D\xf6ring et al. http://dx.doi.org:80/10.1186/1471-2105-9-11
||
|| Please see http://khmer.readthedocs.io/en/latest/citations.html for details.


PARAMETERS:
 - kmer size =    21            (-k)
 - n tables =     4             (-N)
 - max tablesize = 4.8e+10      (-x)

Estimated memory usage is 2.4e+10 bytes (n_tables x max_tablesize / 8)
--------
ERROR: Not enough free space on disk for saved graph files;       Need at least 5.6 GB more.
       Table size: 48.0 GB
       Free space: 42.4 GB
NOTE: This can be overridden using the --force argument

```

PR #1549, opened 2016-12-01
Fix incorrectly computed nodegraph size

standage:
#1548 shows that graph sizes are computed incorrectly, at least for nodegraphs. This PR introduces a regression test for the bug. No solution has yet been implemented.

- [x] Is it mergeable?
<strike>- [ ] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?</strike>
  formatted?
<strike>- [ ] Did it change the command-line interface? Only backwards-compatible
  formats also requires a major version number increment.</strike>
  for more details.
  changes were made?
<strike>- [ ] Do the changes respect streaming IO? (Are they


betatim:
Tried to dig into why this happens and ended up with the following question: `tablesize` is meant to be the number of buckets in the BF or the amount of memory used?

`BitStorage` interprets the table sizes it is passed as number of buckets and correspondingly will use `table_size/8` bytes of memory (per table). For `NibbleStorage` table sizes passed in are also number of buckets -> `table_size/2` bytes per table. For `ByteStorage` one bucket is one byte.

`calculate_graphsize` converts from `args.max_memory_usage` (in bytes) to a number of buckets per table. In `report_on_config` we go the other way again and report max memory use.

Then there is `check_space_for_graph` which gets called with `tablesize` as argument. I think it should be a number of bytes though?

After all this I think the test in this PR shouldn't pass.

standage:
Right-o, @betatim. I updated the test, which passes now. On closer look, I think the confusion is with `kfile.check_space_for_graph`, which is treating `hash_size` as total required memory in bytes. The `load_graph.py` code is handing it the max table size in buckets of *one* of the tables composing the final graph.

I think it makes for `kfile.check_space_for_graph` to work at the level of "bytes required for the whole graph", but there's at least one place where we need fix the code to make sure this is what is passed.

oxli/build_graph.py
```diff
@@ -80,6 +80,7 @@ def main(args):
         check_input_files(fname, args.force)
 
     graphsize = calculate_graphsize(args, 'nodegraph')
+    space_needed = args.n_tables * graphsize / 8
```
standage: This was the key to fixing #1548.

khmer/kfile.py
```diff
@@ -162,18 +167,19 @@ def check_space_for_graph(outfile_name, hash_size, force,
...
+        message = 'Not enough free space on disk for saved graph files;'
+        message += '\n       Need at least {:s} more.'.format(mem_needed)
+        message += '\n       Table size: {:s}'.format(mem_requested)
+        message += '\n       Free space: {:s}'.format(mem_available)
+        if force:
+            print('WARNING:', message, file=sys.stderr)
+        else:
+            message = 'ERROR: ' + message + \
+                '\nNOTE: This can be overridden using the --force argument'
+            raise SystemExit(message)
```
standage: The `check_space` function could use similar cleanup.

tests/test_scripts.py
```diff
@@ -689,6 +689,21 @@ def test_oxli_build_graph_fail():
     assert "** ERROR: the graph structure is too small" in err
 
 
+@pytest.mark.known_failing
+def test_oxli_build_graph_yuge():
+    script = 'oxli'
+    args = ['build-graph', '-M', '800T', '-k', '20']
```
standage: Like all of the rest of the `load-graph.py` tests, this is marked "known failing" since loading the oxli library for pytest seems like an unsolved problem.

khmer/kfile.py
```diff
@@ -151,7 +151,12 @@ def check_space(in_files, force, _testhook_free_space=None):
...
                           _testhook_free_space=None):
-    """Check that we have enough size to write the specified graph."""
+    """
+    Check that we have enough size to write the specified graph.
+
+    The "hash_size" parameter should equal the total bytes required for the
+    entire data structure.
+    """
     dir_path = os.path.dirname(os.path.realpath(outfile_name))
     target = os.statvfs(dir_path)
```
standage: By the way: `statvfs` is deprecated. Not urgent for this PR, but something to be aware of. https://docs.python.org/2/library/statvfs.html

betatim: I think `statvfs` (the module) != `os.statvfs` (the function call). So as long as we don't use the module we should be good.

oxli/build_graph.py
betatim: `space_needed` is never used? :mag:

tests/test_script_arguments.py
```diff
@@ -71,22 +72,54 @@ def test_check_space():
...
-    outfile = utils.get_test_data('truncated.fq')
-    save_stderr, sys.stderr = sys.stderr, io.StringIO()
+@pytest.mark.parametrize('graph_type,buckets_per_byte', [
+    ('countgraph', 1),
+    ('smallcountgraph', 2),
+    ('nodegraph', 8),
+])
+def test_check_tablespace(graph_type, buckets_per_byte):
+    oldstderr = sys.stderr
+    sys.stderr = StringIO()
```
betatim: There is a `capfd` fixture in pytest that takes care of capturing stdout/err.

```
capfd -- ..
    Enable capturing of writes to file descriptors 1 and 2 and make
    captured output available via ``capfd.readouterr()`` method calls
    which return a ``(out, err)`` tuple.
```

potentially less brittle than our homebaked solution. Not quite sure what the difference is to the `capsys` fixture.

betatim: http://docs.pytest.org/en/latest/capture.html?highlight=capfd#accessing-captured-output-from-a-test-function (more reading)

khmer/kfile.py
ctb: ok to punt to issue if it's not done in this PR! label it low hanging fruit...

tests/test_script_arguments.py
ctb: punt to an issue, pls.

tests/test_scripts.py
ctb: what - really? let's punt that to an issue too :)

oxli/build_graph.py
ctb: Agreed. what? :)

tests/test_script_arguments.py
betatim: See #1579

tests/test_scripts.py
betatim: For me `oxli` isn't on the PATH, so this isn't just pytest related.

oxli/build_graph.py
standage: Oops, there were some changes that were accidentally left out of the last commit. Buuuuuuut...there are still some other issues to fix!

standage: Fixed now.

khmer/kfile.py
standage: See #1585.

tests/test_scripts.py
standage: See #1586.

standage:
Ok, should be ready for review and merge! @betatim @ctb

khmer/__init__.py
```diff
@@ -86,6 +86,13 @@
 del get_versions
 
 
+buckets_per_byte = {
```
betatim: `_buckets_per_byte`? Just to prevent people from being tempted to modify values here.

oxli/build_graph.py
```diff
@@ -80,7 +80,8 @@ def main(args):
         check_input_files(fname, args.force)
 
     graphsize = calculate_graphsize(args, 'nodegraph')
-    check_space_for_graph(args.output_filename, graphsize, args.force)
+    space_needed = args.n_tables * graphsize / 8
```
betatim: Could use `khmer.buckets_per_byte['nodegraph']`

Issue #1035, opened 2015-05-27
Implement 8-bit minifloat and/or 4-bit counters for CountingHash objects.


ctb:
See http://en.wikipedia.org/wiki/Minifloat; also see #872.

luizirber:
I've started working on a minifloat implementation, based on the `half` float implementation from [openexr][1] (and exploring C++11, [unit testing][2] and [dependency management][3]). 
Repo: http://github.com/luizirber/minifloat

[1]: https://github.com/openexr/openexr/blob/91015147e5a6a1914bcb16b12886aede9e1ed065/IlmBase/Half/half.h
[2]: https://github.com/philsquared/Catch
[3]: https://biicode.com
[1]: https://github.com/openexr/openexr/blob/91015147e5a6a1914bcb16b12886aede9e1ed065/IlmBase/Half/half.h

ctb:
hey @luizirber what is the prognosis for adding minifloat into khmer?

ctb:
another question, @luizirber - do you know how to increment minifloats?

standage:
4-bit counter now available as "Nibble" storage class.

PR #1551, opened 2016-12-02
Nibble storage class

betatim:
This introduces a storage class that uses 4bit per counter.

The storage class is `NibbleStorage` and it is used in `SmallCounttable`.

Todo:
- [x] save/load
- [ ] multi threading?!
- [x] big count and/or overflow?
- [x] `SmallCountgraph` ?
- [x] check conversion of size arguments

Checklist:
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -479,12 +479,15 @@ def add_loadgraph_args(parser):
 
 def calculate_graphsize(args, graphtype, multiplier=1.0):
     """Transform the table parameters into a size."""
-    if graphtype not in ('countgraph', 'nodegraph'):
+    if graphtype not in ('countgraph', 'smallcountgraph', 'nodegraph'):
         raise ValueError("unknown graph type: %s" % (graphtype,))
 
     if args.max_memory_usage:
         if graphtype == 'countgraph':
             tablesize = args.max_memory_usage / args.n_tables / \
+                float(multiplier)
+        elif graphtype == 'smallcountgraph':
+            tablesize = 2 * args.max_memory_usage / args.n_tables / \
                 float(multiplier)
         elif graphtype == 'nodegraph':
             tablesize = 8. * args.max_memory_usage / args.n_tables / \
@@ -543,7 +546,10 @@ def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1):
         sys.exit(1)
 
     tablesize = calculate_graphsize(args, 'countgraph', multiplier=multiplier)
-    return khmer.Countgraph(ksize, tablesize, args.n_tables)
+    if args.small_count:
+        return khmer.SmallCountgraph(ksize, tablesize, args.n_tables)
+    else:
+        return khmer.Countgraph(ksize, tablesize, args.n_tables)
 
 
 def report_on_config(args, graphtype='countgraph'):
@@ -553,7 +559,7 @@ def report_on_config(args, graphtype='countgraph'):
     made available by this module.
     """
     check_conflicting_args(args, graphtype)
-    if graphtype not in ('countgraph', 'nodegraph'):
+    if graphtype not in ('countgraph', 'smallcountgraph', 'nodegraph'):
         raise ValueError("unknown graph type: %s" % (graphtype,))
 
     tablesize = calculate_graphsize(args, graphtype)
@@ -568,6 +574,11 @@ def report_on_config(args, graphtype='countgraph'):
             "Estimated memory usage is {0:.2g} bytes "
             "(n_tables x max_tablesize)".format(
                 args.n_tables * tablesize))
+    elif graphtype == 'smallcountgraph':
+        log_info(
+            "Estimated memory usage is {0:.2g} bytes "
+            "(n_tables x max_tablesize / 2)".format(
+                args.n_tables * tablesize / 2))
     elif graphtype == 'nodegraph':
         log_info(
             "Estimated memory usage is {0:.2g} bytes "
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -460,6 +460,10 @@ def build_counting_args(descr=None, epilog=None, citations=None):
     """Build an ArgumentParser with args for countgraph based scripts."""
     parser = build_graph_args(descr=descr, epilog=epilog, citations=citations)
 
+    parser.add_argument('--small-count', default=False, action='store_true',
+                        help='Reduce memory usage by using a smaller counter'
+                        ' for individual kmers.')
+
     return parser
 
 
@@ -545,10 +549,13 @@ def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1):
         print_error("\n** ERROR: khmer only supports k-mer sizes <= 32.\n")
         sys.exit(1)
 
-    tablesize = calculate_graphsize(args, 'countgraph', multiplier=multiplier)
     if args.small_count:
+        tablesize = calculate_graphsize(args, 'smallcountgraph',
+                                        multiplier=multiplier)
         return khmer.SmallCountgraph(ksize, tablesize, args.n_tables)
     else:
+        tablesize = calculate_graphsize(args, 'countgraph',
+                                        multiplier=multiplier)
         return khmer.Countgraph(ksize, tablesize, args.n_tables)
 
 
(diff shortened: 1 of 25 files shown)


betatim:
Addresses part of #1035

ctb:
Looks good on a first skim -- is this something that should be merged into #1511, do you think, or is it best => master? If into #1511 it would be able to take advantage of k > 32.

betatim:
I'm easy either way. Merge this into master and then rebase #1511 or merge this into #1511 and then later into master. I'd pick the option where it gets used sooner.

betatim:
Now the counter doesn't overflow anymore but stops counting up at 15. This seems much more sensible than overflowing. It means you can run
```
$ trim-low-abund.py -k 32 -M 1e6 -C 3 -V -Z 15 ../zaok/data/mixed-reads.fa.gz -o temp/mixed-trimV-2.fa
```
using the same amount of memory but with bigger tables (-> smaller fp) or less memory (at constant fp):
```
$ trim-low-abund.py -k 32 -M 1e6 -C 3 -V -Z 15 ../zaok/data/mixed-reads.fa.gz -o temp/mixed-trimV-2.fa
```

lib/storage.hh
```diff
@@ -237,6 +237,7 @@ public:
 class NibbleStorage : public Storage
 {
 protected:
+    // table size is measured in number of entries in the table, not in bytes
```
standage: I was just gonna ask about this. Thanks for clarifying!

betatim:
@standage do you have an opinion on where to merge this to? I assume you'd be one of the keen first users.

standage:
@betatim I'd say directly to master. Definitely related to #1511, but orthogonal enough to warrant separate consideration.

betatim:
@standage do you have time for a review?

There is one outstanding question: how to do this when the code is running on multiple threads. The simple answer is locks. So that is still "todo" but otherwise I think ready for a second pair of eyes.

tests/test_countgraph.py
```diff
@@ -218,6 +229,18 @@ def test_get_raw_tables_view():
...
+def test_get_raw_tables_view_smallcountgraph():
+    ht = khmer.SmallCountgraph(20, 1e5, 4)
+    tables = ht.get_raw_tables()
+    for tab in tables:
+        assert sum(tab.tolist()) == 0
+    ht.consume('AAAATTTTCCCCGGGGAAAA')
+    # the actual count is 1 but stored in the first 4bits of a Byte
+    # and so becomes 16
+    for tab in tables:
+        assert sum(tab.tolist()) == int('00010000', 2)
```
standage: 👍

tests/test_countgraph.py
```diff
@@ -634,6 +657,27 @@ def do_test(ctfile):
...
+
+        loaded = khmer.load_countgraph(savepath, small=True)
+
+        orig_count = orig.n_occupied()
+        loaded_count = loaded.n_occupied()
+        assert orig_count == 3881, orig_count
+        assert loaded_count == orig_count, loaded_count
+
+    for ctfile in ['temp.ct', 'temp.ct.gz']:
+        do_test(ctfile)
```
standage: @ctb Isn't a fan of loops in tests, although I must admit I'm not sure why. Perhaps it's just a matter of keeping them simple? I'm ok with this, but I'll let him comment on it.

tests/test_nibblestorage.py
```diff
@@ -0,0 +1,129 @@
...
+    sct = SmallCounttable(4, 1e6, 4)
+    a = "AAAA"
+    b = "AAAT"
+
+    # try to overflow our 4bit counter
+    for n in range(17):
+        sct.add(a)
+
+    assert sct.get(a) == 15
+    assert sct.get(b) == 0
```
standage: 👍

standage:
Shall I hold off with merging until multithreading support is there?

ctb:
On Wed, Dec 07, 2016 at 11:09:23AM -0800, Daniel Standage wrote:
> Shall I hold off with merging until multithreading support is there?

let me take a look.

ctb:
On Wed, Dec 07, 2016 at 11:07:51AM -0800, Daniel Standage wrote:
> +    for ctfile in ['temp.ct', 'temp.ct.gz']:
> +        do_test(ctfile)
> 
> @ctb Isn't a fan of loops in tests, although I must admit I'm not sure why. Perhaps it's just a matter of keeping them simple? I'm ok with this, but I'll let him comment on it.

Exactly - we shouldn't have to debug our test code, ya know?  Loops and
conditionals make programming difficult :)

Plus, when a test fails, the for loop makes figuring out which part of the
test triggered the problem tricky.

betatim:
Tests got parametrised to remove the loops.

ctb:
It seems like the last remaining concern here is thread safety - @betatim, thoughts?

khmer/khmer_args.py
```diff
@@ -479,13 +487,16 @@ def add_loadgraph_args(parser):
 
 def calculate_graphsize(args, graphtype, multiplier=1.0):
     """Transform the table parameters into a size."""
-    if graphtype not in ('countgraph', 'nodegraph'):
+    if graphtype not in ('countgraph', 'smallcountgraph', 'nodegraph'):
```
ctb: *net

lib/storage.hh
```diff
@@ -221,6 +221,175 @@ public:
...
+    // which you then need to select the correct nibble from
+    uint64_t _table_index(const HashIntoType k, const uint64_t tablesize) const
+    {
+        const uint64_t bins = tablesize / 2 + 1;
+        return (k / 2) % bins;
+    }
+    // Compute which half of the byte to use for this hash value
+    uint8_t _mask(const HashIntoType k) const
+    {
+        return k%2 ? 15 : 240;
```
ctb: This may be premature optimization, but instead of k%2 what about just looking at the first bit?

betatim:
Yes thread safety is missing. Thinking we can do what the other storages are doing. Somehow got distracted by #1248 

Let's wrap this up before 🎄

betatim:
The latest commit is semi related to the actual content of the PR.

While trying to track down the source of #1248 and thinking about multithreading for `NibbleStorage` I ended up with 61a07a1 as the only way to fix #1248. In which case we don't need anything threadsafe in `NibbleStorage`. Can someone help me understand why locking neither, just `get_next_read` or just `check_and_process_read` don't work?

I used `for n in {1..20}; do time python scripts/abundance-dist-single.py --threads 2 -s -x 1e7 -N 2 -k 17 -z tests/test-data/test-large.fa /tmp/test.dist 2>&1 | grep Total; done` to establish whether or not something was thread safe.

standage:
@betatim Sorry, only experience I have with parallel processing is OpenMP. I did some copy-n-paste today with #1571 but I can't say I have more than a superficial understanding of thread-safe implementation.

betatim:
Solution to the difference in n_unique_kmers for ByteStorage and NibbleStorage is in `0b2fa13`. Basically by splitting a given byte nible storage always had an even number of buckets. current double modulus isn't pretty or fast but it is correct (famous last words).

betatim:
I've ticked the multi-threading box. You can run `python scripts/abundance-dist-single.py --small-count --threads 5 -s -x 5e7 -N 2 -k 29 -z tests/test-data/test-large.fa /tmp/test.dist` (or without `--small-count`) and get a consistent number of unique k-mers. If you remove the mutex introduced in c58293e neither `Counttable` nor `SmallCounttable` manage to give a consistent answer over multiple runs.

ctb:
@betatim I don't understand why we need a mutex guard around the read retrieval code...

betatim:
I don't understand it either but empirically you need it otherwise the number of unique kmers you find changes :-/

ctb:
that is a very unsatisfying answer :)

ctb:
Seems to me like this could be merged. Update from master & merge?

standage:
Yeppers.

betatim:
:+1: 

(I can't convince myself either way whether we need the mutex or not. Basically, can we observe the mistakes from race conditions or not? Maybe something for a separate issue)

ctb:
yes, bump to issue with brief summary, pls!

PR #1556, opened 2016-12-14
Fix memory leak when converting to HashIntoType

### LINKED ISSUE #1555: Leaky k-mer abundance queries?
I've taken a script from my project kevlar, stripped it down, and modified it to iterate over the input 1e6 times. The original script populates countgraphs with real data: in this script the countgraphs are empty, but the oberved behavior is the same.

```python
#!/usr/bin/env python

from __future__ import print_function
import sys
import screed
import khmer

def kmer_is_interesting(kmer, casecounts, controlcounts, case_min=5,
                        ctrl_max=1):
    caseabund = casecounts.get(kmer)
    if caseabund < case_min:
        return False

    ctrlabunds = [counts.get(kmer) for counts in controlcounts]
    ctrlpass = [abund > ctrl_max for abund in ctrlabunds]
    if True in ctrlpass:
        return False

    return [caseabund] + ctrlabunds

proband = khmer.Countgraph(31, 1e8, 4)
mother = khmer.Countgraph(31, 1e8, 4)
father = khmer.Countgraph(31, 1e8, 4)
parents = [mother, father]

for _ in range(1000000):
    if _ > 0 and _ % 50 == 0:
        print(_, 'iterations', file=sys.stderr)
        sys.stderr.flush()
    for n, record in enumerate(screed.open(sys.argv[1])):
        for i, kmer in enumerate(proband.get_kmers(record.sequence)):
            counts = kmer_is_interesting(kmer, proband, parents, 11, 0)
            if counts:
                print('Zing!')
                sys.stdout.flush()
```

Invoked on the command line like so: `./find.py NA19240.1000reads.fq.gz > out 2> err`.

Here is the input: [NA19240.1000reads.fq.gz](https://github.com/dib-lab/khmer/files/649817/NA19240.1000reads.fq.gz).

I've run this script in various HPC environments and on my Macbook, and in each case the memory consumption gradually grows and grows until it exhausts all available resources.

Is there a leak with k-mer abundance queries? If it's not that, I'm not sure what else it could be with this script.

betatim:
Fixes #1555 Missed a DECREF when converting to `HashIntoType`

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/khmer/_khmer.cc
+++ b/khmer/_khmer.cc
@@ -187,14 +187,17 @@ static bool ht_convert_PyObject_to_HashIntoType(PyObject * value,
     if (PyInt_Check(value) || PyLong_Check(value)) {
         return convert_PyLong_to_HashIntoType(value, hashval);
     } else if (PyUnicode_Check(value))  {
-        std::string s = PyBytes_AsString(PyUnicode_AsEncodedString(
-                                             value, "utf-8", "strict"));
+        PyObject* val_as_str = PyUnicode_AsEncodedString(value,
+           "utf-8", "strict");
+        std::string s = PyBytes_AsString(val_as_str);
         if (strlen(s.c_str()) != ht->ksize()) {
+            Py_DECREF(val_as_str);
             PyErr_SetString(PyExc_ValueError,
                             "k-mer length must equal the k-mer size");
             return false;
         }
         hashval = ht->hash_dna(s.c_str());
+        Py_DECREF(val_as_str);
         return true;
 
     } else if (PyBytes_Check(value)) {


betatim:

ctb:
thanks @betatim - nice work!

standage:
Awesome, thanks!

I'm surprised that such a central feature of the code base has had an undiscovered memory leak for so long. I didn't believe it at first. Perhaps it survived for so long because most of the code that does a lot of abundance queries is straight C++ and doesn't use the CPython interface?

Anyway, thanks for the fix! 👍

ctb:
On Wed, Dec 14, 2016 at 10:38:27AM -0800, Daniel Standage wrote:
> Awesome, thanks!
> 
> I'm surprised that such a central feature of the code base has had an undiscovered memory leak for so long. I didn't believe it at first. Perhaps it survived for so long because most of the code that does a lot of abundance queries is straight C++ and doesn't use the CPython interface?
> 
> Anyway, thanks for the fix! ???? 

Hi Daniel, good question!

I don't think it was in the last release - it's certainly new since 2.0.

betatim:
I think this was a new bug, we (read Tim) moved around a lot of code and refactored the conversion code. Was wondering if there is a way to run python in "debug reference counts" mode to make it easier to find this type of leak in the future?

luizirber:
There is https://gcc-python-plugin.readthedocs.io/en/latest/cpychecker.html,
@mr-c ran it before in our codebase

Issue #1378, opened 2016-06-10
refactor consume_fasta function(s)


ctb:
We've had a challenging profileration of consume_fasta functions...

From @camillescott comment in labelhash.cc:

```
Might be time for a refactor: could do a general consume_fasta               
function which accepts a consume_sequence function pointer as a parameter
```

Issue #70, opened 2013-06-29
Remove lower_bound and upper_bound arguments in counting.cc


ctb:
Nothing uses 'em.

PR #1571, opened 2017-01-03
Introduce consume_seqfile_banding method

### LINKED ISSUE #1570: Implement a new 'consume' method that only stores a fraction of the observed k-mers
I'm proposing a new method for the (count|node)(table|graph) classes that will deterministically consume only a fraction of the k-mers observed in the reads processed. The basic idea is to use some property of the k-mer's hashed value to split it into multiple (roughly equally sized) bins, so that those bins can be analyzed independently with much lower memory requirements. The simplest approach would be to just split up the hashed values using a modulus, but this is expensive to compute. An alternative approach would be to split on the first few or last few nucleotides (which depending on the hash function used might require discarding k-mers before hashes are computed), but some prefixes/suffixes are much more abundant than others (for example, A-rich sequences in human). It would probably be best to split into 2^n bins by looking at the value encoded in the n least significant bits of the hashed k-mer.

I would suggest a Python interface like this.

```python
cg = khmer.Countgraph(31, 1e9, 4)
# consume only k-mers with 0011 as their last 4 bits
cg.consume_fasta_bitsplit('reads.fq', 16, 3)
```

standage:
WIP that, when finished, closes #1570. Implements a variant of `consume_fasta` (and all of the functions it calls) that consumes a subset of the observed k-mers. Given `num_batches = 2^n` and `batch \in range(num_batches)`, the function only stores k-mers whose n least significant bits encode `batch`.

Remaining tasks:
- [x] add checks to verify `num_bins` is a power of 2 and `tag < num_bins`
- [x] Implement CPython glue code
- [x] Add tests
- [x] Integrate with #1511 and/or related issues; this new functionality requires a proper hash function, not just 2-bit encoding

----------

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


standage:
This PR more or less duplicates the contents of several functions. If it turns out to be generally useful and something we want to keep long term, we'll need to consider how to refactor the code to minimize duplication and maintenance burden. Note, we've already acknowledged the need for some cleanup in #1378.

standage:
Ready for initial review @ctb @luizirber @camillescott. This may be easier to merge before #1511. If so I'll clean it up for final review.

lib/hashtable.cc
```diff
@@ -122,6 +122,63 @@ consume_fasta(
     delete parser;
 }
 
+void Hashtable::consume_fasta_bitsplit(
```
betatim: bitsplit doesn't mean anything to me :-/ Coudl we call it `consume_fasta_sampling` or some such?

lib/hashtable.cc
```diff
@@ -122,6 +122,63 @@ consume_fasta(
...
+    std::string const &filename, unsigned int num_bins, unsigned int tag,
+    unsigned int &total_reads, unsigned long long &n_consumed
+)
+{
+    bool powerof2 = !(num_bins == 0) && !(num_bins & (num_bins - 1));
+    if (!powerof2 || tag >= num_bins) {
+        return;
```
betatim: why not raise a `ValueError`?

betatim:
As mentioned here https://github.com/dib-lab/khmer/issues/1570#issuecomment-270215126 this needs a good hashfunction. Should make it either impossible to call with 2bit encoding or check at run time that the user knows what they are doing.

lib/hashtable.cc
standage: Agreed.

lib/hashtable.cc
standage: I agree the name is wonky. `consume_fasta_sampling` is better but honestly I've thought of half a dozen names and I'm not in love with any of them.

standage:
> tables use murmurhash

And to be clear, the concern with the hash function isn't correctness *per se*, but getting a roughly even split. Unless the sequence is truly random (simulated and biologically uninteresting), some k-mer suffixes are going to be overrepresented and skew the split. At least for my purposes, this wouldn't be a problem for *correctness* as much as for *performance*. I want to be able to split the work into evenly sized bins.

standage:
I did some [basic profiling](https://gist.github.com/standage/d8635983d02a686d0f7a72ca5a559416). For very small memory sizes, there is little difference in runtime between `consume_fasta` and `consume_fasta_bitsplit` on ≈1/8 of the k-mers. But as memory increases, the runtime of `consume_fasta` grows linearly-ish with memory while the runtime of `consume_fasta_bitsplit` increases only marginally. What's great is that `consume_fasta_bitsplit` can operate at a reasonable FPR using less memory, presumably taking less of a performance hit from NUMA.

standage:
But regardless, increasing memory only has a marginal effect on runtime. Perhaps it becomes more I/O bound?

betatim:
When you say increasing memory you mean making the hashtable bigger?

So for small hashtable sizes we spend most of our time doing I/O, hence throwing away kmers after reading them in doesn't speed things up. For large hashtables we spend most of our effort on entering stuff into the hashtable, so reading in kmers and then chucking them makes things faster?! Interesting.

ctb:
excellent. @betatim I've lost track (vacation brain) of your results on cache misses; this is probably not due to cache missing, right?

ctb:
@standage, you should be able to separate out the IO effects from the memory size effects by an affine fit, yes?

ctb:
What's old is new: #70

ctb:
Commit 786a3bec58de9c0

lib/hashtable.cc
standage: Yeah, I quite like `banding`. And I think `num_batches` and `batch` are good replacements for `num_bins` and `tag`.

tests/test_banding.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+@pytest.mark.parametrize('ksize,memory,epsilon,numbands', [
+    (21, 5e7, 1, 2),
+    (21, 5e7, 1, 4),
+    (21, 5e7, 1, 8),
+    (21, 5e7, 2, 16),
+])
+def test_banding_in_memory(ksize, memory, epsilon, numbands):
+    infile = utils.get_test_data('test-reads.fa')
```
standage: I'm not sure this function meets @ctb's simplicity criteria for automated tests. I'm not sure there's a simpler, clearer way to iterate over the data.

tests/test_banding.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.parametrize('ksize,memory,numbands', [
+    (21, 5e7, 2),
+    (21, 5e7, 4),
+    (21, 5e7, 8),
+    (21, 5e7, 16),
+])
+def test_banding_to_disk(ksize, memory, numbands):
+    infile = utils.get_test_data('test-reads.fa')
+    path1 = utils.get_temp_filename('normal.ct')
+    path2 = utils.get_temp_filename('banding.ct')
```
standage: This function is simple enough, and it provides a stronger guarantee. Perhaps it's all we need?

ctb:
*shrug* I don't like procedural complexity.  At least this is declarative :)

standage:
Procedural complexity == branching?

standage:
My recent tests have put my concerns regarding the hash functions and banding to rest. Now that #1511 is merged, this is now ready for review and merge. @ctb @betatim @camillescott @luizirber

lib/hashtable.cc
```diff
@@ -120,6 +121,82 @@ void Hashtable::consume_fasta(
...
+            read = parser->get_next_read();
+        } catch (NoMoreReadsAvailable) {
+            break;
+        }
+
+        if (read.sequence.length() < _ksize) {
+            continue;
+        }
+
+        bool is_valid = true;
```
betatim: Can I persuade you to use `read.set_clean_seq();` plus `read.cleaned_seq != read.sequence` instead of doing your own processing/normalising?

betatim: Maybe what we need is a method that does the toupper() for us to avoid having "loop-over-seq-and-uppercase-then-check" here.

tests/test_tabletype.py
```diff
@@ -15,6 +15,7 @@
 code...
 """
 
+from __future__ import absolute_import
```
betatim: Was this missing before? Puzzled how travis didn't find it.

betatim:
Do you want to add one more test that computes the hash of two kmers, one that should be added and one that shouldn't, and explicitly checks that they ended up in the CT? A bit like https://github.com/dib-lab/khmer/pull/1571/files#diff-7352ff13b9e5cdbeacb24c9e7bcb49b1R1136 except here I am guessing that this is what you wanted because I don't know the hash values off the top of my head.

ctb:
see also #1590

lib/hashtable.cc
standage: Two concerns / observations.

- Comparing cleaned seq vs original seq will disqualify reads with lowercase ACGT, which I think is a lousy default with unexpected consequences.
- The current `consume_fasta` function in Hashtable doesn't use this utility yet either.

As @ctb mentioned, there's some work to be done with removing redundant read cleaning code (see #1590), and that's not all going to be resolved in this PR. Ultimately this will be handled by the parser instead of the consume function, right?

tests/test_tabletype.py
standage: Yeah, weird.

byallaccountsitdoesntmakesense

standage:
@betatim Yeah, that's exactly what I was going for with that test. Are you saying to test the hash value instead of the k-mer itself? Please clarify.

betatim:
I was going for a test like this:
```
ct = Counttable(...)
h = ct.hash(some_kmer)
assert h % 8 == 3
ct.consume_seqfile_banding(file_that_contains_some_kmer, 8, 3)
assert ct.get(some_kmer) == 1
```
we could even add the assert for the hash value to the current test. The motivation is to leave an explicit clue as to why you expect a certain kmer to be present and which ones not.

tests/test_tabletype.py
betatim: Now we need it to make the tests pass or do they work without as well?

lib/hashtable.cc
betatim: Ah yes, humm, so we definitely need a method that can do the comparison that is centralised. And with cython'isation things will move again. Punt?

ctb:
On Thu, Feb 09, 2017 at 08:21:10PM -0800, Daniel Standage wrote:
> Procedural complexity == branching?

yes, specifically 'if' statements suck. if we have to debug our test code
itself, it adds complexity.

standage:
Closing in favor of #1720.

PR #1573, opened 2017-01-15
enable static building using pkg-config

satta:
- [X] Is it mergeable?
- [X] `make test` Did it pass the tests?

This PR does not touch the C++ or Python code. It only adds to line to the pkg-config template that allow pkg-config to output the correct flags for static linking. That makes it possible to develop against liboxli by, for example:

```bash
g++ -o test-prog-static -static -std=c++11 `pkg-config oxli --static --cflags` \
  test-compile.cc `pkg-config oxli --static --libs`
```

### DIFF
--- a/lib/oxli.pc.in
+++ b/lib/oxli.pc.in
@@ -10,5 +10,7 @@ URL: http://khmer.readthedocs.io/
 Version: @VERSION@
 
 Requires:
+Requires.private: zlib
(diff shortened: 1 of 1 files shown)


ctb:
thank you!

this sounds fine to me @luizirber @betatim but I would appreciate your
thoughts.

Issue #1541, opened 2016-11-24
More thoughts on non-ACGT characters.


ctb:
Spurred by #1540, I'm converging on two principles regarding non-ACGT:

* we shouldn't ever silently ignore an entire sequence record, because that leads to surprisingly inconsistent behavior between short reads and genomic sequences, as in the above where an entire genome vanishes from the k-mer counts. So we should either raise an exception OR automatically convert.
* automatic conversion to As or something is probably the right default behavior, unless `--check-sequence` or `--pedantic` is specified.  In part this is because even NCBI records have non-ACGT IUPAC bases in them.

There is a fun cautionary tale [in this twitter exchange](https://twitter.com/bioinformer/status/800789395271991296) that is relevant, I think; we want to get the k-mer breakdown at least approximately correct :).  This is another reason I think automatic conversion is appropriate - if the sequence type is completely wrong (e.g. protein, or RNA, or text) then the k-mer counts will be visibly off, but this is rare compared to the situation where you are looking at mostly legit DNA sequence and there's an S in the middle of a 5 Mbp genome.  (I used to think we should automatically convert until some threshold was reached, like "1% of your sequence is bad... STOP", but that's overly clever and will no doubt bite us.

Last but not least, we should include a script (in khmer? screed?) that does a breakdown of the non-ACGT content of your data set so that people can debug more easily.

betatim:
How do the non-ACGT letters get into the data in the first place?

ctb:
Various intentional and unintentional bits of cleverness.  Some software
uses [IUPAC codes](http://www.bioinformatics.org/sms2/iupac.html) to indicate
observed ambiguity, for example.

betatim:
So there is the case where the user is probably not doing what they think they are doing (feeding not DNA sequence into khmer) and the cleverness case. For the latter, should we be clever and sample a replacement from the possibilities (e.g. replace S with `random.sample(['G', 'C'])`)? Or even add one kmer with each option? Or is that (trying to be clever)**2?

ctb:
I think that's trying to be too clever ;).

ctb:
I wonder if we can look at the beginning of the file and do some simple statistics (on, say, the first 100kb of sequence) and then barf if it looks bad?

This might be a good use for issues with bad pairing as well - scripts that are told to expect pairing info and don't get any pairs could barf.

After the first 100kb we could then assume that all the data is "good format" and go from there. Will probably catch 90%+ of problems...

[...]

PR #1590, opened 2017-01-24
[MRG] remove redundant ACGT-checking code.

ctb:
This removes `check_and_process_read` from `Hashtable`, and eliminates sequence cleaning from non-bulk-loading code. All such Python-accessible code should use the `cleaned_seq` attribute that is available from both the `ReadParser` code / `Read` objects and the `khmer.utils.broken_paired_reader` code. 

----

See https://github.com/dib-lab/khmer/issues/1541#issuecomment-274828660 for background context.

Prior to this PR, 

* `consume_fasta` & other bulk sequence loading functions automatically upper-cased DNA and ignored any sequence (no matter how long) that contained non-ACGT.
* `trim_on_abundance` and `trim_below_abundance` similarly upper-cased DNA and ignored non-ACGT-containing strings.
* `find_spectral_error_positions` raised an exception on non-ACGT-containing strings.

This PR updates the test code introduced in #1633 to match the new behavior.

This PR also includes #1661.

-----

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


ctb:
Although reading #1491 I'm wondering if we should be using `cleaned_seq` or equivalent in the C++ bulk `consume_fasta` code, instead of `read.sequence`. If so, then we really need to add some tests so that the current code breaks :).

ctb:
Also see https://github.com/dib-lab/khmer/issues/1619#issuecomment-279801266

camillescott:
Also see #1595 -- some more sophisticated alphabet checking was added to the parsers. Basically, you can give it one of the alphabets (as defined in `alphabets.hh`) by name, and it'll use that for its checks; as perhaps implied by that, I'm in favor of moving this sort of checking code into the parsers. Alternatively, the graphs could take a parameter for whether they should run any checks on what they consume (also using the alphabets).

camillescott:
And, looking at #1511 closer, moving that checking code into the parsers is already becoming SOP. Yay.

ctb:
yeppers.

ctb:
@standage note that one thing the partitioning code does is fail to output reads contaiing Ns.

standage:
> note that one thing the partitioning code does is fail to output reads containing Ns

Roger that, good to know.

ctb:
On Mon, Feb 20, 2017 at 08:07:57AM -0800, Daniel Standage wrote:
> > note that one thing the partitioning code does is fail to output reads containing Ns
> 
> Roger that, good to know.

(for now - fixing it in this PR :)

ctb:
I wonder if this code is any faster now that we don't iterate across every sequence 2 or 3 times?

betatim:
Time to review?

ctb:
On Wed, Mar 22, 2017 at 07:47:42AM -0700, Tim Head wrote:
> Time to review?

yes! but this will not be merged until after release.

betatim:
Ran `./scripts/abundance-dist-single.py -x 1e8 -b ecoli_ref-5m.fastq somehisto.hist -k 31 -s` which takes about 120s on my machine both on master and this branch :-/

ctb:
@betatim @luizirber @camillescott @standage this PR is now ready for merge into master and (IMO) should be given high priority for review.  @camillescott's prediction of many merge conflicts was mistaken, thank goodness - only one small conflict to be resolved!

This merge significantly changes the details around ACTGN handling and will require a bump to khmer 3.0.

ctb:
Tests pass!

tests/test_sequence_validation.py
```diff
@@ -152,51 +152,60 @@ def test_read_cleaning_trim_functions_lowercase(Tabletype, reads):
...
-    x = Tabletype(8, PRIMES_1m)
+    x = Countingtype(8, PRIMES_1m)
     for read in reads:
         x.consume(read.cleaned_seq)       # consume cleaned_seq
 
+    # the precise behavior of these functions is all *undefined*
+    # because different hash functions do different things with
+    # non-ACTG characters.  So all we want to do is verify that the
+    # functions execute w/o error on the k-mers before the "bad" DNA,
+    # and don't return positions in the "good" DNA.
```
standage: The meaning of the phrase *and don't return positions in the "good" DNA* is unclear to me.

standage (approved):
Looks good to me. As I understand it, khmer now makes no attempt to clean up kmers that are passed to the `sketch.add()` function, and does a `[^ACGT] --> A` conversion for any bulk sequence loading code. Correct?

standage:
Also, does the partitioning code now output reads containing Ns @ctb?

ctb:
@standage yes; re partitioning and ignoring of reads containing N, see: [this removed line](https://github.com/dib-lab/khmer/pull/1590/files#diff-22f0123dd35e0b63ab97e9c8a2c649f9L378)

tests/test_sequence_validation.py
ctb: The behavior of the hash functions on ACTG should all be the same with respect to "good" sequences, and there should be no errors in there to trim at.

PR #1538, opened 2016-11-21
Add broken pair reader

betatim:
WIP

Copy functionality of `utils.broken_paired_reader` to C++

- [ ] Is it mergeable?
- [ ] `make test` Did it pass the tests?
- [ ] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?
- [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
  formatted?
- [ ] Did it change the command-line interface? Only backwards-compatible
  additions are allowed without a major version increment. Changing file
  formats also requires a major version number increment.
- [ ] For substantial changes or changes to the command-line interface, is it
  documented in `CHANGELOG.md`? See [keepachangelog](http://keepachangelog.com/)
  for more details.
- [ ] Was a spellchecker run on the source code and documentation after
  changes were made?
- [ ] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)

PR #1591, opened 2017-01-24
Add  cleaned_seq attribute to `Read` class

betatim:
Compute the cleaned version of a read's sequence within the `Read` object.

This isn't ideal yet because the user of `Read` has to remember to call `set_cleaned_seq`. This is only a problem for c++ users, in python we can automate this. In c++ it is difficult because of the way `seqan` works which gets access directly to the sequence member of `Read` (we can't make it private).

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/utils.py
+++ b/khmer/utils.py
@@ -183,7 +183,7 @@ def broken_paired_reader(screed_iter, min_length=None,
         raise ValueError("force_single and require_paired cannot both be set!")
 
     # handle the majority of the stream.
-    for record in screed_iter:
+    for record in clean_input_reads(screed_iter):
         if prev_record:
             if check_is_pair(prev_record, record) and not force_single:
                 if min_length and (len(prev_record.sequence) < min_length or
@@ -258,6 +258,19 @@ def write_record_pair(read1, read2, fileobj):
         fileobj.write(bytes(recstr, 'ascii'))
     except TypeError:
         fileobj.write(recstr)
+
+
+def clean_input_reads(records):
+    """Add a cleaned_seq attribute to records that do not have one
+
+    Use this to convert a stream of records that might not have a
+    `cleaned_seq` attribute to one that does. Use this to extend
+    Records loaded by `screed.open()`.
+    """
+    for record in records:
+        if not hasattr(record, 'cleaned_seq'):
+            record.cleaned_seq = record.sequence.upper().replace('N', 'A')
+        yield record
 
 
 class ReadBundle(object):
--- a/khmer/utils.py
+++ b/khmer/utils.py
@@ -183,7 +183,7 @@ def broken_paired_reader(screed_iter, min_length=None,
         raise ValueError("force_single and require_paired cannot both be set!")
 
     # handle the majority of the stream.
-    for record in clean_input_reads(screed_iter):
+    for record in screed_iter:
         if prev_record:
             if check_is_pair(prev_record, record) and not force_single:
                 if min_length and (len(prev_record.sequence) < min_length or
--- a/khmer/utils.py
+++ b/khmer/utils.py
@@ -265,11 +265,11 @@ def clean_input_reads(records):
 
     Use this to convert a stream of records that might not have a
     `cleaned_seq` attribute to one that does. Use this to extend
-    Records loaded by `screed.open()`.
+    Records loaded by `screed.open()`. It is a mistake to apply
+    this to a `ReadParser` stream.
     """
     for record in records:
-        if not hasattr(record, 'cleaned_seq'):
-            record.cleaned_seq = record.sequence.upper().replace('N', 'A')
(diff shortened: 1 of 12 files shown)


betatim:
With 2126d83 write access to the `cleaned_seq` attribute is gone for python users. Until now `cleaned_seq` was special because unlike the other attributes it did allow python people to write to it.

betatim:
If you run this branch with `screed` after dib-lab/screed#66 has been merged all the tests pass except for one that creates a screed Record by hand. How do we handle this?

I think it makes sense to remove  `clean_input_reads` as part of this PR. `khmer` will provide clean reads by default after this PR. We could add it to `screed` maybe, but I prefer dib-lab/screed#66.

ctb:
I'm -1 on this PR at the moment.

The main reason is that we've deliberately kept screed agnostic to the
question of sequence content.  So it handles RNA and protein as well as
DNA, "supports" IUPAC by not touching it, and handles genomes and genes as well
as sequencing data.  This is in stark contrast to khmer which really only is
intended to deal with DNA and sequencing data.

In particular, I bet that RNA and protein loading is going to return
non-sensical 'cleaned_seq' results :).  But, generally speaking, there are
many ways to torture input data.

The other reason (which is less of a blocker) is that when you're loading
a 3 GB genome to keep in memory, you don't want to make extra copies of each
sequence, or even preprocess it in any particular standard way.

(I was -0.5 until I wrote this all down, and now I'm solidly against it ;)

I'm not sure what the implications are for khmer and #769, but that seems like
khmer's problem, not screed's.  #1466 and #1483 are relevant.  Maybe the rule
is that khmer decorates `screed.Record` as it pleases, and `Read` and the
*decorated* `screed.Record` objects should look the same?

ctb:
Whoops, sorry, commented on the wrong issue. Sorry for 'sploding your head, @betatim.

betatim:
Added `clean_input_reads` back. It checks if there is a `cleaned_seq` attribute and if not sets it.

Any smart ideas how to work around `seqan`'s need to set members of `Read` directly and not being able to make them private?

ctb:
What about removing `clean_input_reads` from the `broken_paired_reader` altogether, i.e. altering `broken_paired_reader` to expect sequences to already have that attribute?

Motivation/thoughts:

* this way we avoid the extra `if` statement in `broken_paired_reader`;
* we can still wrap screed iterators in `clean_input_reads` if and when we use them, which in any case will be rare now that ReadParser is a drop-in replacement;
* it's a pretty unambiguous error for code downstream of `broken_paired_reader` to raise an exception because `cleaned_seq` isn't available;

betatim:
Good shout :smile: so now I think we should merge this before #1538

betatim:
Questions to tick off before merge:

* happy with the interface and the new burden of having to call `set_cleaned_seq` if you create Read`s in c++
* performance/memory concerns regarding having a copy of every bit of sequence that is loaded
* new PR to switch over c++ code to use `cleaned_seq` instead of `sequence`?
* can you think of a situation where we would end up with `sequence` and `cleaned_seq` being out of sync? Right now I can't (beyond a c++ coder making a mistake :-/ )

khmer/utils.py
```diff
@@ -260,9 +260,16 @@ def write_record_pair(read1, read2, fileobj):
...
-        record.cleaned_seq = record.sequence.upper().replace('N', 'A')
+def clean_input_reads(records):
+    """Add a cleaned_seq attribute to records that do not have one
+
+    Use this to convert a stream of records that might not have a
+    `cleaned_seq` attribute to one that does. Use this to extend
+    Records loaded by `screed.open()`.
+    """
+    for record in records:
+        if not hasattr(record, 'cleaned_seq'):
```
ctb: This can be removed now, right?

ctb:
> happy with the interface and the new burden of having to call `set_cleaned_seq` if you create Read`s in c++

yep!

> performance/memory concerns regarding having a copy of every bit of sequence that is loaded

no, not for sequencing reads (which is what khmer is designed for)

> * new PR to switch over c++ code to use `cleaned_seq` instead of `sequence`?

maybe, I think some tests need to be added to #1590 first so that we can see the problem with the C++ code!

> can you think of a situation where we would end up with `sequence` and `cleaned_seq` being out of sync? Right now I can't (beyond a c++ coder making a mistake :-/ )

I cannot!

khmer/utils.py
betatim: It depends. Without this if you accidentally pass a stream of `khmer.Read`s into this then it will break because all attributes are readonly. So mostly a policy question.

ctb:
as long as it visibly breaks, that sounds perfect!

but we should have a test for that behavior just to be sure :)

PR #1596, opened 2017-01-26
Deal with some hashing issues.

### LINKED ISSUE #1540: unique-kmers.py ignores the entirety of any sequence with non-ACGTN in it
e.g. if a chromosome from a genome sequence has a single non-ACGTN in it, the entire sequence will be ignored. This is unexpected.

This should probably be changed so that such k-mers can be skipped over, or optionally raise an exception - e.g. in the sbt_search branch of sourmash compute, I now convert non-ACGT bases into 'N', unless `--check-sequence` is specified.  The khmer-consistent approach would be to convert them into 'A' (see `khmer.utils.clean_input_reads`).

ctb:
A few fixes and simplifications --

Internal hash functions no longer "care" about bad DNA. This was a result of [this comment](https://github.com/dib-lab/khmer/pull/1511#issuecomment-274712426) and [this comment](https://github.com/dib-lab/khmer/pull/1511#issuecomment-274820149) in #1511.  In any case, sequence checking was done haphazardly in the reverse complement calculation.

`_hash_murmur` (at the C++ level) now requires that the k-mer size gets passed in; this fixes some mismatches between has the old hash functionality and the new hash functionionality.

Also changed a bunch of `RuntimeError` exceptions over to `ValueError`.

I think this fixes #1540, too.

The best (or at least *first*) place to look is at the changes to the tests...

- [x] Is it mergeable?


ctb:
@betatim @standage any thoughts welcome; I'll merge this in the next day or two.

khmer/_khmer.cc
```diff
@@ -176,7 +176,15 @@ static bool ht_convert_PyObject_to_HashIntoType(PyObject * value,
...
         }
-        hashval = ht->hash_dna(s.c_str());
+
+        try {
+            hashval = ht->hash_dna(s.c_str());
+        } catch (khmer_exception &e) {
+            PyErr_SetString(PyExc_ValueError, e.what());
+            Py_DECREF(val_as_str);
+            return false;
+        }
```
standage: I've passively seen code like this before, but I've never had a reason to pay close attention. With recent issues fresh on my mind, nice to see an example of C++ to Python exception handling.

lib/kmer_hash.cc
```diff
@@ -59,10 +60,14 @@ HashIntoType _hash(const char * kmer, const WordLength k,
...
 {
     // sizeof(HashIntoType) * 8 bits / 2 bits/base
-    if (!(k <= sizeof(HashIntoType)*4) || !(strlen(kmer) >= k)) {
+    if (k > sizeof(HashIntoType)*4) {
         throw khmer_exception("Supplied kmer string doesn't match the underlying k-size.");
     }
 
+    if (strlen(kmer) < k) {
+        throw khmer_exception("k-mer is too short to hash.");
+    }
```
standage: Glad this is being handled at the C++ level and not just the CPython level.

tests/test_tabletype.py
```diff
@@ -64,7 +67,22 @@ def test_hash(tabletype):
...
+
+    x = kh.hash("ATGYC")
+
+
+def test_hash_bad_length(tabletype):
+    # hashing of bad dna length -> error
+    kh = tabletype(5, PRIMES_1m)
+
+    with pytest.raises(ValueError):
+        x = kh.hash("ATGGGC")
```
betatim: for good measure also add one with a kmer that is too short?

tests/test_tabletype.py
```diff
@@ -76,7 +94,7 @@ def test_reverse_hash(tabletype):
     except ValueError:
         pytest.skip("reverse_hash not implemented on this table type")
 
-    assert type(x) == str
+    assert type(x) in (unicode, str)
```
betatim: `isinstance(x, (unicode, str))`

betatim:
LGTM, if you indulge my nitpicks

tests/test_tabletype.py
ctb: done in 7e1ba7b

tests/test_tabletype.py
ctb: done in 64fe859

PR #1451, opened 2016-09-20
Support for different read parsers using functor templates

standage:
Working branch. Will replace the `IParser` abstract class with a concrete class `ReadParser`. Imprinting reads in various formats (Fast[a,q] and BAM, initiallly) will be delegated by functor classes, which are declared in the `ReadParser` class using templates.
- [ ] Is it mergeable?
- [ ] `make test` Did it pass the tests?
- [ ] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?
- [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
  formatted?
- [ ] Did it change the command-line interface? Only additions are allowed
  without a major version increment. Changing file formats also requires a
  major version number increment.
- [ ] Is it documented in the `ChangeLog`?
  http://en.wikipedia.org/wiki/Changelog#Format
- [ ] Was a spellchecker run on the source code and documentation after
  changes were made?
- [ ] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)
- [ ] Is the Copyright year up to date?

Issue #1574, opened 2017-01-17
Minimum GCC requirements should be documented


ctb:
See: https://github.com/dib-lab/sourmash/issues/98#issuecomment-272892910

betatim:
As I don't know the explicit minimum version, here an overview of what I found:

On travis linux we use gcc-4.8.4, https://gcc.gnu.org/projects/cxx-status.html#cxx11 says that 4.8.1 was the first version to fully support c++11. AFAIK we don't use any nasty hacks, just c++11.

standage:
Fixed in #1600.

PR #1597, opened 2017-01-26
Support for different input formats using templates: take 2

standage:
I originally explored this approach in #1451, but got stuck on the templating details. After giving this a rest for a few months, I revisited today and have been able to address the linker issues that plagued me before. I had hoped to avoid using pointers altogether and take a more explicit approach, but this turned out to be untenable. Instead I transitioned to the use of smart pointers with read parsers.

This PR doesn't add BAM support, but it lays the groundwork to do so!

Remaining:

- [x] Replace `typename ParseFunctor` with something else since the functor approach didn't end up working with pointers.
- [x] Figure out why `abundance_distribution` tests are failing.
- [ ] Figure out why `RIGHT` and `LEFT` had conflicts with SeqAn on this branch but not master.

-----

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/lib/hashgraph.cc
+++ b/lib/hashgraph.cc
@@ -276,20 +276,20 @@ void Hashgraph::consume_sequence_and_tag(const std::string& seq,
 //
 
 // TODO? Inline in header.
-template<typename ParseFunctor>
+template<typename SeqIO>
 void Hashgraph::consume_fasta_and_tag(
         std::string const &filename,
         unsigned int &total_reads,
         unsigned long long &n_consumed
 )
 {
-    ReadParserPtr<ParseFunctor> parser = get_parser<ParseFunctor>((std::string&)filename);
-    consume_fasta_and_tag<ParseFunctor>(parser, total_reads, n_consumed);
+    ReadParserPtr<SeqIO> parser = get_parser<SeqIO>((std::string&)filename);
+    consume_fasta_and_tag<SeqIO>(parser, total_reads, n_consumed);
 }
 
-template<typename ParseFunctor>
+template<typename SeqIO>
 void Hashgraph::consume_fasta_and_tag(
-        ReadParserPtr<ParseFunctor>& parser,
+        ReadParserPtr<SeqIO>& parser,
         unsigned int &total_reads,
         unsigned long long &n_consumed
 )
@@ -344,7 +344,7 @@ void Hashgraph::divide_tags_into_subsets(unsigned int subset_size,
 // consume_partitioned_fasta: consume a FASTA file of reads
 //
 
-template<typename ParseFunctor>
+template<typename SeqIO>
 void Hashgraph::consume_partitioned_fasta(
         const std::string &filename,
         unsigned int &total_reads,
@@ -354,7 +354,7 @@ void Hashgraph::consume_partitioned_fasta(
     total_reads = 0;
     n_consumed = 0;
 
-    ReadParserPtr<ParseFunctor> parser = get_parser<ParseFunctor>((std::string&)filename);
+    ReadParserPtr<SeqIO> parser = get_parser<SeqIO>((std::string&)filename);
     Read read;
 
     string seq = "";
--- a/lib/hashgraph.cc
+++ b/lib/hashgraph.cc
@@ -898,3 +898,8 @@ template void Hashgraph::consume_fasta_and_tag<read_parsers::FastxReader>(
     unsigned int &total_reads,
     unsigned long long &n_consumed
 );
+template void Hashgraph::consume_partitioned_fasta<read_parsers::FastxReader>(
+    const std::string &filename,
+    unsigned int &total_reads,
+    unsigned long long &n_consumed
+);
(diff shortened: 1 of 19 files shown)


standage:
I'm still stumped at why I had to redefine LEFT and RIGHT to TRAVERSAL_LEFT and TRAVERSAL_RIGHT to avoid conflicts with SeqAn, when I didn't touch that code otherwise and the master branch has no problems. Any thoughts @camillescott?

Otherwise, ready for review! @ctb @betatim @luizirber

ctb:
One question - does this increase the minimum required gcc version?

lib/hashgraph.cc
```diff
@@ -276,29 +276,22 @@ void Hashgraph::consume_sequence_and_tag(const std::string& seq,
...
-    IParser *	  parser =
-        IParser::get_parser( filename );
-
-    consume_fasta_and_tag(
-        parser,
-        total_reads, n_consumed
-    );
-
-    delete parser;
+    ReadParserPtr<SeqIO> parser = get_parser<SeqIO>((std::string&)filename);
```
betatim: do you know why we need this (std::string&) business?

lib/hashgraph.cc
```diff
@@ -276,29 +276,22 @@ void Hashgraph::consume_sequence_and_tag(const std::string& seq,
 //
 
 // TODO? Inline in header.
-void
-Hashgraph::
-consume_fasta_and_tag(
-    std:: string const  &filename,
-    unsigned int	      &total_reads, unsigned long long	&n_consumed
+template<typename SeqIO>
```
betatim: Why do we need to template this? `consume_fasta_and_tag` (and other `*_fasta_*` methods) will always use the the fastx parser no? Or would you call `consume_fasta_` with a BAM parser instead? That seems weird.

In my mind a Fasta and BAM parser both subclass a common `Parser` so we can pass around a pointer/reference to the instance after we create it (at the point where the decision is made).

Help me see where you are going with this :)

lib/hashgraph.cc
standage: I'm using casting to eliminate errors with inconsistent use of `const` in the function call tree. I'll see if I can revisit this and make it cleaner.

standage:
> One question - does this increase the minimum required gcc version?

Not sure, but I doubt it. This PR uses `#include <memory>` and `std::unique_ptr`, but these are standard C++11 features that I thought were already in use in khmer (although checking again just now it looks like not--maybe in a pending PR?).

lib/hashgraph.cc
standage: My ultimate plan is to change `consume_fasta` and company to `consume_seq` or `consume_seqfile` or `consume_infile` or something like that (suggestions welcome), with ReadParser handling the generic stuff and templated readers handling the format-specific sequence loading.
When all is said and done the function instantiations would look like `Hashtable::consume_seqfile<FastxReader>` and `Hashtable::consume_seqfile<BamReader>`.
However, I thought it better to split this up into three PRs rather than a single one.

- In the first PR, just simply replace the existing functionality with the new data structure and template framework.
- In the second PR, replace the `consume_fasta` function names with something more generic.
- In the third PR, introduce the BAM reader and extend the Python API to enable parsing Fastx or BAM.

Yes, this all could be done with inheritance, but I figured object composition and templating would enable better performance (handle polymorphism at compile time rather than runtime).

standage:
gcc 4.8.4 is the version that comes standard on Ubuntu 14.04, and it has no problem with either the master branch or this PR. Based on https://github.com/dib-lab/khmer/issues/1574#issuecomment-273501670 it looks like this is pretty close to the earliest complete C++11 support. I'm happy with that! :-)

lib/hashgraph.cc
betatim: Ok.

I have no speed data on `consume_seq<SomeReader>(fname)` vs `consume(fname, SomeReaderInstance)` vs passing in the reader when constructing the Hashtable and storing it as a member.

standage: Do you think we need that data before this PR is merged?

betatim: Don't think so. If we want to back a decision with "for speed" we should measure it, otherwise we can just say "liked this better" :) Personal preference is `consume(fname, reader)` but that is mostly taste, not any objective reason.

lib/hashgraph.cc
betatim: We can't move this cast further down by adding more `const`s?

betatim:
I have nothing on your `RIGHT`/`LEFT` thing :-(

lib/hashgraph.cc
standage: 👍

lib/hashgraph.cc
standage: Yeah, just pushed a commit that cleaned this up a bit.

standage:
I don't think the `RIGHT` / `LEFT` thing is a blocker, arbitrary enough and easy enough switch back. I addressed the const casting weirdness, so barring any additional concerns this should be ready to merge once Travis is happy.

betatim:
Looks like the OSX build failed because of travis. Restarted the build. Happy to merge after.

standage:
Failed again and I had to restart :-(

standage:
Build passed, and since @betatim probably isn't awake, I'm merging while the merging's good!

betatim:
Got to take advantage of those small windows when all the travis stars align!

PR #1600, opened 2017-01-30
Add coment about gcc version requirement

betatim:
Related to #1574. When looking for docs on the minimum gcc version I found that the docs already contain info. Added a sentence stating which version we use on travis.

- [x] Is it mergeable?
  formatted?


betatim:

Issue #667, opened 2014-11-20
RFC: Exposing the raw data in a count-min sketch/bloom filter via numpy?


kdmurray91:
Hi all,

Firstly, forgive me if this has been discussed elsewhere and my google-fu is lacking.

I've  been hacking with various things around khmer that use the values of buckets in a count min sketch. Currently, I don't see any simple way to go thru the CountingHash's tables bin-wise. What do you think about making a transparent (possible R/W) numpy interface to these tables. Passed by reference of course, which I think would involve creating a numpy C array struct and pointing its data member it at the start of the table.

In asking this question, I am volunteering to do this :smile:, however it might not happen very soon.

The API I'd propose would be:
```
ht.tables = (np.array([0,0,0,0,0, dtype=uint8]),
             np.array([0,0,0,0,0,0,0, dtype=uint8]))
```
for a CMS of 2 tables >= 5

Cheers,
K

kdmurray91:
I'm happy for this to be closed now #671 is merged :)

PR #1601, opened 2017-01-30
Atomic nibble instead of mutex

betatim:
WIP

`NibbleStorage` uses a big (bad) mutex. This is pretty slow. This PR removes that and instead uses an array of atomic bytes.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


lib/storage.hh
```diff
@@ -393,7 +403,7 @@ public:
 
     Byte ** get_raw_tables()
     {
-        return _counts;
+        return (Byte **)_counts;
```
betatim: Completely evil :-/

betatim: Not sure what to do about this. What is the use case for `get_raw_tables()` beyond nosy devs wanting to peak inside?

ctb:
the explicit rationale is somewhere in the merged pull request history...
seek answers there :)

betatim:
Seems like this was the first time this came up: https://github.com/dib-lab/khmer/issues/667 From what I gather of the discussion there `get_raw_tables` for anything using <8bits per bucket is not what they were expecting to get. Same goes for the Node* classes no?

I don't think we can use the buffer interface to fake the "round up to nearest byte".

Should we move it up the inheritance tree to expose it only for `Count*` classes?

standage:
> Should we move it up the inheritance tree to expose it only for Count* classes?

I think this is reasonable.

betatim:
`get_raw_tables` has been moved and tests adjusted. 

Issue #1402, opened 2016-07-10
make: *** [khmer/_khmer.so] Error 1 on MSU HPCC


ljcohen:
Installing khmer dev version in a clean conda environment on msu hpcc, but it is failing. Any ideas?

Steps:
conda create -n khmer_dev anaconda
source activate khmer_dev
pip install -U setuptools --ignore-installed
git clone https://github.com/dib-lab/khmer.git
make install

```
(khmer_dev) [ljcohen@dev-intel14 khmer]$ make install
./setup.py build install
running build
running build_py
running egg_info
writing requirements to khmer.egg-info/requires.txt
writing khmer.egg-info/PKG-INFO
writing top-level names to khmer.egg-info/top_level.txt
writing dependency_links to khmer.egg-info/dependency_links.txt
reading manifest file 'khmer.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'CONTRIBUTING.md'
warning: no files found matching 'get_version.py' under directory 'lib'
warning: no files found matching 'README.txt' under directory 'scripts'
warning: no previously-included files matching '*.orig' found anywhere in distribution
warning: no previously-included files matching '*.pyc' found anywhere in distribution
writing manifest file 'khmer.egg-info/SOURCES.txt'
UPDATING build/lib.linux-x86_64-2.7/khmer/_version.py
set build/lib.linux-x86_64-2.7/khmer/_version.py to '2.0+191.g45a5195'
running build_ext
bash -c cd third-party/zlib && ( test Makefile -nt configure || bash ./configure --static ) && make -f Makefile.pic PIC
make[1]: Entering directory `/mnt/home/ljcohen/bin/khmer/third-party/zlib'
make[1]: Nothing to be done for `PIC'.
make[1]: Leaving directory `/mnt/home/ljcohen/bin/khmer/third-party/zlib'
bash -c cd third-party/bzip2 && make -f Makefile-libbz2_so all
make[1]: Entering directory `/mnt/home/ljcohen/bin/khmer/third-party/bzip2'
#gcc -shared -Wl,-soname -Wl,libbz2.so.1.0 -o libbz2.so.1.0.6 blocksort.o huffman.o crctable.o randtable.o compress.o decompress.o bzlib.o
#gcc -fpic -fPIC -Wall -Winline -O2 -g -D_FILE_OFFSET_BITS=64 -o bzip2-shared bzip2.c libbz2.so.1.0.6
[...]

PR #1602, opened 2017-01-30
Catch error trying to delete tempdir on NFS

standage:
Issue #1402 manifests itself consistently on various HPC systems, both in the tests and when `trim-low-abund.py` is run interactively. I've had a few successful runs now by simply commenting out the code that deletes the temp directory at the end of the script. This PR puts a little bit of lipstick on that pig, making a backwards-compatible addition to the command-line interface. This (hopefully) isn't a permanent solution, but I don't have the time or interest to research the problem more thoroughly right now.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,6 @@ under semantic versioning, but will be in future versions of khmer.
 
 ## [Unreleased]
 ### Added
-- New `--keep-temp` option for `trim-low-abund.py` that does not delete the
-  temporary directory created by the script. This sidesteps NFS-related issues
-  we've seen on multiple HPC file systems.
 - New `--no-reformat` option for `interleave-reads.py` script disables default
   read name correction behavior.
 - New `HashSet` data structure for managing collections of k-mer hashes and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ under semantic versioning, but will be in future versions of khmer.
 
 ## [Unreleased]
 ### Added
+- New `--keep-temp` option for `trim-low-abund.py` that does not delete the
+  temporary directory created by the script. This sidesteps NFS-related issues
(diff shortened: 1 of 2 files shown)


standage:
Ready for review!

betatim:
In principle fine, do you understand what causes the removing of the directory to fail? The only alternative would be to catch the exception of `rmtree`, try again (and again (and again)).

According to google these `.nfs000` files get created when someone deletes a file that is still open by another process. So the 'real" solution will involve finding who that is (presumably somewhere in `khmer`).

betatim:
My guess would be that https://github.com/dib-lab/khmer/blob/915069c89945ef2f6061ca7da3e64e4dc782ecf8/scripts/trim-low-abund.py#L419 is what might still have a file open when you want to remove the directory.

standage:
> Do you want to do this in this PR or merge this and move on to other things? In which case I can take a look.

A proper fix would be nice, but I'm not eager to look into it myself.

betatim:
Created #1603 to remind us to look into this.

While we wait for travis, @ctb (or others) any opinions on what we do with this "temporary" commandline argument? I think we can make it vanish again before the next release, so we don't run into the rule against changing the interface.

ctb:
A simpler approach might be to put try/except around the rmtree and simply report the error without failing, no?

standage:
> A simpler approach might be to put try/except around the rmtree and simply report the error without failing, no?

Great suggestion. Just updated and got my first successful `make test` on the Cabernet cluster for a long time. I usually have a dozen OSErrors from trim-low-abund tests to ignore.

Ready for a re-review and merge. @ctb @betatim

standage:
@ctb See #1608.

PR #1607, opened 2017-02-07
Resolves the current merge conflict...

standage:
...between `refactor/hashing2` and `master`.


standage:
I should note that the only file with a conflict was `lib/hashtable.cc`, and it was trivial to resolve.

PR #1617, opened 2017-02-13
Fix generation of coverage for c++ code

betatim:
Addresses https://github.com/dib-lab/khmer/pull/1511#issuecomment-279251128 where @ctb noticed that we do not generate coverage for c++ code anymore.

- [x] Is it mergeable?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,7 @@ clean: FORCE
 	rm -f diff-cover.html
 	rm -Rf build dist
 	rm -rf __pycache__/ .eggs/ khmer.egg-info/
+	-rm *.gcov
 
 debug: FORCE
 	export CFLAGS="-pg -fprofile-arcs -D_GLIBCXX_DEBUG_PEDANTIC \
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ notifications:
 # for a PR. So we disable build-on-push for all branches
 # except the ones whitelisted here.
 branches:
-  only: 
+  only:
     - master
 
 matrix:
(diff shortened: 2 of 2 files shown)


ctb:
Well, *something's* working, although it looks a bit weird :).

betatim:
So it seems like `make coverage-debug` thinks there is nothing to be done after `make test` has been run. However without recompiling with `-fprofile-arcs -ftest-coverage` etc you don't generate the files that `gcovr` needs to calculate the coverage.

Compare for example: https://codecov.io/gh/dib-lab/khmer/tree/32fcf25137937de2dc52d214724f87312434c51a and https://codecov.io/gh/dib-lab/khmer/pull/1617/tree one of them ignores all the files that aren't `.py`.

Makefile
```diff
@@ -136,6 +136,7 @@ clean: FORCE
 	rm -f diff-cover.html
 	rm -Rf build dist
 	rm -rf __pycache__/ .eggs/ khmer.egg-info/
+	-rm *.gcov
```
betatim: Is there a better way to pattern match these files that has less potential for accidentally deleting stuff that it shouldn't at some point in the future?

betatim:
If someone can think of a nicer `rm *gcov` pattern I'll do that, and if a second pair of eyes and brains could think about if there is a nicer way of doing this that would also be good. Otherwise I think we should merge. And then spend our time trying to dig ourselves out of this loss of coverage :(

Makefile
standage: The risk for inadvertently destroying unrelated files is extremely low.

PR #1620, opened 2017-02-14
Add missing header to build dependency

### LINKED ISSUE #1611: Look into stale object files/header files
@standage may have run into a problem with stale object files running an old version of a hash function.  We're not sure exactly what happened, but things to check --

* does `make clean` correctly remove all .o and .so files?
* does changing any header file result in recompilation of everything (as it seems), or are there new header files that have not been properly entered into setup.py?
* it seems unlikely that a virtualenv snafu could have led to this, but there are situations where multiple different installs can yield surprising imports based on how virtualenv installs stuff.
* it seems like setup.py installs stuff in the virtualenv even when we don't run `make install`. Is this intentional?

Perhaps we should create a debugging checklist with some sanity checks for when you think you might be going crazy:

* remove all installed khmer packages from your virtualenv;
* run tests on installed version (do we have docs for this?)

etc

betatim:
Potentially fixes #1611 

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -136,7 +136,7 @@ clean: FORCE
 	rm -f diff-cover.html
 	rm -Rf build dist
 	rm -rf __pycache__/ .eggs/ khmer.egg-info/
-	-rm *.gcov
+	-rm -f *.gcov
 
 debug: FORCE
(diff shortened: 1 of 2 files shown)


betatim:

PR #1622, opened 2017-02-16
[WIP] Add KarpRabin hash iterator

### LINKED ISSUE #1616: Explore @lemire's rolling hash function as an alternative for Counttable
See https://github.com/lemire/rollinghashcpp/blob/master/example.cpp

Should be much faster than murmurhash.

betatim:
Fixes #1616 

Exploring how to add RabinKarp hashing

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@
 
 # `SHELL=bash` Will break Titus's laptop, so don't use BASH-isms like
 # `[[` conditional expressions.
-CPPSOURCES=$(wildcard lib/*.cc lib/*.hh khmer/_khmer.cc khmer/*.hh) setup.py
+CPPSOURCES=$(wildcard lib/*.cc lib/*.hh lib/*.h khmer/_khmer.cc khmer/*.hh) setup.py
 PYSOURCES=$(filter-out khmer/_version.py, \
 	  $(wildcard khmer/*.py scripts/*.py oxli/*.py) )
 SOURCES=$(PYSOURCES) $(CPPSOURCES) setup.py
--- a/lib/characterhash.h
+++ b/lib/characterhash.h
@@ -1,0 +1,72 @@
+#ifndef CHARACTERHASH
+#define CHARACTERHASH
+
+typedef unsigned long long uint64;
+typedef unsigned int uint32;
+typedef unsigned int uint;
+
+#include <cassert>
+#include <iostream>
+#include <stdexcept>
+#include <random>
+
+using namespace std;
+
+
+
+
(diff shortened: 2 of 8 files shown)


betatim:
For the moment this is stalled because there is no `reverse_update` for Rabin Karp. We need to remove the last base and prepend a new base when computing the reverse complement in the iterator. `CyclingHash` supports this but `RabinKarp` doesn't :(

```
fwd: CGGTA
rev: TACCG
```
we'd like to update the hash from `h(ACCG)` to `h(TACC)`. Will switch to CyclingHash I think.

betatim:
A related question, do we want to add a new type `CycleCounttable` or replace the use of murmur hash? I think at least at the python level we should not add a new type but handle it via a `__init__` argument if we don't straight up replace murmurhash.

ctb:
agree with not adding a new type, i.e. handling it with __init__.  I think
we will want murmurhash as an option tho, since it's being used in other
places (the minhash world).

ctb:
we could also provide a single stranded option? or keep both strands
around in the iterator?

luizirber:
FYI: I was playing with something similar in the MurmurHashKmerIterator: https://github.com/dib-lab/khmer/compare/refactor/murmur_iterator

When I was benchmarking HLL `_revcomp` was by far what was taking more time to run (almost half of total runtime!), so the idea here is to update both the kmer being calculated and its revcomp base by base (instead of recalculating the revcomp for each kmer), and then calculate the hash value for both.

I had to change the implementation of `_hash_murmur_forward`  (because it was calling `_hash_murmur`, which always calls `_revcomp`...), and right now it is buggy (because I'm losing bases along the way? =P), but you can check the implementation here: https://github.com/dib-lab/khmer/compare/refactor/murmur_iterator#diff-8277863672a7ea5bfac461ec9cfea68dR282

ctb:
On Thu, Feb 16, 2017 at 11:12:32AM -0800, Luiz Irber wrote:
> FYI: I was playing with something similar in the MurmurHashKmerIterator: https://github.com/dib-lab/khmer/compare/refactor/murmur_iterator

sounds good to me!

betatim:
Thanks for the links. Computing both the forward and reverse hash of a kmer by adding and removing just one base to each is the goal, because as you say it uses a lot of CPU otherwise. And like you I have a bug somewhere right now :)

ctb:
Abandoned now that #1792 is merged, I think.

PR #1633, opened 2017-02-19
some bulk sequence loading tests that nail down current ACGTN behavior.

ctb:
These tests nail down behavior prior to #1590, which, when merged, will alter how we handle non-ACTG characters.  Note, no behavior is changed in this PR; it's just (lots of) new tests.

This explicitly puts in place tests for sequences that contain one of lowercase, Ns, and non-ACGTN characters, for:

* `consume_fasta` and all other bulk-sequence loading functions on Hashtables and derived classes;
* `trim_on_abundance`, `trim_below_abundance` and `find_spectral_error_positions`

This PR includes #1661.

Adds new test file `tests/test_sequence_validation.py` and data file `tests/test-data/valid-read-testing.fq`.

----

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


ctb:
Wow, [these tests](https://github.com/dib-lab/khmer/blob/c0c5d2cb77639a7dc163618ffdb910f36afe670d/tests/test_read_parsers.py#L487) are like a showcase of horrible inconsistency in loading sequences.

tests/test_read_parsers.py
```diff
@@ -483,5 +483,204 @@ def test_clean_seq():
...
+    x = Counttable(15, int(1e6), 4)
+    for read in ReadParser(infile):
+        x.consume(read.sequence)          # consume raw sequence
+
+    # the relevant read will be entirely ignored
+    # (b/c ReadParser does not uppercase)
+    kmer = "caggcgcccaccacc".upper()
+    assert x.get(kmer) == 1
+
+    # the 2nd read with this k-mer in it has an N in it; 'consume' will ignore.
```
betatim: I'd rephrase the comment. At the moment I am very puzzled after reading it (ah its being ignored so count should be one, but it is asserted to be two ...??)

betatim: Maybe "`consume` will ignore the invalid base and continue consuming the read, so this kmer after the N should have abundance 2"

betatim:
I'd move these tests to a new file maybe `test_sequence_loading.py`? Then we have one file that deals exclusively with verifying the assumptions we make about who cleans what, when and how.

Because it is easy to do I would parametrise the tests (that make sense) on the class so we test all combinations of Count/Node and table/graph. In which case we should get the parametrisation stuff from `test_tabletype.py` (or put these tests there, I'm -0)

tests/test_read_parsers.py
```diff
@@ -483,5 +483,204 @@ def test_clean_seq():
...
+    x.add_tag(kmer)
+    x.set_partition_id(kmer, 2)
+    
+    kmer = 'CCGGCGTG'
+    x.add_tag(kmer)
+    x.set_partition_id(kmer, 3)
+    
+    x.output_partitions(infile, savepath)
+
+    read_names = [ read.name for read in ReadParser(savepath) ]
```
betatim: The `pep8` :police_car: doesn't like the extra white space (hence the travis failure)

betatim:
+1 on merging this before 2.1, then #1590 can update the behaviour and fix these tests (remove the "in the future ..." comments)

tests/test_read_parsers.py
ctb: fixed in fe48787

tests/test_read_parsers.py
ctb: fixed 756454e

ctb:
Tests moved to `test_sequence_validation.py` and parameterized as per @betatim suggestion! Of course now a bunch of tests are failing... :)

betatim:
Linux build fails because of a pep8 violation. The OSX build fails because some tests fail and then we exit ungracefully because there are [too many open files](https://travis-ci.org/dib-lab/khmer/jobs/205207362#L3470). Will take a look at the latter.

betatim:
Ha, locally on OSX all tests pass as well and no "too many open files" warnings.

betatim:
Switched to using a fixture for `ReadParser` which allows us to explicitly close it after using it. Not sure this fixes the problem, but it is my best guess as I can't reproduce it locally.

betatim:
🎉

ctb:
This is now ready for review & merge (although #1661 should be merged first :)

@luizirber @betatim @standage

ctb:
(Upon merge, we should switch #1590 over to be against master.)

tests/test_sequence_validation.py
```diff
@@ -0,0 +1,337 @@
...
+        posns = x.find_spectral_error_positions(s, 1)
+    # assert posns == []                    # in future, should return [11]
+
+
+def test_read_cleaning_output_partitions(graphtype):
+    infile = utils.get_test_data('valid-read-testing.fq')
+    savepath = utils.get_temp_filename('foo')
+
+    # read this in using "approved good" behavior w/cleaned_seq
+    x = _Nodegraph(8, PRIMES_1m)
```
betatim: -> `graphtype`?

tests/test_sequence_validation.py
```diff
@@ -0,0 +1,337 @@
...
+    return request.param
+
+
+# all the graph types!
+@pytest.fixture(params=[_Countgraph, _Nodegraph])
+def graphtype(request):
+    return request.param
+
+
+@pytest.fixture
```
betatim: Now that I've looked this up: can we change this back to `yield` and change the decorator to `@pytest.yield_fixture`? Link to the pytest 2.9 docs: http://doc.pytest.org/en/2.9.2/fixture.html#fixture-finalization-executing-teardown-code

ctb:
done

ctb:
all comments addressed in e419a7b

betatim:
:tada:

Issue #1528, opened 2016-11-15
Look into caching OS X Miniconda install on Travis CI


standage:
Travis CI allows you to cache components/directories created as part of your build. This won't improve the availability of OS X VMs, but once an OS X VM is available this should speed up the build significantly.

standage:
Also, is there any reason to run extra tests/builds on OS X vs Linux? Since we always seem to be waiting on  OS X VMs, perhaps we could move these to one of the Linux VMs?

betatim:
We spend about 3minutes on installing things on OSX.

The [travis docs](https://docs.travis-ci.com/user/caching/) suggest not caching stuff which is dominated by "downloading stuff from the network" as the cache tarball comes over the network as well.

From a quick browse of the logs it seems we don't really compile much stuff, which suggests we are waiting on the network. I will take a look how complex it would get to cache the various different directories.

We use the OSX build as the reference for coverage and formatting. `make test` takes 155s. To compute the coverage we recompile things and re-execute the tests (120s). Can we combine these two steps?

(The availability of OSX nodes is pretty shocking, especially during California day time)

standage:
Yeah, I think I was off base with my initial comment about caching. The Miniconda install is a relatively small contributor to the long runtime. I think it's mostly the multiple builds. Not sure why everything is recompiled from scratch to compute coverage, etc. Now I'm just wondering why OS X is used as reference, and whether it could save us some work-hours backlogging if we moved this to one of the Linux VMs.

ctb:
On Wed, Nov 16, 2016 at 09:38:27AM -0800, Daniel Standage wrote:
> Yeah, I think I was off base with my initial comment about caching. The Miniconda install is a relatively small contributor to the long runtime. I think it's mostly the multiple builds. Not sure why everything is recompiled from scratch to compute coverage, etc. Now I'm just wondering why OS X is used as reference, and whether it could save us some work-hours backlogging if we moved this to one of the Linux VMs.

+1

betatim:
> Now I'm just wondering why OS X is used as reference

More or less arbitrary. The reason that tipped it was that the version of `make format` on linux and on OSX disagree on what the correct format should be. So I guessed that most people use OSX locally -> make osx the reference. The way forward would be to figure out how to install the OSX/newer version of the formatting tool on debian (I spent some time on this when we moved to travis but gave up, my debian-fu was not strong enough).

standage:
ping @betatim

betatim:
[...]

PR #1639, opened 2017-02-23
Switch coverage computation to linux

betatim:
Let's see what happens if we compute coverage on linux instead of OSX. Related to #1528.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ matrix:
       dist: trusty
       env:
         - TESTATTR="'not huge and not known_failing'"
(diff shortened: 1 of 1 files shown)


betatim:
So even with this setup codecov will not post the coverage as a comment here until the OSX build is done. Kinda suboptimal.

betatim:
It does not seem like this gains us anything in getting faster turn around so closing.

PR #1642, opened 2017-02-24
Updated docs for setting memory

### LINKED ISSUE #1558: Update docs on setting memory params
https://khmer.readthedocs.io/en/latest/user/choosing-table-sizes.html should probably be updated to refer to the new 'unique-kmers.py' script and the -U parameter to load-into-counting that sets memory based on the # of unique kmers.

standage:
- switched from scientific notation to human-readable
- added section for `unique-kmers.py` and `-U` option (closes #1558)
- updated diginorm FPR threshold from 10% to 20%

Any comments on this last point? The default threshold for normalize-by-median is 0.2, and it seems like we either need to update the docs to reflect this or change the threshold if the docs are correct. Unless I'm missing somthing.

- [x] Is it mergeable?
  changes were made?

### DIFF
--- a/doc/user/choosing-table-sizes.rst
+++ b/doc/user/choosing-table-sizes.rst
@@ -4,24 +4,24 @@
    Copyright (C) 2015 The Regents of the University of California.
    It is licensed under the three-clause BSD license; see LICENSE.
    Contact: khmer-project@idyll.org
-   
+
    Redistribution and use in source and binary forms, with or without
    modification, are permitted provided that the following conditions are
    met:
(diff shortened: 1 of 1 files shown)


standage:
Ready for a review @betatim @ctb @luizirber @camillescott.

standage:
I think we'd better wait for @ctb to comment on the FPR threshold before merging.

ctb:
FPR update is fine. @qingpeng actually showed that it works up past 80% ;).

Issue #1514, opened 2016-11-14
Check examples and lib/Makefile regularly


ctb:
add examples/c++-api and lib/Makefile to our continuous integration, maybe?

(the former uses the latter, note)

betatim:
What is `lib/Makefile` used for? been wondering that for a while but never asked...

ctb:
On Thu, Nov 17, 2016 at 04:25:31AM -0800, Tim Head wrote:
> What is `lib/Makefile` used for? been wondering that for a while but never asked...

building liboxli for people who want to use the C++ non-API.

standage:
See #1364. We could easily add a few simple examples to test the most commonly used components.

PR #1643, opened 2017-02-24
Fix stamps example and add it to travis runs

betatim:
A proposal for #1514

Seems pretty straight forward and a good test of sticking to the CLI API. Let's see how much it increases the travis run time.

- [x] Is it mergeable?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ script:
   - make test
   - make pep8
   - make doc
+  - cd examples/stamps && ./do.sh
 
 # generate all the diagnostic reports
 after_success:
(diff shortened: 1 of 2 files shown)


betatim:
Increases the time for a linux build to run to about 9min. What do people think? Is it worth it?

(Clean it up a bit so that we don't see all that output from its run)

standage:
Well, given that we were waiting 2hrs for the OS X VMs to become available at peak time this afternoon, 9 additional minutes hardly seems a big ask.

betatim:
I edited `do.sh` so "make it pass" and didn't spend any time trying to think about whether what it is (now) doing makes sense from a science perspective. Would be good if someone could take a look.

standage:
I'll let you update this one @betatim since it's on your fork.

betatim:
Ready for review  @standage

PR #1648, opened 2017-02-25
Fix "make pydocstyle" so that versioneer code is ignored

### LINKED ISSUE #1531: Have 'make pydocstyle' ignore versioneer stuff
The `make pydocstyle` target currently whines about the `khmer/_version.py` file, which is installed 3rd-party code, not ours. We should make sure to ignore this file so this target doesn't spew noise that is annoying for devs and potentially misleading for newcomers (who are asked to `make pydocstyle` with their first PR).

standage:
Closes #1531. Good news: this PR fixes the bug so that versioneer code is ignored. Bad news: the default value of the `match` parameter (see [pydocstyle docs](http://www.pydocstyle.org/en/latest/usage.html)) means that even though we were ostensibly checking the test code before, no issues were reported for any files beginning with `test_` (i.e. all of the test code). More to clean up!

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?


betatim:
To be merged or not ready yet?

standage:
My initial thought was to handle the pydocstyle errors in this PR, but maybe that's unnecessary. What do you think?

betatim:
I'm ok with the tests not being subject to pydocstyle.

Fixing all the errors would be great. Considering it is quite a bit of work I would say it is only worth it if we have support to make `make pydocstyle` compulsory for future PRs. Otherwise we will end up in a situation like this again soon.

standage:
See #1652. I'm +1 for handling this _at some point_, I'm -1 on that _some point_ being before the 2.1 release.

Ready for review and merge!

betatim:
I would restore the behaviour of `test_blah.py` being excluded, otherwise LGTM.

standage:
I agree. I don't think the docstyle of the tests is critical to...anything, really. The `tests/` directory was explicitly added to the target recipe, so someone at some time thought it was important. I'll give everyone a day or two to comment on this if they have any strong feelings before removing `tests/`. cc @ctb @luizirber @camillescott @betatim

Issue #1646, opened 2017-02-24
Make pytest work for pip installs


standage:
#1403 notes that we need to update our release checklist with respect to the nosetests/pytest replacement. We also need to update the install docs. More importantly, things don't currently seem to work properly, so additional configuration may be required.

```
$ pip install git+https://github.com/dib-lab/khmer.git
...
blah blah blah
...
$ pytest --pyargs khmer -m 'not known_failing and not jenkins and not huge and not linux'
=========================================================== test session starts ============================================================
platform darwin -- Python 3.5.2, pytest-3.0.4, py-1.4.31, pluggy-0.4.0
rootdir: /Users/standage, inifile:
plugins: cov-2.4.0
collected 0 items 

======================================================= no tests ran in 0.01 seconds =======================================================
$
```

standage:
Replacing `khmer` with `khmer.tests` results in the following error.

```
ERROR: file or package not found: khmer.tests
```

betatim:
For me `pip install git+https:/.....` in a new env leads to:
```
$ pytest -s --pyargs khmer -m 'not known_failing and not jenkins and not huge and not linux'                                                    (khmer-rc)
=================================================================== test session starts ===================================================================
platform linux -- Python 3.5.3, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /tmp, inifile: 
collected 1255 items / 1 errors 

========================================================================= ERRORS ==========================================================================
____________________________________________________________________ ERROR collecting  ____________________________________________________________________
Using pytest.skip outside of a test is not allowed. If you are trying to decorate a test function, use the @pytest.mark.skip or @pytest.mark.skipif decorators instead.
!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!! Interrupted: 1 errors during collection !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
=================================================================== 25 tests deselected ===================================================================
========================================================= 25 deselected, 1 error in 0.64 seconds ==========================================================
```

betatim:
If I create a fresh environment, clone master myself, `make install`, change directories to somewhere outside the repository then:
```
$ pytest --pyargs khmer.tests.test_sandbox_scripts -m 'not known_failing and not jenkins and not huge and not linux' 
```
[...]

PR #1649, opened 2017-02-27
Fix the skipping of sandbox script tests

betatim:
scripts in the sandbox will only be tested if the tests are executed
from within the source code repository. Use `skipif` to skip a test instead
of having an exception being raised that then confuses `runscript()`.

Related to #1646 

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?


betatim:
In a new env:
```
$ pip install git+https://github.com/dib-lab/khmer@fix/sandbox-script-testing
$ pytest -s --pyargs khmer.tests.test_sandbox_scripts::test_sweep_reads
```
now succeeds. As in skips the test.

betatim:
@standage could you try out those two lines to see if this also makes it work for you? Not sure I have a good idea why it would work on linux but not OSX.

betatim:
Related question, should we add to travis a step of "install this as a module and re-run the tests" to check this doesn't break again in the future?

ctb:
note, it's part of the release checklist

betatim:
Exactly, trying to think of a reason why we shouldn't automate it. One less thing that comes as an unpleasant surprise when going through the checklist.

ctb:
it would add a reasonable amount of compute time, no? I guess I'm -0 on it
for that reason alone, but that's a weak -0.

betatim:
Probably would add ~3min to something that takes 11min to run (plus the wait for a OSX instance which rules us all).

---

Ready for review! @luizirber, @camillescott, @standage , or @ctb In particular if someone can confirm that the two steps above work locally on OSX.

PR #1651, opened 2017-02-28
`make doc` fails on travis

### LINKED ISSUE #1647: Travis is giving green lights even when the "make doc" target fails
...and it's failing on all recent builds as far as I can see, Linux and OS X.

betatim:
Fixes #1647 For some reason `make doc` fails on travis but not locally. Trying to find out why.

- [x] Is it mergeable?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,6 +66,7 @@ install:
 script:
   - make test
   - make pep8
+  - make doc
 
 # generate all the diagnostic reports
 after_success:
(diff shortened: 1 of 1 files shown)


betatim:
Moved `make doc` to be a required part of the build. This means we won't have the situation happen again where there was a problem in the build and we didn't notice for a while. As a side effect the problem of the docs not building seems to have gone away :man_shrugging:

betatim:

betatim:
My money was on some left over build artefact from computing the coverage, but couldn't confirm that locally :-/

PR #1653, opened 2017-03-01
[MRG] Add function to calculate Nodegraph size matched to a Countgraph

### LINKED ISSUE #1408: Weird memory usage in abundance-dist-single.py
Using the command:

```
abundance-dist-single.py -M 6e9 -b 7086_S82_L007.abundtrim.fq.gz 7086_S82_L007.abundtrim.fq.gz.hist -k 21 -s
```

I'm seeing 10 GB+ of RAM being used (see screenshot from top). More than 6 GB was requested and failed previously; this is on a 15 GB of RAM computer.

screen shot 2016-07-18 at 9 53 18 am

betatim:
Fixes #1408 

Sometimes you want a Nodegraph with roughly the same number of
buckets as a Countgraph you are using.

Is having the number of buckets in the `Nodegraph` match those of the `Countgraph` actually what people want?

Still needs a bit of fine tuning/understanding how/why the sizes differ slightly in the two cases (`-x` vs `-M`).

There is also the option that we interpret `-M` as the absolute upper total memory used by everything in this script. In which case we need to be a bit smarter when calculating table sizes as they have to take into account what comes later in the script. Essentially making it specific for each script :-/

- [x] Implement this for all scripts that use more than one Countgraph/table
- [x] Is it mergeable?
  `scripts/` is it tested?
- [ ] `make format diff_pylint_report doc pydocstyle` Is it well
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -569,6 +569,27 @@ def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1):
         return khmer.Countgraph(ksize, tablesize, args.n_tables)
 
 
+def create_matching_nodegraph(args, multiplier=1.0):
+    """Create a Nodegraph matched in size to a Countgraph
+
+    Use this to create Nodegraphs for kmer tracking and similar. The
+    created Nodegraph will have a similar number of buckets in its
+    tables as your Countgraph.
+    """
+    tablesize = 0
+    # the use of 'countgraph' and 'smallcountgraph' is on PURPOSE
+    # as we want to find out how big the corresponding countgraphs
+    # would be so we can make a matching Nodegraph
+    if args.small_count:
+        tablesize = calculate_graphsize(args, 'smallcountgraph',
+                                        multiplier=multiplier)
+    else:
+        tablesize = calculate_graphsize(args, 'countgraph',
+                                        multiplier=multiplier)
+    print('NG tablesize:', tablesize)
+    return khmer.Nodegraph(args.ksize, tablesize, args.n_tables)
+
+
 def report_on_config(args, graphtype='countgraph'):
     """Print out configuration.
 
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -510,8 +510,8 @@ def calculate_graphsize(args, graphtype, multiplier=1.0):
         max_memory = args.max_memory_usage
         if max_memory > 1e8:
             max_memory -= 3e7
-        tablesize = (khmer._buckets_per_byte[graphtype] *
-                     max_memory / args.n_tables / float(multiplier))
+        tablesize = float(multiplier) * (khmer._buckets_per_byte[graphtype] *
+                                         max_memory / args.n_tables)
     else:
         tablesize = args.max_tablesize
 
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -504,8 +504,10 @@ def calculate_graphsize(args, graphtype, multiplier=1.0):
         raise ValueError('unknown graph type: ' + graphtype)
 
     if args.max_memory_usage:
-        tablesize = float(multiplier) * (khmer._buckets_per_byte[graphtype] *
-                                         args.max_memory_usage / args.n_tables)
+        # subtract about 30MB for the python interpreter
+        max_memory = args.max_memory_usage - 3e7
+        tablesize = (khmer._buckets_per_byte[graphtype] *
+                     max_memory / args.n_tables / float(multiplier))
     else:
         tablesize = args.max_tablesize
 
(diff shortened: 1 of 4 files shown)


ctb:
What about extracting `hashsizes()` from the `Countgraph` and then using passing those directly into `_Nodegraph`?

betatim:
I had thought about that but ditched it because it means using `_Nodegraph` in a script. Now that you mention it though we could handle that like this:
```
def create_matching_nodegraph(countrgraph):
  return _Nodegraph(..., countgraph.get_hashsizes())
```
so it is all nicely hidden away. I think I like that even more than what I first proposed here.

From a short slack conversation with @ctb the conclusion is that `-M` should set the total upper amount of memory a process uses. So not giving everything to the main data structure and then using "extra" memory for the supporting stuff. Instead we need to work out up front who needs how much. Also we should turn all the `multipliers` to `1.` when `-M` is passed in.

khmer/khmer_args.py
```diff
@@ -504,8 +504,14 @@ def calculate_graphsize(args, graphtype, multiplier=1.0):
...
     if args.max_memory_usage:
+        # subtract about 30MB for the python interpreter if user is requesting
+        # a large amount of memory that suggests they want to limit the
+        # total amount of RAM used for purposes of not running out of memory
+        max_memory = args.max_memory_usage
+        if max_memory > 1e8:
+            max_memory -= 3e7
         tablesize = (khmer._buckets_per_byte[graphtype] *
-                     args.max_memory_usage / args.n_tables / float(multiplier))
+                     max_memory / args.n_tables / float(multiplier))
```
betatim: @standage why is it called a multiplier but then divides? Is there something I am missing? I was expecting that passing `multiplier=1.1` would increase the amount of memory used by 10% but it makes it use less.

khmer/khmer_args.py
```diff
@@ -504,8 +504,14 @@ def calculate_graphsize(args, graphtype, multiplier=1.0):
         raise ValueError('unknown graph type: ' + graphtype)
 
     if args.max_memory_usage:
+        # subtract about 30MB for the python interpreter if user is requesting
+        # a large amount of memory that suggests they want to limit the
+        # total amount of RAM used for purposes of not running out of memory
+        max_memory = args.max_memory_usage
+        if max_memory > 1e8:
+            max_memory -= 3e7
```
betatim: What do you think of this? If we really want to keep the total memory below the `-M` limit we need to subtract some memory for "the rest". This works if the max memory requested is large, but falls over if you set a limit of "10MB" or so. My guess is that in those cases the user is either testing or doing it for some reason that isn't limiting the true max memory used.

khmer/khmer_args.py
standage: Bucket size multiplier maybe? The multiplier parameter certainly predates any of my hacking on memory related arguments.

ctb:
I think this was my fault, back when. Doesn't matter. Just make it make
sense and work right :)

khmer/khmer_args.py
standage: Hmm, I'm not sure about this. I think the behavior is good and necessary, and even though it will involve more code I'd suggest this might be better handled on a script level (with properly documented motivations). I think the library code's responsibility is to give the programmer what she asks for.

Something else to consider would be making `1e8` and `3e7` configurable parameters. I'd suggest `1e9` would be a better default for the first.

ctb:
The problem is that our calculation of tablesize etc is kind of opaque
due to all of the conflicting command line parameters.  This is a rare
case and we can special-case it IMO.

standage:
> This is a rare case and we can special-case it IMO.

If you're talking about `abundance-dist-single.py` I agree. My last comment was about a change to the `khmer_args` module, which would affect all scripts.

betatim:
It is true that this is the only instance where this is needed.

So we reduce the multiplier a bit by using 8/(9+epsilon) instead of 8/9. Is what I'd do I think.

standage:
> So we reduce the multiplier a bit by using 8/(9+epsilon) instead of 8/9.

👍

standage:
Looking better after the latest commits IMO.

betatim:
Close but no cigar? 😛 

I am happy with it as it is, so let me know what/where we should tweak stuff. Or we merge?

standage:
No, looks good to me. I jumped in to the conversation because I had something to add, but never saw a request for a review. Will merge now.

khmer/khmer_args.py
```diff
@@ -504,8 +504,8 @@ def calculate_graphsize(args, graphtype, multiplier=1.0):
         raise ValueError('unknown graph type: ' + graphtype)
 
     if args.max_memory_usage:
-        tablesize = (khmer._buckets_per_byte[graphtype] *
-                     args.max_memory_usage / args.n_tables / float(multiplier))
+        tablesize = float(multiplier) * (khmer._buckets_per_byte[graphtype] *
```
standage: This makes more sense now.

betatim:
The new test looks good to me. Sorry for not getting to this earlier.

PR #1654, opened 2017-03-01
restore cppcheck

mr-c:
Sorry for being late to the `cppcheck` party. Here is what it took to fix it on my Debian system. Could a Mac user test?

- [x] Is it mergeable?
- ~~`make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?~~
- ~~`make format diff_pylint_report doc pydocstyle` Is it well
  formatted?~~
- ~~Did it change the command-line interface? Only backwards-compatible
  formats also requires a major version number increment.~
- ~~For substantial changes or changes to the command-line interface, is it
  for more details.~~
- ~~Was a spellchecker run on the source code and documentation after
  changes were made?~~
- ~~Do the changes respect streaming IO? (Are they
  tested for streaming IO?)~~


betatim:
Not tried it locally but [travis says it runs](https://travis-ci.org/dib-lab/khmer/jobs/206584984#L4535). Do you know if that last message "(information) Cppcheck cannot find all the include files (use --check-config for details)" means cppcheck gave up or it informed us and then did lots of checking?

@standage do you have some details what it was that made it broken on your setup?

mr-c:
@betatim that message is due to excluding the analysis of seqan as a speedup, see `make cppcheck-long`

You could have TravisCI run the long version @ https://github.com/dib-lab/khmer/blob/master/.travis.yml#L80 so that this exclusion doesn't hide other potential issues

mr-c:
One could also run `make cppcheck-long` on Travis w/ OSX too: 
OSX side too: https://github.com/dib-lab/khmer/blob/master/.travis.yml#L49

betatim:
`make cppcheck` runs on my Mac.

betatim:
My usual question: do the regular committers like cppcheck enough for us to make it a mandatory check?

Makefile
```diff
@@ -49,8 +49,21 @@ GCOVRURL=git+https://github.com/nschum/gcovr.git@never-executed-branches
...
 	| sed 's/+/-/')
 
+DEFINES += -DNDEBUG -DVERSION=$(VERSION) -DSEQAN_HAS_BZIP2=1 \
+	   -DSEQAN_HAS_ZLIB=1 -UNO_UNIQUE_RC
+
 INCLUDESTRING=$(shell gcc -E -x c++ - -v < /dev/null 2>&1 >/dev/null \
 	    | grep '^ /' | grep -v cc1plus)
+INCLUDEOPTS=$(shell gcc -E -x c++ - -v < /dev/null 2>&1 >/dev/null \
+	    | grep '^ /' | grep -v cc1plus | awk '{print "-I" $$1 " "}')
+PYINCLUDE=$(shell python2.7-config --includes)
```
betatim: We should switch to something that works with python as well as legacy python. While looking into this I also realised that I don't know how to make this tell me the headers that match the python version in my virtual/conda env

standage:
I'm +0.5 on having a `make cppcheck` target, -1 on making it a mandatory part of the build. Anything python2 or python3 would be a concern.

Makefile
mr-c: Fixed, thanks for catching that

mr-c:
@betatim Yay, glad it works on your mac

My "I don't work here anymore, but opinions are free, right?" view:

`cppcheck` still should be run by the CI (preferably the `cppcheck-long` version)
New warnings & errors should fail builds

I'm fine if it doesn't make it on the checklist

Cheers,

betatim:
Annoyingly `python-config` doesn't actually pick up the one from my conda env, because that only has `python3-config` ... not sure if this is conda's fault or what.

What do you think of using something like this instead:
```
$ python -c "from __future__ import print_function; import sysconfig; flags = ['-I' + sysconfig.get_path('include'), '-I' + sysconfig.get_path('platinclude')]; print(' '.join(flags))"
```
which will definitely pickup "the current python". Does mean duplicating code from python-config ...

mr-c:
@betatim I would say that is conda's fault, unless they allow mixed py3/py2 conda envs (but then they still should have a fallback). 

Here `virtualenv` does the right thing: all python{,-*} point to whatever version of python the virtualenv was told to build with 

Seems like a fair workaround; wouldn't hurt to file a bug with the conda people either

Can I leave these items to you?
- [ ] decide on workaround, if any, for conda py3 issue
- [ ] choose between `make cppcheck` and `make cppcheck-long` for [travisCI's invocation](https://github.com/dib-lab/khmer/blob/master/.travis.yml#L80)

standage:
The `make cppcheck` target is now working on my system, but I get the following output.

```
(env) » make cppcheck
[/Users/standage/Software/kenv/include/python3.5m/Python.h:18]: (error) #error "Python's source code assumes C's unsigned char is an 8-bit type."
[/Library/Developer/CommandLineTools/usr/include/c++/v1/type_traits:3791]: (error) #endif without #if
[/Library/Developer/CommandLineTools/usr/include/c++/v1/cstddef:102]: (error) #endif without #if
(information) Cppcheck cannot find all the include files (use --check-config for details)
```

Running with the `--check-config` flag makes the last line go away, but doesn't change anything else. Additional suppressions needed? I'm on OS 10.12.3.

standage:
Also, even though these errors are reported, `make check` returns status 0. If we *really do* want this to break the build, it needs to return non-zero on error.

mr-c:
Oh Apple, you are almost POSIX :-P

@standage You can add two Mac specific blanket suppression like https://github.com/dib-lab/khmer/pull/1654/files#diff-b67911656ef5d18c4ae36cb6741b7965R65
`--suppress='*:/Library/*'` and `--suppress='*:/Users/*/Software*'`

In the Jenkins days we relied upon `make cppcheck-result.xml` which Jenkins knew what to do with
https://github.com/dib-lab/khmer/pull/1654/files#diff-b67911656ef5d18c4ae36cb6741b7965R161

I think the reason cppcheck didn't exit with non-zero was because the source of error wasn't in any of the explicit input files.

standage:
The `--suppress='*:/Library/*'` flag should work for all Macs, but the `/Users/standage/Software/` suppression doesn't make sense. The `Software` directory is a personal organizational choice, not a standard fixture in Mac home directories (`/Users/` is the Mac / BSD equivalent of `/home/`, and `kenv` is simple a virtual environment in a directory called `Software` that I myself created).

That leaves the following error with the virtualenv-installed Python.

```
[/Users/standage/Software/kenv/include/python3.5m/Python.h:18]: (error) #error "Python's source code assumes C's unsigned char is an 8-bit type."
```

betatim:
Can we add `--suppress='*:*/include/python3*/Python.h'` to match all python3 `Python.h`? Maybe even expand it to all versions of python?

Makefile
```diff
@@ -163,7 +165,7 @@ cppcheck-result.xml: $(CPPSOURCES)
 
 ## cppcheck    : run static analysis on C++ code
 cppcheck: FORCE
-	@$(CPPCHECK)
+	@$(CPPCHECK) || exit 1
```
standage: This should be redundant with `--error-exitcode=1` amirite?

betatim: Yes you are. Left over from fiddling around.

mr-c:
@standage @betatim Ready to merge

PR #1655, opened 2017-03-01
Add infrastructure for fixed random seeds in tests

betatim:
Proposal for fixing random seeds in tests to make them reproducible.

The basic idea is that if you need randomness, you call `check_random_state(seed)` and that you expose the `seed` argument to your callers. If they don't care they can not pass anything in, in which case `seed=None` and we get random numbers that aren't reproducible. If they do care they can either pass in a number or an existing random number generator.

@standage this is what I had in mind. Shamelessly cribbed from http://scikit-learn.org/stable/modules/generated/sklearn.utils.check_random_state.html

We also use this approach in scikit-optimize.

- [ ] Is it mergeable?
  `scripts/` is it tested?
- [ ] `make format diff_pylint_report doc pydocstyle` Is it well
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/tests/test_assembly.py
+++ b/tests/test_assembly.py
@@ -39,12 +39,11 @@
 from __future__ import print_function
 from __future__ import absolute_import
 
-import itertools
+import numbers
 import random
 
 import khmer
 from khmer.khmer_args import estimate_optimal_with_K_and_f as optimal_fp
-from khmer import ReadParser
 from khmer import reverse_complement as revcomp
 from . import khmer_tst_utils as utils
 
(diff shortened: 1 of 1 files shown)


standage:
Got it.

So at some point we need to print out the random seed being used so that when tests fail we can reproduce their behavior, right?

betatim:
Not really. If we do everything right in writing the tests they always use the same seed on all the platforms. The only time things change is when you reorder/change number of calls to the RNG. Though usually you don't do that for debugging a failure. Alternatively you can go in and fix the seeds for the intermediate steps.

(I've seen it done that you provide different seeds as parameters to a `test_blah(seed)` method if you wanted to run the same test for different seeds. Though I would argue that the value of the seed shouldn't change the outcome of your test)

betatim:
If you like it, I will spread the approach further around the tests.

standage:
> If we do everything right in writing the tests they always use the same seed on all the platforms.

Ah, I see. I misunderstood.

In contrast, the GenomeTools library (to which I was previously a frequent contributor) has the CI generate and report a new random seed for each build. This way it's are testing the overall robustness of the randomized procedures, not just for a single seed. It was pretty rare, but every once in a while one of the tests would fail on an edge case, and the seed would enable us to track down exactly what happened.

Of course, this often requires a different approach to evaluating the output, so I don't know how relevant it would be here, especially in the short term.

PR #1658, opened 2017-03-15
[MRG] C++ and Python demos

### LINKED ISSUE #1514: Check examples and lib/Makefile regularly

standage:
The current master has a single C++ demo. This PR will expand on this a bit, adding additional C++ and Python demos and integrate this into the CI build.

Closes #1514.

- [x] Is it mergeable?
  `scripts/` is it tested?
- [x] `make format diff_pylint_report doc pydocstyle` Is it well
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


examples/c++-api/Makefile
```diff
@@ -1,10 +1,22 @@
-all: count-demo
+CFLAGS=--std=c++11 \
+       -I ../../lib \
+	   -I ../../third-party/smhasher \
+	   -I ../../third-party/seqan/core/include/
```
betatim: Is there a way (without too much pain) to make the example so that it doesn't use relative paths and relies on having run `make` fror khmer before? Thinking that as a real user I probably have my code in a different directory and want to somehow use the "installed" version of khmer instead of having my stuff in a subdirectory of khmer.

standage: Maybe, but not a priority here. These examples are provided by way of exposition, and even though I'm adding them to the build here the primary purpose is not for the programs to be run but for the code to be digested by a programmer who wants to use khmer.

standage:
Ready for review and merge @betatim @ctb @camillescott @luizirber.

examples/c++-api/exact-counting.cc
```diff
@@ -19,23 +20,21 @@ int main()
     // If `ksize` is even, note that k-mers will collapse with their reverse
     // complement. In that case, a table size of 4**(k-1) + k is required.
 
-    std::vector<HashIntoType> tablesize;
-    tablesize.push_back(pow(4, ksize));
+    std::vector<uint64_t> tablesize = {nkmers};
```
betatim: c++11 FTW!

Makefile
```diff
@@ -381,6 +381,16 @@ list-citation:
 		> authors.csv
 	python sort-authors-list.py
 
+## cpp-demos   : run programs demonstrating access to the (unstable) C++ API
+cpp-demos:
```
betatim: could you make these depend on the shared library build step? Just tried to run it locally on a fresh checkout and you get a "weird" error message. 
`cpp-demos:` -> `cpp-demos: sharedobj` should work I think

betatim:
LGTM after the `Makefile` tweak

examples/python-api/exact-counting.py
```diff
@@ -0,0 +1,24 @@
...
+
+cg.count('ATGGCA')
+cg.count('ATGGCA')
+cg.count('ACATGG')
+cg.count('AAAAAA')
+cg.count('TTTTTT') # this will be counted towards AAAAAA
+
+# The reverse_hash function is only available for Countgraph and Nodegraph,
+# not Counttable and Nodetable.
+for i in range(0, nkmers):
```
betatim: -> `range(nkmers)`

standage: Done

Issue #975, opened 2015-04-16
write test to ensure that output from "trim-low-abund --out -" is in good format


ctb:
In #963, test_trim_low_abund_stdout() should actually parse the output, because a substring mismatch may happen even if there's other garbage in the output.

ctb:
Resolution from talking with @mr-c: will probably add a JSON output flag, as per `load-into-counting`. Punting to 2.0+.

Issue #974, opened 2015-04-16
trim-low-abund no longer distinguishes between types of information output


ctb:
in #963, trim-low-abund was modified to no longer separate diagnostic output from informative output (output that should be saved vs output that merely indicates things are running); this was because the -o/--out option was added.  We need to add this behavior back in somehow, because it's important for tracking specific information for publications and post-analysis (see https://github.com/ged-lab/2014-streaming/).

I see at least two options --

one is to track this info internally and optionally output a JSON or CSV summary file;
another is to split the output in two, one for "runtime info" and the other for "stats", and by default have them both go to stderr, but provide options to redirect the stats.

I think the first approach (JSON/CSV summary file) might be the best, but open for discussion.

ctb:
See #975.

PR #1674, opened 2017-04-13
[MRG] Add support for run time info to json or tsv in trim-low-abund

betatim:
Related to #974 and #975. This lets the user output a file recording information about the run to a JSON or TSV file.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/scripts/trim-low-abund.py
+++ b/scripts/trim-low-abund.py
@@ -521,7 +521,8 @@ def main():
         ct.save(args.savegraph)
 
     if args.summary_info is not None:
-        # when the output is stdout the name is == 1
+        # note that when streaming to stdout the name of args.output will
+        # be set to 1
         if args.output is not None and args.output.name != 1:
             base = args.output.name
         # no explicit name or stdout stream -> use a default name
--- a/scripts/trim-low-abund.py
+++ b/scripts/trim-low-abund.py
@@ -50,6 +50,7 @@ import os
 import json
 import khmer
 import tempfile
+import time
 import shutil
 import textwrap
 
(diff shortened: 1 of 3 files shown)


scripts/trim-low-abund.py
```diff
@@ -483,6 +514,24 @@ def main():
         log_info("Saving k-mer countgraph to {graph}", graph=args.savegraph)
         ct.save(args.savegraph)
 
+    if args.output is not None and args.output.name != 1:
+        base = args.output.name
+    # no explicit name or stdout stream get a default name
+    else:
+        base = 'trim-low-abund'
```
betatim: how should we pick the filename when the user did not specify an output filename (and there are multiple input files) or when the output is redirected to stdout?

standage: I'd suggest something with the script name as well as an ISO 8061 compatible timestamp in the log file name.

betatim:

scripts/trim-low-abund.py
```diff
@@ -488,6 +520,28 @@ def main():
         log_info("Saving k-mer countgraph to {graph}", graph=args.savegraph)
         ct.save(args.savegraph)
 
+    if args.summary_info is not None:
+        # when the output is stdout the name is == 1
+        if args.output is not None and args.output.name != 1:
```
standage: This comment is redundant at best, misleading at worst. Comment says `== 1`, code says `!= 1`.

tests/test_scripts.py
```diff
@@ -2565,9 +2563,52 @@ def test_trim_low_abund_trimtest_savegraph():
...
+    args = ["-k", "17", "-x", "1e7", "-N", "2", "--summary-info", "tsv",
+            "-o", "summary", infile]
+    _, out, err = utils.runscript('trim-low-abund.py', args, in_dir)
+
+    summary_fname = os.path.join(in_dir, "summary.info.tsv")
+    assert os.path.exists(summary_fname), summary_fname
+    with open(summary_fname) as f:
+        reader = csv.DictReader(f, dialect='excel-tab')
+        lines = [row for row in reader]
+        assert len(lines) == 1
```
standage: Tests look good.

scripts/trim-low-abund.py
betatim: Reworded it, what I am trying to do is leave a hint as to why `1` is a special name/value. Basically trying to explain the weird condition of the if which is "if we have an explicit name for the output use it unless that name is 1 which is the magic value of stdout"

standage: Great. Feel free to merge when the build is done, I'm about to hit the sack.

PR #1667, opened 2017-04-05
Try out the Counting Quotient Filter as a replacement? for Bloom filter/CMS.

ctb:
Ref https://github.com/dib-lab/khmer/issues/1665#issuecomment-291860432.

Current status: this compiles on Titus's laptop, but no CQF code is yet used :)

- [ ] Is it mergeable?
- [ ] `make test` Did it pass the tests?
- [ ] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?
- [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
  formatted?
- [ ] Did it change the command-line interface? Only backwards-compatible
  additions are allowed without a major version increment. Changing file
  formats also requires a major version number increment.
- [ ] For substantial changes or changes to the command-line interface, is it
  documented in `CHANGELOG.md`? See [keepachangelog](http://keepachangelog.com/)
  for more details.
- [ ] Was a spellchecker run on the source code and documentation after
  changes were made?
- [ ] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)

PR #1675, opened 2017-04-18
[MRG] CQF based storage

betatim:
This is a PR on top of #1667 

With this the mechanics of calling:
```python
import khmer
cqf = khmer.QFCounttable(3, ...)
cqf.add("ACG")
cqf.get("ACG")
```
works. It doesn't yet do anything particularly smart other than create a small CQF and put stuff into it.

- [x] load and save support

---

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


betatim:
I think this is basically working and looks pretty promising!

Used https://gist.github.com/betatim/65b970943bee792e0d83c82fc8688935 to do some benchmarking and try to understand how the CQF works/behaves. (The way to measure extra mem used is pretty unreliable...watch the process' virtual mem in `top` instead)

False positive rate after storing `entries` unique kmers evaluated using 20k kmers that are definitely not in the filter:

entries | BF | CQF
----------|------|-------
5000000 | 0.00145 | 0.00015
8000000 | 0.00315 | 0.0003
9000000 | 0.00395 | 0.0003

For roughly same amount of memory used per data structure we get an order of magnitude less false positives?! Quite positive ...

betatim:
Updated gist now contains script for timing measurements: `python fp.py [bf|cqf]`

kind | load | query | t_load (s) | t_queryNP (s) | t_queryP (s)
-------|-------|----------|----------|----------------|-------------
bf | 9000000 |1000000 | 6.6050 | 0.77769 | 0.805
cqf | 9000000 | 1000000 | 5.5954 | 0.6589 | 0.72047

load and query are the number of kmers loaded into the filter and the number of kmers queried. t_X is time taken. queryNP = query kmers not present, queryP = query kmers present.

betatim:
Using `-Ofast` and `-msse4.2 -D__SSE4_2_` as compile options for all of khmer seems to slow things down. Where as for the CQF they make a difference.

khmer/_oxli/graphs.pxd
```diff
@@ -42,16 +42,16 @@ cdef extern from "oxli/hashtable.hh" namespace "oxli":
         void count(const char *)
         void count(HashIntoType)
         bool add(const char *)
-        bool add(HashIntoType)
+        #bool add(HashIntoType)
```
betatim: A "hack" found on SO: change the second line to `bool add2 "add"(HashIntoType)`. Somehow tricks cython but because you provide a "real name" that is the same as above the compiler figures it all out. This seems to make the compiler happy but segfaults the second time I try to add a kmer :-/

betatim: Segfault also happens with the original code on OSX.

khmer/_oxli/graphs.pxd
```diff
@@ -42,16 +42,16 @@ cdef extern from "oxli/hashtable.hh" namespace "oxli":
         void count(const char *)
         void count(HashIntoType)
         bool add(const char *)
-        bool add(HashIntoType)
+        bool add2 "add"(HashIntoType)
```
betatim: dat hack?!

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,74 @@
...
+
+    def count(self, kmer):
+        """Increment the count of this k-mer
+
+        `kmer` can be either a string or an integer representing the hashed
+        value of the kmer.
+        """
+        if isinstance(kmer, (unicode, str)):
+            data = _bstring(kmer)
+            return deref(self.c_table).add(deref(self.c_table).hash_dna(data))
```
betatim: Does one of the level 7 cython wizards know why you have to convert the string to a hash yourself instead of getting the overloaded method to do our bidding for us?

khmer/__init__.py
```diff
@@ -359,6 +363,17 @@ def __new__(cls, k, starting_size, n_tables):
         return counttable
 
 
+class QFCounttable(_QFCounttable):
```
camillescott: Note that now (in the dawn of the Glorious Revolution), we'll be removing this pattern -- this `__new__` method should just be implemented in the `__cinit__`.

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,194 @@
...
+        read_parser = FastxParser(file_name)
+        deref(self.c_table).consume_seqfile[CpFastxReader](read_parser._this,
+                                                           total_reads,
+                                                           n_consumed)
+
+    def abundance_distribution(self, file_name, tracking):
+        """Calculate the k-mer abundance distribution of the given file_name."""
+        read_parser = FastxParser(file_name)
+        cdef uint64_t * x = deref(self.c_table).abundance_distribution[CpFastxReader](
+                read_parser._this, (<CPyHashtable_Object>tracking).hashtable)
```
betatim: Need help again @camillescott :-/ `tracking` is a `Nodegraph` which I'd like to somehow unpack so I can take the `hashtable` attribute and pass it along.

A bit like:
```c++
khmer_KHashtable_Object * tracking_obj = NULL;
PyArg_ParseTuple(args, "O!", &khmer_KHashtable_Type, &tracking_obj)
// now use tracking_obj->hashtable
```

but how? It seems silly that we can't reach into these extension types 😕

betatim:
sse4, cqf, 9000000, 1000000, 5.88, 0.66, 0.70
avx, cqf, 9000000, 1000000, 6.02, 0.66, 0.72
w/o, cqf, 9000000, 1000000, 5.90, 0.66, 0.72

Whether or not we use the SSE4.2 extension seems to make only a very small difference. For now disabling it as we would need a good way to detect whether or not the CPU supports it (if someone has a good idea on that let me know).

betatim:
Ready for review! @luizirber, @camillescott (I seek approval from the cython deity 🙇), @standage , or @ctb

There is a "detail" to sort out with `n_occupied()` and how to handle the data structure becoming full. However there is enough code to start reviewing it and then we sort things after/during. This doesn't make the CQF available from any of the scripts so I think we are OK to merge this and then doctor around a bit more after.

ctb:
+1 for review

betatim:
My conclusion from https://github.com/splatlab/cqf/issues/4 is that there is no way to get the CQF to behave like a BF which allows the user to continue inserting items beyond the point where the FP rate "explodes". The good thing is we have a way to gracefully exit when we reach that point.

ctb:
good to know & good to have!

betatim:
Can we decide on how to handle that situation when we actually expose this via one of the scripts?

ctb:
How about raising an exception and then having scripts output some useful
information e.g. total # of k-mers?

But I'm leary of putting too much effort into this until we have a try at
it ourselves.  No idea what to expect / how big a problem it really will
be.

include/oxli/storage.hh
```diff
@@ -403,6 +405,54 @@ public:
...
+ *
+ * \brief A Quotient Filter storage
+ */
+ class QFStorage : public Storage {
+protected:
+  QF cf;
+
+public:
+  QFStorage(int size) {
+    qf_init(&cf, (1ULL << size), size+8, 0);
```
camillescott: Maybe some brief code comments on these init params?

khmer/_oxli/graphs.pxd
```diff
@@ -85,6 +85,43 @@ cdef extern from "oxli/hashtable.hh" namespace "oxli":
     cdef cppclass CpNodetable "oxli::Nodetable" (CpHashtable):
         CpNodetable(WordLength, vector[uint64_t])
 
+    cdef cppclass CpQFCounttable "oxli::QFCounttable":
```
camillescott: this should be derived from `CpHashtable` (see above with Nodetable and Counttable), which will allow you to omit all these redefinitions

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,258 @@
...
+    def __cinit__(self, int k, int starting_size):
+        # starting size has to be a power of two
+        power_of_two = ((starting_size & (starting_size - 1) == 0) and
+                        (starting_size != 0))
+        if not power_of_two:
+            raise ValueError("starting_size has to be a power of two.")
+        self.c_table.reset(new CpQFCounttable(k, int(log(starting_size, 2))))
+
+    def add(self, kmer):
+        """Increment the count of this k-mer.
```
camillescott: Note that in liboxli, `add` returns whether the k-mer was new, while `count` does not. I don't think this is explicitly documented beyond just being that way in the code, however.

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,258 @@
...
+        return self.count(kmer)
+
+    def count(self, kmer):
+        """Increment the count of this k-mer
+
+        `kmer` can be either a string or an integer representing the hashed
+        value of the kmer.
+        """
+        if isinstance(kmer, str):
+            temp = kmer.encode('utf-8')
```
camillescott: i believe you should be able to write this as `cdef char * temp = ...`?

luizirber:
Quick comment: I started the wrapper for C++ -> Python exceptions in [the HLL Cython PR][0], [this][1] is how you declare it on the Cython class.

[0]: https://github.com/dib-lab/khmer/blob/6850b3d62df232751ea541e05eac025fe4dc9441/khmer/_oxli/oxli_exception_convert.cc
[1]: https://github.com/dib-lab/khmer/pull/1730/files#diff-8749b1403e4ad7a5717cbaf397bc9b13R34

khmer/_oxli/graphs.pxd
betatim: Tried that and you end up with a bunch of these:
```
khmer/_oxli/graphs.pyx:58:43: Cannot assign type 'char *' to 'HashIntoType'

Error compiling Cython file:
------------------------------------------------------------
...
        For Nodetables and Counttables, this function will fail if the
        supplied k-mer contains non-ACGT characters.
        """
        if isinstance(kmer, str):
            temp = kmer.encode('utf-8')
            return deref(self.c_table).get_count(<char*>temp)
```

not sure why defining them again in the same order (instead of inheriting) fixes this :-/

betatim: Basically, agree that this is stupid but don't know how to improve.

camillescott: This referring to the class hierarchy or to the temp variable thing?

betatim: The class hierarachy

khmer/_oxli/graphs.pyx
betatim: You can't define a `cdef` inside the if statement, so you could declare it outside and then assign here but I'd claim that is even more clunky than the cast.

betatim:
Like the exception handling, for the moment I will add `except +` to those methods that need it, once #1730 is merged, I'll switch to that.

include/oxli/hashtable.hh
```diff
@@ -140,6 +140,7 @@ protected:
     HashIntoType    bitmask;
     unsigned int    _nbits_sub_1;
 
+public:
```
betatim: What is a better way of doing this? I needed to make the destructor of `Hashtable` public for unique_ptr's `reset()` to work. Could someone tell me why it was protected? I assume it is to make sure it doesn't get called for a derived class. Wisdom welcome.

khmer/_oxli/graphs.pxd
```diff
@@ -64,19 +64,19 @@ cdef extern from "oxli/hashtable.hh" namespace "oxli":
...
+        void get_kmer_counts(const string &,
                              vector[BoundedCounterType] &) const
         uint8_t ** get_raw_tables()
         BoundedCounterType get_min_count(const string &)
         BoundedCounterType get_max_count(const string &)
-        uint64_t * abundance_distribution[SeqIO](unique_ptr[CpReadParser[SeqIO]]&, 
-                                          CpHashtable *)
         uint64_t * abundance_distribution[SeqIO](string, CpHashtable *)
+        uint64_t * abundance_distribution[SeqIO](unique_ptr[CpReadParser[SeqIO]]&,
+                                          CpHashtable *)
```
betatim: This seems crazy, but swapping the order makes things work.

betatim:

betatim:
An alternative to merging this would be to pull out the `Hashtable` stuff and get that merged first, then think about CQF stuff more (if needed). With the `Hashtable` stuff merged we can bring about more of the glorious cython future for other parts.

khmer/_oxli/graphs.pyx
camillescott: fair 'nuff

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,261 @@
...
+        cdef unsigned long long n_consumed = 0
+        cdef unsigned int total_reads = 0
+
+        read_parser = ReadParser(file_name)
+        cdef CPyReadParser_Object* parser = <CPyReadParser_Object*>read_parser
+        deref(self.c_table).consume_seqfile[CpFastxReader](parser.parser,
+                                                           total_reads,
+                                                           n_consumed)
+
+    def abundance_distribution(self, file_name, tracking):
```
camillescott: Why not require this to be `Hashtable tracking`? The CPython version gets deprecated via this PR anyhow.

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,261 @@
...
+        cdef CPyReadParser_Object* parser
+        if isinstance(file_name, str):
+            read_parser = ReadParser(file_name)
+            parser = <CPyReadParser_Object*>read_parser
+
+        else:
+            raise ValueError('Expected file_name to be string, '
+                             'got {} instead.'.format(type(file_name)))
+
+        cdef CPyHashtable_Object* hashtable
```
camillescott: ie here we can just pull the `_this` pointer from the Cython extension class.

khmer/_oxli/graphs.pyx
```diff
@@ -1,20 +1,261 @@
...
+                             'Nodegraph, got {} instead.'.format(type(tracking)))
+
+        cdef uint64_t * x = deref(self.c_table).abundance_distribution[CpFastxReader](
+                parser.parser, hashtable.hashtable)
+        abunds = []
+        for i in range(MAX_BIGCOUNT):
+            abunds.append(x[i])
+        return abunds
+
+    def abundance_distribution_with_reads_parser(self, read_parser, tracking):
```
camillescott: Same thoughts as above. Additionally, `read_parser` can be required to be a `FastxParser` from `oxli.parsing`, with the underlying `CpFastxReader` pulled from the this pointer.

betatim: Trying to remember why exactly I went with this. Part of this is there are a bunch of tests that start failing when I add:
```python
from khmer._oxli.parsing import FastxParser as ReadParser
```
to khmer/__init__.py, which would seem to be a useful thing to do to keep all the people using `ReadParser`s happy. Not sure this is super tricky, but can't work it out atm :-/ Merge now, improve later?

ctb:
Looks like test failures in the assembly code cc @camillescott

betatim:
Related to the discussion between @standage and @camillescott on slack yesterday regarding bytes vs strings in the latest version of cython, I think.

(nb. assembly code -> the other ASM)

betatim:
Nearly forgot the most important part: :hand: (high five) and thanks for the patient commenting :) Long may the cython revolution continue :fist_raised: ;)

PR #1678, opened 2017-04-21
Add test that parses stdout from trim-low-abund.py

### LINKED ISSUE #975: write test to ensure that output from "trim-low-abund --out -" is in good format

betatim:
Fixes #975

Explicitly check that the output of trim-low-abund can be parsed as
FASTA when it is being redirected to stdout.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/tests/test_scripts.py
+++ b/tests/test_scripts.py
@@ -2616,7 +2616,10 @@ def test_trim_low_abund_stdout():
     args = ["-k", "17", "-x", "1e7", "-N", "2", infile, "-o", "-"]
     _, out, err = utils.runscript('trim-low-abund.py', args, in_dir)
 
-    assert 'GGTTGACGGGGCTCAGGG' in out
+    # attempt to parse output to check it is in FASTA format
+    stream = io.StringIO(out)
+    assert list(screed.fasta.fasta_iter(stream)), "can't parse stdout"
+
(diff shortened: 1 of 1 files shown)


betatim:

tests/test_scripts.py
```diff
@@ -2616,7 +2616,10 @@ def test_trim_low_abund_stdout():
     args = ["-k", "17", "-x", "1e7", "-N", "2", infile, "-o", "-"]
     _, out, err = utils.runscript('trim-low-abund.py', args, in_dir)
 
-    assert 'GGTTGACGGGGCTCAGGG' in out
+    # attempt to parse output to check it is in FASTA format
+    stream = io.StringIO(out)
+    assert list(screed.fasta.fasta_iter(stream)), "can't parse stdout"
```
standage: We could check the content of the record(s) returned here.

betatim: I think they'd all be "GGTTGACGGGGCTCAGGG"? Should we make an issue?

standage: See #1679.

PR #1681, opened 2017-04-27
Add comment to the Makefile to document where to change things

### LINKED ISSUE #1552: Figure out correct instructions for turning off RC collapsing
See http://lists.idyll.org/pipermail/khmer/2016-December/000775.html for where NO_UNIQUE_RC hack doesn't work.

betatim:
Fixes #1552

I added a comment in the Makefile pointing people at `setup.{py,cfg}`. That is the place to edit stuff if you want to influence how things are compiled. The variables in `Makefile` are only used by `cppcheck` but I guess most people start by looking in the `Makefile` and then get confused why it doesn't seem to do anything. Ideally we could extract the values from `setup.cfg` instead of duplicating them, but my `distutils` foo is not strong enough for that. Any ideas?

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,8 @@ GCOVRURL=git+https://github.com/nschum/gcovr.git@never-executed-branches
 VERSION=$(shell ./setup.py version | grep Version | awk '{print $$4}' \
 	| sed 's/+/-/')
 
+# The following four variables are only used by cppcheck. If you want to
+# change how things are compiled edit `setup.cfg` or `setup.py`.
 DEFINES += -DNDEBUG -DVERSION=$(VERSION) -DSEQAN_HAS_BZIP2=1 \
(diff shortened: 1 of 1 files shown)


betatim:

standage:
Yeah, since so many of the compilation details are delegated to the Python ecosystem via setup.py, configuration can be non-obvious and non-trivial. I remember you mentioning `make -j` recently. I've tried that myself before, and it stinks that little things like that aren't straightforward.

betatim:
I think I spotted the equivalent for `make -j` in a `setup.*` somewhere or docs I read, I'll try to refind it.

@luizirber do you have an idea if there is any hope of running a (simple) python command to get the "compiler flags used by `setup.py`" so we can insert that in the `Makefile`?

PR #1682, opened 2017-04-27
Add to CLI usage information

### LINKED ISSUE #1676: command usage descriptions
Going through the `--help` messages for the [zsh completion](https://github.com/kloetzl/biozsh/commit/1567b15e1d9f54c73f6ba98367f3e74077a6a487) I found a number of small issues.

* `abundance-dist-single`: `-f` and `-s` have the same description.
* `abundance-dist-single`: `--savegraph` has an empty default
* `do-partition`: `--keep-subsets` has two defaults.
* various: `--max-memory-usage` states that is is used for “maximum amount of memory to use for data structure”. The default is “None”, which can be read as “no memory”. A better description would be “unlimited”.
* In some cases the short option is given first, in others the long version. See `khmer extract-paired-reads --help` for `-d` and `-p`, resp.
* Some descriptions end with a full stop, others don't.
* `normalize-by-median`: `-q, -R, --report-frequency` are missing a description
* `normalize-by-median`: `-C` s/numer/number/
* `normalize-by-median`: `--force_single` is one of the very few option with an underscore
* `readstats` and `trim-low-abund` are missing multiple argument descriptions

Btw, can short options be stacked? May arguments to long options be given with a `=`?

betatim:
Fixes #1676. There are a few where I didn't add more docs because it seems not needed (e.g. --quiet)

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/scripts/abundance-dist-single.py
+++ b/scripts/abundance-dist-single.py
@@ -94,7 +94,7 @@ def get_parser():
     parser.add_argument('-s', '--squash', dest='squash_output', default=False,
                         action='store_true',
                         help='Overwrite output file if it exists')
-    parser.add_argument('--savegraph', default='', metavar="filename",
+    parser.add_argument('--savegraph', metavar="filename",
                         help="Save the k-mer countgraph to the specified "
                         "filename.")
     parser.add_argument('-f', '--force', default=False, action='store_true',
@@ -112,7 +112,7 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
     report_on_config(args, graph_type)
 
     check_input_files(args.input_sequence_filename, args.force)
-    if args.savegraph:
+    if args.savegraph is not None:
         graphsize = calculate_graphsize(args, graph_type)
         check_space_for_graph(args.savegraph, graphsize, args.force)
     if (not args.squash_output and
@@ -213,7 +213,7 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
         if sofar == total:
             break
 
-    if args.savegraph:
+    if args.savegraph is not None:
         log_info('Saving k-mer countgraph to {savegraph}',
                  savegraph=args.savegraph)
         countgraph.save(args.savegraph)
--- a/scripts/abundance-dist-single.py
+++ b/scripts/abundance-dist-single.py
@@ -98,7 +98,7 @@ def get_parser():
                         help="Save the k-mer countgraph to the specified "
                         "filename.")
     parser.add_argument('-f', '--force', default=False, action='store_true',
-                        help='Overwrite output file if it exists')
+                        help='Override sanity checks')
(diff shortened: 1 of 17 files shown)


standage:
Sorry, I should have let you hit the "update branch" button on this. I thought you had already asked for a review.

betatim:
No problem. This can be reviewed I think :) In particular the changes in `abundance-dist-single.py` would be good to hear your thoughts on.

scripts/extract-paired-reads.py
```diff
@@ -94,11 +94,11 @@ def get_parser():
     parser.add_argument('-d', '--output-dir', default='', help='Output '
                         'split reads to specified directory. Creates '
                         'directory if necessary')
-    parser.add_argument('--output-paired', '-p', metavar="filename",
+    parser.add_argument('-p', '--output-paired', metavar="filename",
```
standage: There are many scripts where the long option is defined before the short option. Any reason for singling out this script?

scripts/do-partition.py
```diff
@@ -94,9 +93,8 @@ def get_parser():
     parser.add_argument('--no-big-traverse', dest='no_big_traverse',
                         action='store_true', default=False,
                         help='Truncate graph joins at big traversals')
-    parser.add_argument('--keep-subsets', dest='remove_subsets',
-                        default=True, action='store_false',
-                        help='Keep individual subsets (default: False)')
+    parser.add_argument('--keep-subsets', default=True, action='store_false',
+                        help='Keep individual subsets')
```
standage: The default should be false here.

scripts/abundance-dist-single.py
```diff
@@ -94,11 +94,11 @@ def get_parser():
     parser.add_argument('-s', '--squash', dest='squash_output', default=False,
                         action='store_true',
                         help='Overwrite output file if it exists')
-    parser.add_argument('--savegraph', default='', metavar="filename",
+    parser.add_argument('--savegraph', metavar="filename",
                         help="Save the k-mer countgraph to the specified "
                         "filename.")
     parser.add_argument('-f', '--force', default=False, action='store_true',
-                        help='Overwrite output file if it exists')
+                        help='Override sanity checks')
```
standage: 👍

scripts/do-partition.py
betatim: I also changed the action to `store_true`

scripts/extract-paired-reads.py
betatim: makes it consistent for this script, is the only reason really.

betatim: Agree on fixing it for all if we start going down this road.

betatim:
I think we are consistent now, I'll have a another look tomorrow morning, right now all usage screens have blurred into one.

PR #1689, opened 2017-05-04
[MRG] Switch SmallCounttable to murmurhash

### LINKED ISSUE #1659: Should SmallCounttable behave like Counttable?
Right now, SmallCounttable uses the k<32 / fully reversible hash function.  I think for name consistency it should behave like a Counttable (using murmurhash). Thoughts?

If desired, this could be done by copy/pasting the hashing code from Counttable, OR by deriving SmallCounttable from Counttable instead of from Hashtable.

betatim:
Fixes #1659 `SmallCounttable` uses murmurhash now. We can't inherit from `Counttable` because that would lock us into the storage used by that. So went with the approach of copying the hashing stuff.

Does someone know if having mixins works as nicely in c++ as it does in python? Then we could have a `MurmurHashMixin` to add the functionality.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?

### DIFF
--- a/lib/hashtable.cc
+++ b/lib/hashtable.cc
@@ -507,57 +507,6 @@ const
     return posns;
 }
 
-class MurmurKmerHashIterator : public KmerHashIterator
-{
-    const char * _seq;
-    const char _ksize;
-    unsigned int index;
-    unsigned int length;
-    bool _initialized;
-public:
-    MurmurKmerHashIterator(const char * seq, unsigned char k) :
-        _seq(seq), _ksize(k), index(0), _initialized(false) {
-        length = strlen(_seq);
-    };
-
-    HashIntoType first() { _initialized = true; return next(); }
-
-    HashIntoType next() {
-        if (!_initialized) { _initialized = true; }
-
-        if (done()) {
-            throw khmer_exception("past end of iterator");
-        }
-
-        std::string kmer;
-        kmer.assign(_seq + index, _ksize);
-        index += 1;
-        return _hash_murmur(kmer, _ksize);
-    }
-
-    bool done() const {
-        return (index + _ksize > length);
-    }
-
-    unsigned int get_start_pos() const {
-        if (!_initialized) { return 0; }
-        return index - 1;
-    }
-    unsigned int get_end_pos() const {
-        if (!_initialized) { return _ksize; }
-        return index + _ksize - 1;
-    }
-};
-
-KmerHashIteratorPtr Counttable::new_kmer_iterator(const char * sp) const {
-    KmerHashIterator * ki = new MurmurKmerHashIterator(sp, _ksize);
-    return unique_ptr<KmerHashIterator>(ki);
-}
-
-KmerHashIteratorPtr SmallCounttable::new_kmer_iterator(const char * sp) const {
-    KmerHashIterator * ki = new MurmurKmerHashIterator(sp, _ksize);
-    return unique_ptr<KmerHashIterator>(ki);
-}
 
 template void Hashtable::consume_seqfile<FastxReader>(
     std::string const &filename,
--- a/lib/hashtable.cc
+++ b/lib/hashtable.cc
@@ -554,6 +554,11 @@ KmerHashIteratorPtr Counttable::new_kmer_iterator(const char * sp) const {
     return unique_ptr<KmerHashIterator>(ki);
 }
 
+KmerHashIteratorPtr SmallCounttable::new_kmer_iterator(const char * sp) const {
+    KmerHashIterator * ki = new MurmurKmerHashIterator(sp, _ksize);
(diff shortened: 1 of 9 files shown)


ctb:
> Does someone know if having mixins works as nicely in c++ as it does in python? Then we could have a `MurmurHashMixin` to add the functionality.

yes, exactly :)

betatim:
Currently think this is the answer: https://github.com/dib-lab/khmer/pull/1690#issuecomment-299392030 to mixins or not.

lib/hashtable.cc
```diff
@@ -507,52 +507,6 @@ const
     return posns;
 }
 
-class MurmurKmerHashIterator : public KmerHashIterator
```
betatim: this got moved as one piece to `hashtable.hh`

betatim:
I think this is a good compromise between mixins and copy&paste.

I haven't yet switched `Nodetable` because currently `load_nodegraph` is used for both Nodetable and Nodegraph. We'd have to add a distinct `load_nodetable` function. Does someone remember why Counttable and Countgraph has separate load functions but Node* doesn't?

betatim:

In particular @standage to discuss what to add here and what to do in #1690 (picking the test from there would be good for example)

betatim:
I will measure the performance of copy&paste vs inheritance.

betatim:
What is the policy for breaking changes for `*table`s being stored in files? If you dump a file in khmer2.0 should you expect it to be read in 2.1? Just realised that by changing the hash functions we should also prevent people from loading files that were written with the old hash functions.

ctb:
yes, as you say. this connects with some of the load* features that Daniel
has been implementing.

betatim:
Daniel's work is in dib-lab/kevlar#71. What I was thinking of is e758660. This bumps the version number of the file format. I think we need this because reading a file created with the old hash function will result in nonsense when you read it in with khmer that uses murmurhash. Which made me wonder, is that kind of change allowed when going from 2.0 to 2.1? It is backward incompatible which suggest major version but I can't find where khmer docs talk about files.

ctb:
look up Semantic Versioning in general, right?

as long as the old stuff continues to work the way it did, adding new
features / command line flags is ok.  that's what this is, right?

On Mon, May 08, 2017 at 06:54:25AM -0700, Tim Head wrote:
> Daniel's work is in dib-lab/kevlar#71. What I was thinking of is e758660. This bumps the version number of the file format. I think we need this because reading a file created with the old hash function will result in nonsense when you read it in with khmer that uses murmurhash. Which made me wonder, is that kind of change allowed when going from 2.0 to 2.1? It is backward incompatible which suggest major version but I can't find where khmer docs talk about files.
> 
> -- 
> You are receiving this because you were mentioned.
> https://github.com/dib-lab/khmer/pull/1689#issuecomment-299873255
-- 

betatim:
Right now this replaces the old hash function which means you should not load old files -> backwards incompatible change. Need to do some thinking how to do this as a flag.

ctb:
Maybe I'm missing something here but our last release didn't include smallcounttable so no problem there. Also the relevant tests have broken afaik so we should be good.

--

standage:
Correct, release 2.0 didn't have NibbleStorage or the murmurhash-based *table structures. This fix shouldn't affect SemVer.

ctb:
On Mon, May 08, 2017 at 05:02:00PM -0700, Daniel Standage wrote:
> Eventually, we want to support various hash functions that are useful in specific contexts. The simplest way to do this would be to define a class for every element in the cross product of hash functions and storage types. This could quickly become very intractable, however, the more hash functions we add. I think this deserves some more thought and should be separated to a dedicated issue thread.

makes sense.  punt to other PR and connect to #1616 as a use case?

betatim:
No more configurable hash functions. I'll move the code for that to a new branch and we can use that as a starting point for discussing the nicest way to make it happen.

betatim:
Using this as benchmark:
```python
import time
import khmer

k = 23
n_slots = 27

ct = khmer.Counttable(k, 2**n_slots, 4)

tic = time.time()
ct.consume_seqfile('ecoli_ref.small.fastq')
toc = time.time()

print(toc-tic)
```

I get virtually identical times (29.866569 vs 30.134745121s) for the version in this branch and https://gist.github.com/betatim/c69abd32d81fe38a047669e1cfe4ea20

betatim:
Additional data point: if you use the old school reversible hash function then this benchmark finishes in ~15s (but of course won't work with k>32).

standage:
Well, looks great. Thanks for your work on this. Anything else needed here or are we ready for a merge?

betatim:
Loose ends tied up. Ready to go.

standage:
👍

PR #1690, opened 2017-05-04
Fixing unexpected behavior w.r.t. reverse_hash and table sketches

standage:
The [Node|Count|SmallCount]table data structures use the irreversible MurmurHash (or at least they *should* be: see #1689). Therefore, any call to `table.reverse_hash(hashval)` should result in a `ValueError`. Currently this is only the case for `Counttable`.

This PR introduces a regression test that should pass once everything is functioning properly.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/lib/hashtable.hh
+++ b/lib/hashtable.hh
@@ -367,6 +367,31 @@ class SmallCounttable : public khmer::Hashtable
 public:
     explicit SmallCounttable(WordLength ksize, std::vector<uint64_t> sizes)
         : Hashtable(ksize, new NibbleStorage(sizes)) { } ;
+
+    inline
+    virtual
+    HashIntoType
+    hash_dna(const char * kmer) const {
+        if (!(strlen(kmer) >= _ksize)) {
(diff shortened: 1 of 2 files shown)


standage:
So...copying and pasting class methods from `Counttable` to `Nodetable` and `SmallCounttable` fixes the problem. But this much code duplication is bound to lead to problems later. Any comments / suggestions?

betatim:
For `SmallCounttable` and murmurhash checkout #1689.

betatim:
(not sure where we should discuss this, so putting it here for the moment)

In python I'd got for having a `MurmurHashMixin`, after some digging around it seems in c++ this isn't a frequently used pattern and you need a bit of infrastructure and stuff that seems to not be worth it for our simple case IMHO.

How about creating `MurmurHashtable` which inherits from `Hashtable` and then inheriting from `MurmurHashtable` when we want murmurhash?

standage:
Fixed with #1689.

Issue #1393, opened 2016-06-26
khmer v2.1 - next release after 2.0


ctb:
Notes:

With #1380 we will be adding --quiet to many scripts, so we need to increment minor version number.

standage:
We don't have a specific checklist or timeline slated for this release, do we?

ctb:
Nope!

We should revisit the '2.0+' issues and retag them as '2.1' as necessary.

betatim:
We are getting close, should we make a release candidate?

betatim:
Just went through the checklist for everything but uploading to the test instance of pypi. If we merge #1688 I will make a rc1 with a tag on github etc.

If one of the people who have the account name and password for pypi could share that would be cool.

ctb:
what's your pypi account?
On Tue, May 02, 2017 at 08:52:11AM -0700, Tim Head wrote:
> Just went through the checklist for everything but uploading to the test instance of pypi. If we merge #1688 I will make a rc1 with a tag on github etc.
> 
> If one of the people who have the account name and password for pypi could share that would be cool.
> 
> -- 
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/dib-lab/khmer/issues/1393#issuecomment-298676911
-- 
C. Titus Brown, ctbrown@ucdavis.edu

betatim:
betatim

ctb:
you now have maintainer access

betatim:
I will make v2.1rc1 tomorrow morning (my time) unless I hear otherwise.

betatim:
You can install khmer v2.1rc1 with:
```
pip install -i https://testpypi.python.org/pypi --pre --no-clean khmer
```

The pypi page needs a bit of massaging as it doesn't render the README's markdown: https://testpypi.python.org/pypi/khmer/2.1rc1

standage:
Awesome. I installed successfully in a fresh virtualenv. (I had to install screed separately, I assume due to using the PyPI test database).

Regarding rendering the README, there has been an open request on the PyPI issue tracker to support Markdown rendering since 2014. I don't see that happening anytime soon. But take a look at https://github.com/pypa/pypi-legacy/issues/148#issuecomment-227757822 for a potential solution. It's a bit of an ugly hack, but if we do the conditional import of `pypandoc` then we only have to make it a requirement for developers cutting a release, which I think is reasonable.

betatim:
[...]

PR #1695, opened 2017-05-10
Add conversion of README text from md to rst

betatim:
Implements [@standage's idea](https://github.com/dib-lab/khmer/issues/1393#issuecomment-300236583) for converting our readme from MD to RST locally. We need to do this if we want the PyPI page to look nice as it seems PyPI won't be rendering markdown any time soon.

Compare `python setup.py --long-description` with and without pandoc installed.

Wondering if there is a way to automatically pull in pypandoc when "making a release". I thought that `'setup_requires'` would take care of this but that only get's executed when `setup()` is called.

- [x] Is it mergeable?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -192,12 +192,20 @@ if "-rc" in versioneer.get_version():
 else:
     CLASSIFIERS.append("Development Status :: 5 - Production/Stable")
 
+# Before making a new release make sure to install pypandoc. Pypi does not
+# render markdown which is why we convert to rst here.
+try:
+    import pypandoc
+    long_description = pypandoc.convert('README.md', 'rst')
+except(IOError, ImportError):
+    long_description = open('README.md').read()
+
 SETUP_METADATA = \
     {
         "name": "khmer",
         "version": versioneer.get_version(),
         "description": 'khmer k-mer counting library',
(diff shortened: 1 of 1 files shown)


ctb:
is there a reason we don't just shift README.md to README.rst once?

betatim:
Not sure, two things come to mind as "why not rst":

* do libraries.io and friends render rst?
* do the humans involved like reading/editing rst?

+0 from my side

standage:
We made the switch from .rst to .md a few months ago since there was a general consensus (or at very least a lack of protest) that Markdown is more pleasant to work with. PyPI rendering is a big strike against .md. Are there any strikes against .rst (aside from personal preferences)?

betatim:
https://twitter.com/teabass/status/862409548312784896 so no technical obstacles?

standage:
Ok. Well, I'm pretty sure I'm the one that switched to .md, so I'd be happy to take care of switching it back. Less pain long term.

betatim:
Ok. Good thing we aren't politicians, just imagine the press "khmer flipflops on README format!!". I'll close this.

ctb:
presumably we will not do much editing or README.rst.

PR #1699, opened 2017-05-18
[WIP] Exploring ntHash as a rolling/recursive hash solution

standage:
After spending some time last night playing with the rollinghashcpp code from @lemire, it doesn't look like it's suitable for our needs. In principle it looks like it could work, but currently it has two incompatibilities with our needs: first, it only supports up to 19 bit hash values; second, the hash functions are designed to be randomized, with no apparent way to seed the functions for deterministic hashing.

In this PR, I'm exploring the ntHash solution suggested by @asl (https://github.com/dib-lab/khmer/issues/1616#issuecomment-300334920). This implements a rolling canonical hash and supports (as far as I can tell) indefinitely large values of k. I have yet to benchmark against MurmurHash, but the basics seem to pass muster.

This PR will not be merged. We need to get the Cython branch merged first, at which point it'll be easier to just open a fresh PR.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


luizirber:
Big issue: it's GPL https://github.com/bcgsc/ntHash/blob/master/LICENSE

lemire:
Because I am pinged, I will comment. Please don't take it as a commentary on the suitability of our code.

> After spending some time last night playing with the rollinghashcpp code from @lemire, it doesn't look like it's suitable for our needs.

I am guessing that you allude to https://github.com/lemire/rollinghashcpp ? 

> with no apparent way to seed the functions for deterministic hashing

The API does not support it right now, true. We rely on a Mersenne Twister and [there is a method to pass a seed value](https://github.com/lemire/rollinghashcpp/blob/master/characterhash.h#L24).

> it only supports up to 19 bit hash values

That's not correct. For historical reasons, [the default bit parameter is set to 19, but you can specify it when calling the constructor](https://github.com/lemire/rollinghashcpp/blob/master/cyclichash.h#L29). As the [examples show](https://github.com/lemire/rollinghashcpp/blob/master/example4.cpp), it can support 64-bit hash values.

lemire:
If you guys want to discuss the issues, just ping me again.

betatim:
If you dig a bit in #1622 there are some modifications/examples I made to pass in the seed. Back then I [decided](https://github.com/dib-lab/khmer/pull/1622#issuecomment-280329571) that actually we want `CyclingHash` as that allows you to add/remove on the left of the string as well as on the right (this is to support computing the hash on the reverse complement). Maybe there is a better way of achieving this or the rabinkarp hash can be modified?

lemire:
Ok. So Cyclic has some nice properties as demonstrated in the following paper: 

- Recursive n-gram hashing is pairwise independent, at best, Computer Speech & Language, Volume 24, Issue 4, October 2010, Pages 698-710 http://arxiv.org/abs/0705.4676

(Just so the context is clear, *I* did not invent it.)

It is fast, simple, convenient, mathematically solid.

So if you have 64-bit hash values, and your "strings" are length "n", then you have "64-n" pairwise independent bits. That's pretty strong.

If you use CyclicHash as in this example, you do get 64 bits...

https://github.com/lemire/rollinghashcpp/blob/master/example4.cpp

As you say, it can allow you to go forward and backward.

I think it is what TwoPaCo uses: https://academic.oup.com/bioinformatics/article-abstract/doi/10.1093/bioinformatics/btw609/2725383/TwoPaCo-an-efficient-algorithm-to-build-the?redirectedFrom=fulltext

If you are hitting some difficulty using it, I'd love to discuss it.

standage:
Thanks for clarifying about the 19 bits @lemire. I assumed from your examples that up to 64 bits would be supported, but when I saw the open issue on the rollinghashcpp repo I figured I had been mistaken.

I did see the MTrand seed function, but didn't invest the time in making it accessible to the CyclicHash class. If the 19-bit limit isn't an issue, then the seed issue should be a straightforward fix, given sufficient interest.

lemire:
Yep

standage:
Huh...the compiler complaints about NTPC64 don't show up on my Macbook.

asl:
I've refactored & slightly generalized the code by @lemire. More importantly, I removed the hard dependency on Mersenne Twister since we've mostly interested in hashing of ACGT sequences.

See https://gist.github.com/asl/fce137a47fc7d3cd7ac2178d499d5c31

ctb:
Hi @asl / @lemire thank you for your engagement! Would you be willing to provide the code under a BSD-compatible license?  khmer is under BSD and we don't want to change that.

lemire:
Sure, we'll make sure it is BSD comp. if it isn't already.

ctb:
Thank you!

asl:
The original code of @lemire is under Apache 2.0, so it's pretty much BSD compatible. I'm fine to use any license as convenient for my code to be used inside khmer.

ctb:
that's excellent, thanks again!

lemire:
Ok. Let us know if there is an issue regarding the license or something else that needs our attention.

mohamadi:
We just changed the ntHash license from GPL3 to MIT License.

ctb:
Abandoned => #1792.

Issue #1565, opened 2016-12-21
'pip show khmer' fails due to author unicode


ctb:
```
% pip show khmer
Name: khmer
Version: 2.0+715.gd841a57
Summary: khmer k-mer counting library
Home-page: https://khmer.readthedocs.io/ 
--- Logging error ---
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.5/lib/python3.5/logging/__init__.py", line 982, in emit
    stream.write(msg)
UnicodeEncodeError: 'ascii' codec can't encode character '\xe1' in position 277: ordinal not in range(128)
Call stack:
  File "/Users/t/dev/jup/bin/pip", line 11, in <module>
    sys.exit(main())
  File "/Users/t/dev/jup/lib/python3.5/site-packages/pip/__init__.py", line 233, in main
    return command.main(cmd_args)
  File "/Users/t/dev/jup/lib/python3.5/site-packages/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/Users/t/dev/jup/lib/python3.5/site-packages/pip/commands/show.py", line 42, in run
    results, list_files=options.files, verbose=options.verbose):
  File "/Users/t/dev/jup/lib/python3.5/site-packages/pip/commands/show.py", line 133, in print_results
    logger.info("Author: %s", dist.get('author', ''))
Message: 'Author: %s'
[...]

PR #1705, opened 2017-05-23
[WIP] Try to fix encoding

standage:
I'm unable to reproduce the behavior @ctb describes in https://github.com/dib-lab/khmer/issues/1565#issuecomment-303251807. Perhaps this is the culprit/solution?

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,6 +38,7 @@
 
 import ez_setup
 
+import codecs
 import os
 import sys
 from os import listdir as os_listdir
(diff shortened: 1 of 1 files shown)


ctb:
I've updated my default encoding to en_US.utf-8. I'm not sure we need to fix this since so far it's only been me.

standage:
Ok, great. I thought from your last comment that it was still a problem you wanted resolved.

PR #1712, opened 2017-05-26
Update CPP examples to new oxli/cython setup

betatim:
This doesn't contain the modification to `setup.py` that I needed to make it work locally. Need to decide why/if it is needed or not.

- [x] Is it mergeable?

### DIFF
--- a/examples/c++-api/Makefile
+++ b/examples/c++-api/Makefile
@@ -1,5 +1,5 @@
 CFLAGS=--std=c++11 \
-       -I ../../lib \
+	   -I ../../include \
 	   -I ../../third-party/smhasher \
 	   -I ../../third-party/seqan/core/include/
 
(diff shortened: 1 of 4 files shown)


betatim:
Paging @camillescott and @standage for cython wisdom and merge.

betatim:
This PR is against #1595

standage:
Compiles and runs on my machine :)

betatim:
Merge?

standage:
However, I don't understand why it took 45+ seconds of hard thinking to execute `make help` on this branch, which shouldn't be touching `setup.py` or anything else in the dev environment AFAICT.

standage:
I'll defer to @camillescott on merging.

betatim:
Not sure about `make help`. My guess would be one of the variables that uses `shell` is to blame?

betatim:
Is the `make help` thing also a problem on the original cython branch?

It would be good if we can keep rolling to get this cython thing done and dusted, so shall we merge this instead of waiting for Camille?

camillescott:
Welp sorry to say that I fixed this already on the plane to Hawaii last week, but forgot to push the commit. So, closing this pull. Sorry @betatim

betatim:
Tzz

Issue #1716, opened 2017-05-31
Add a khmer.load function


luizirber:
Right now we have two functions:
- `khmer.load_countgraph`
- `khmer.load_nodegraph`

It should be possible to figure out which one is the right one (based on the file headers?) and have one function `khmer.load` that calls the appropriate function for each data structure.

standage:
Relevant thread: https://github.com/dib-lab/kevlar/issues/67.

For now, it should(?) be easy enough to add new `ht_type`s for \*table data structures. But just as we're modularizing storage, hash functions, etc, we should consider adding a new field indicating which hash function was used. This would require a major version bump though, so not until we're ready for 3.0.

PR #1717, opened 2017-06-01
[WIP] Enable automatic detecting and loading all filter/sketch types

standage:
See #1716.

So far this PR replaces the *ad hoc* `ht_type` defines with an enumerated type, and adds values for the \*table types. Now the question is how to distinguish between \*graph and \*table types in the storage class, which by design is agnostic to this.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/include/oxli/oxli.hh
+++ b/include/oxli/oxli.hh
@@ -87,13 +87,20 @@ private:\
 
 #   define SAVED_SIGNATURE "OXLI"
 #   define SAVED_FORMAT_VERSION 4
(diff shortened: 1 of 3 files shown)


standage:
It looks like we either need to store a `ht_type` in the storage object at creation time, or pass the value to storage when `load` and `save` are invoked. The latter seems less problematic, but neither option sounds very appealing.

PR #1720, opened 2017-06-02
[MRG] Introduce consume_seqfile_banding method, part duex

### LINKED ISSUE #1570: Implement a new 'consume' method that only stores a fraction of the observed k-mers

standage:
In #1571, I introduced bulk loading method with k-mer banding. Two issues though: 1) It only supported using 2^N bands; 2) The PR was against the pre-Cython codebase. I figured it would be easier to start over than merge master and fix.

This new PR splits the `uint64_t` hash space into N intervals and checks whether the hashed k-mer is in the specified interval.

Closes #1570.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


standage:
I considered adding these functions to the `MurmurHashtable` class instead of the `Hashtable` class, since the banding isn't very effective with the 2-bit encoding. I decided not to for two reasons. First, using banding on \*graph data structures won't lead to any issues with **correctness**, it just won't have the same performance improvements that the \*table data structures get. Second, the CPython code doesn't have any scaffolding for the MurmurHashtable class, and I am loathe to slog through that.

standage:
That said, ready for review and merge! @camillescott @luizirber @betatim @ctb

ctb (approved):
Looks good to me! Some tests on num bands that aren't powers of two might be good if you're feeling suspicious, though?

standage:
> Some tests on num bands that aren't powers of two might be good if you're feeling suspicious, though?

Done. Will merge when the build completes.

PR #1723, opened 2017-06-07
[WIP] Add tests for small-count vs bigcount CLI conflict

### LINKED ISSUE #1722: load-into-counting.py: ValueError: bigcount is not supported for this storage
`load-into-counting.py` with `--small-count` yields `ValueError: bigcount is not supported for this storage.`. Adding `--no-bigcount` doesn't seem to help.

```
❯❯❯ load-into-counting.py --small-count -x 1e7 khmer2 foo.fq.gz
PARAMETERS:
 - kmer size =     32 		(-k)
 - n tables =      4 		(-N)
 - max tablesize = 1e+07 	(-x)
Estimated memory usage is 0.0 Gb (4e+07 bytes = 4 bytes x 1e+07 entries / 1 entries per byte)
--------
Saving k-mer countgraph to khmer2
Loading kmers from sequences in ['foo.fq.gz']
making countgraph
  File "/Users/sjackman/.homebrew/bin/load-into-counting.py", line 221, in <module>
    main()
  File "/Users/sjackman/.homebrew/bin/load-into-counting.py", line 139, in main
    countgraph.set_use_bigcount(args.bigcount)
ValueError: bigcount is not supported for this storage.
❯❯❯ load-into-counting.py --version
khmer 2.1.1
```

betatim:
Fixes #1722 

The two arguments conflict with each other as you can not set
"bigcount" on a graph or table that uses smallcount. This adds
tests and modifies the code to check if smallcount is set or
not.

Should we add a warning message or abort if the CLI arguments are asking
for smallcount and big count at the same time? Right now it will just run and
ignore the request for big counting.

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/khmer_args.py
+++ b/khmer/khmer_args.py
@@ -569,7 +569,10 @@ def create_countgraph(args, ksize=None, multiplier=1.0, fp_rate=0.1):
     else:
         tablesize = calculate_graphsize(args, 'countgraph',
                                         multiplier=multiplier)
-        return khmer.Countgraph(ksize, tablesize, args.n_tables)
+        cg = khmer.Countgraph(ksize, tablesize, args.n_tables)
+        if hasattr(args, 'bigcount'):
+            cg.set_use_bigcount(args.bigcount)
+        return cg
 
 
 def create_matching_nodegraph(countgraph):
--- a/scripts/abundance-dist-single.py
+++ b/scripts/abundance-dist-single.py
@@ -132,7 +132,6 @@ def main():  # pylint: disable=too-many-locals,too-many-branches
     # for the countgraph and 1/(9+eps) for the tracking nodegraph
     # `eps` is used to account for the memory used by the python interpreter
     countgraph = khmer_args.create_countgraph(args, multiplier=8 / (9. + 0.3))
-    countgraph.set_use_bigcount(args.bigcount)
 
     log_info('building k-mer tracking graph')
     tracking = khmer_args.create_matching_nodegraph(countgraph)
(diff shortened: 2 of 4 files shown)


betatim:
The only CLI flags I can find related to big count are to turn it off, so I would say we don't need to emit a warning each time someone selects `--small-count` but doesn't also set `--no-bigcount`.

luizirber:
Can we make it a choice?
https://docs.python.org/3/library/argparse.html#choices

standage:
> ...so I would say we don't need to emit a warning...

I agree.

standage:
After this is merged we'll need to backport to `maint/2.1` branch.

standage:
> Can we make it a choice?

It's not clear what the choice would be. Default behavior is to keep track of counts >255 with ancillary data structure if I understand correctly. The `--no-bigcount` disables this ancillary behavior, capping abundance at 255. The `--small-count` flag uses 4 bits per counter instead of 8 bits, exchanging memory efficiency for an even lower cap. So I think the two flags would be more appropriate than something like `--count default/no-big/small`, if that's what you had in mind.

standage (approved):
Thanks @betatim!

betatim:
I like the idea of making it a `--count small/no-big/default` and we could implement it as a choice in `argparse`. It changes the exist args though :-/

betatim:
Maybe best to leave it as it is and consider this as something together with how we expose having multiple hash functions. It seems that is somehow connected as it also allows the user to configure a lot more than before.

standage:
Let's split this suggestion off to a separate issue for further discussion and consideration. In any case, for the 2.1 bugfixes we can't change the existing CLI args, so it should be handled separately anyway.

standage:
See #1724.

betatim:
Thanks!

PR #1727, opened 2017-06-07
Fix links in paper markdown

standage:
Fixes links in Markdown for JOSS v2.1 paper. See https://github.com/openjournals/joss-reviews/issues/272#issuecomment-306649961.

- [x] Is it mergeable?


sjackman:
The links don't work. 😢
```
[doi:10.12688/f1000research.6924.1](10.12688/f1000research.6924.1)
```
should be
```
[doi:10.12688/f1000research.6924.1](http://doi.org/10.12688/f1000research.6924.1)
```

standage:
Ugh, I was too hasty with the copy/pasty.

standage:
Fixed on `maint/2.1` in eb7d9509.

PR #1730, opened 2017-06-09
[MRG] Move HLLCounter into the glorious Cython future

luizirber:
- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)


khmer/_oxli/hllcounter.pxd
```diff
@@ -1,38 +1,47 @@
...
+
+cdef extern from "oxli/oxli_exception_convert.hh":
+    cdef void oxli_raise_py_error()
+
+
 cdef extern from "oxli/hllcounter.hh" namespace "oxli":
     cdef cppclass CpHLLCounter "oxli::HLLCounter":
-        CpHLLCounter(double, WordLength)
-        CpHLLCounter(int, WordLength)
+        CpHLLCounter(double, WordLength) except +oxli_raise_py_error
```
luizirber: This allows capturing oxli exceptions in C++ and converting to appropriate Python exceptions. (But maybe we can use a shorter name...)

khmer/_oxli/hllcounter.pyx
```diff
@@ -0,0 +1,104 @@
...
+        "Error rate for this HLL counter."
+        "Can be changed prior to first counting, but becomes read-only after "
+        "that (raising AttributeError)"
+        return deref(self._this).get_erate()
+
+    @error_rate.setter
+    def error_rate(self, erate):
+        deref(self._this).set_erate(float(erate))
+
+    @error_rate.deleter
```
luizirber: Added this to reproduce behavior from previous version, but I don't think anyone ever tried to delete a property out of HLLCounter in practice.

setup.py
```diff
@@ -154,15 +154,16 @@ def build_dir():
 BUILD_DEPENDS = glob.glob(path_join("include", "khmer", "_cpy_*.hh"))
 BUILD_DEPENDS.extend(path_join("include", "oxli", bn + ".hh") for bn in [
     "khmer", "kmer_hash", "hashtable", "labelhash", "hashgraph",
-    "hllcounter", "khmer_exception", "read_aligner", "subset", "read_parsers",
-    "kmer_filters", "traversal", "assembler", "alphabets", "storage"])
+    "hllcounter", "oxli_exception", "read_aligner", "subset", "read_parsers",
+    "kmer_filters", "traversal", "assembler", "alphabets", "storage",
+    "oxli_exception_convert"])
```
luizirber: I added `oxli_exception_convert.hh` to `src/include/oxli`, but the implementation (in `src/khmer/oxli_exception_convert.cc`) depends on `Python.h`. Should we keep both header and implementation inside `khmer/_oxli` instead (since they are extension code, and not necessary for the C++ library)?

src/oxli/oxli_exception_convert.cc
```diff
@@ -0,0 +1,22 @@
+#include "Python.h"
+#include <exception>
+#include <string>
+#include "oxli/oxli_exception.hh"
+#include "oxli/oxli_exception_convert.hh"
+
+
+void oxli_raise_py_error()
```
luizirber: Implementation of the conversion from C++ to Python exceptions. These are the one I used in HLLCounter, but we will need to map all the others too.

Possible problem: methods that throw the same C++ exception, but it is mapped to different Python exceptions. Didn't see any examples yet, but if it happens we might need to have more than one of this mapping functions around.

tests/test_hll.py
```diff
@@ -240,9 +240,6 @@ def test_hll_change_error_rate():
     with pytest.raises(TypeError):
         del hllcpp.error_rate
 
-    with pytest.raises(TypeError):
```
luizirber: Removed this test because it made sense for the original code, but Cython will convert integers to floats (and then print an error, because the error rate need in a range in (0, 1) but not containing any of them).

luizirber:
Ready for review and merge @camillescott @ctb @standage

ctb (approved):
Lots of unanswered questions in there but overall looks good to me - do you want to wait for @camillescott and others to address questions? Or move them over to github issues?

luizirber:
I ended up moving 'oxli_exception_convert.{hh,cc}` into `khmer/_oxli`, what do you think @camillescott?

khmer/_oxli/hllcounter.pxd
```diff
@@ -1,38 +1,47 @@
...
+                                    uint64_t &) except +oxli_raise_py_error
 
-        void consume_seqfile[SeqIO](unique_ptr[CpReadParser[SeqIO]]&,
-                                    bool,
-                                    unsigned int &,
-                                    unsigned long long &)
+#        void consume_seqfile[SeqIO](unique_ptr[CpReadParser[SeqIO]]&,
+#                                    bool,
+#                                    unsigned int &,
+#                                    unsigned long long &)
```
betatim: Would this https://github.com/dib-lab/khmer/pull/1675/files#diff-91092b4a16651bcd34d40d87061f3845R175 help with not having to comment this out?

khmer/_oxli/hllcounter.pyx
```diff
@@ -0,0 +1,103 @@
...
+    def error_rate(self):
+        raise TypeError("Cannot delete attribute")
+
+    @property
+    def ksize(self):
```
betatim: So far this has been a method you need to call instead of a property for Counttable. Should we switch all of them over? #consistency

betatim: For HLL it has always been a property, just wondering if we should use this as an excuse to switch

betatim:
Not sure what is missing here. Luiz do you want to do some more fiddling or merge?

betatim:
🎉

PR #1734, opened 2017-06-14
[MRG] Docs don't build correctly without Cython

standage:
Arguably this should be handled by running `make install-dependencies`, but since there doesn't seem to be an easy way to do so with ReadTheDocs, this PR duplicates the dependency in `doc/requirements.txt`.

- [x] Is it mergeable?


standage:
Ugh, I feel silly. I couldn't find `install_requires` and so I assumed it wasn't there. Turns out I was looking in the wrong file (`Makefile` instead of `setup.py`). I've updated to make sure Cython is installed with khmer (and removed along with screed from the development package list). Hopefully this works!

standage:
So, adding Cython to `install_requires` is The Right Thing To Do, but for some cryptic reason is insufficient to get ReadTheDocs to build correctly. So it looks like adding Cython to `doc/requirements.txt` is also necessary. I guess we can live with that if that's what it takes.

standage:
https://readthedocs.org/projects/khmer/builds/5561087/

standage:
Ready for review and merge! @camillescott @luizirber @ctb @betatim

setup.py
```diff
@@ -266,9 +266,8 @@ def build_dir():
         "packages": ['khmer', 'khmer.tests', 'oxli', 'khmer._oxli'],
         "package_data": {'khmer/_oxli': ['*.pxd']},
         "package_dir": {'khmer.tests': 'tests'},
-        "install_requires": ['screed >= 1.0', 'bz2file'],
-        "setup_requires": ["pytest-runner>=2.0,<3dev", "setuptools>=18.0",
-                           "Cython>=0.25.2"],
+        "install_requires": ['screed >= 1.0', 'bz2file', 'Cython>=0.25.2'],
```
betatim: Is cython really needed when you `pip install khmer`? I thought we were doing [this](http://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules (or planning to) which means you don't need cython to install khmer. This would require the generated .cpp files to be distributed via PyPI and I have no idea how this then works for RTD which doesn't install from PyPI :-/

standage: I don't think we're distributing `.cpp` files now. I'm not against that, but honestly I'm mostly concerned about getting the khmer and kevlar docs to build again. I'd suggest (barring any YUGE concerns) that we merge this and then open up a thread to discuss The Right Way to do it.

betatim (approved):
Need to find a packaging expert to help us out but until then: merge to get the docs building.

PR #1745, opened 2017-07-21
[MRG] Pin cython version until cython issue 1790 is resolved

betatim:
See discussion here: https://github.com/cython/cython/issues/1790

I propose we pin the version of cython for the time being. Sounds like this is a bug in cython -> behaviour in 0.25.2 is correct and will be restored in a future version.

- [x] Is it mergeable?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,8 +102,8 @@ help: Makefile
 install-dep: install-dependencies
 
 install-dependencies:
-	pip install --upgrade --ignore-installed $(DEVPKGS)
-	pip install --upgrade --requirement doc/requirements.txt
+	pip install $(DEVPKGS)
+	pip install --requirement doc/requirements.txt
 
 ## sharedobj   : build khmer shared object file
 sharedobj: $(EXTENSION_MODULE)
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -1,4 +1,4 @@
 sphinxcontrib-autoprogram==0.1.2
 setuptools>=3.4.1
 guzzle_sphinx_theme==0.7.11
-Cython>=0.25.2
+Cython==0.25.2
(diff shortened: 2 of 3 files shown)


ctb:
+1

camillescott (dismissed):
seems like a fine temp workaround to me

betatim:
[This line](https://github.com/dib-lab/khmer/blob/master/Makefile#L105) causes an upgrade of cython to 0.26. [A bit of pip logic explained](https://pip.pypa.io/en/stable/user_guide/#only-if-needed-recursive-upgrade), so I think we would have to install all the dependencies explicitly in the CI environment and not call `make dep`? A bit suboptimal

camillescott:
Seems to me like we shouldn't be upgrading packages for folks -- just
giving pip the reqs file without `--upgrade` should only upgrade the
packages that need it, no?

betatim:
As usual, I assume that `--upgrade` and `--ignore-installed` were put there for a reason...

I think without `--upgrade` `pip install -r req.txt` will update your foobar package from vX (installed on your system) to vY (minimum required by khmer) but won't move it all the way to vZ (latest).

So I think we should we should drop `--upgrade --ignore-installed`, but see first line of this comment :)

betatim:
Do we really need cython as a requirement in `docs/requirements.txt`? That seems weird no?

standage:
Yes, I was troubleshooting broken ReadTheDocs builds. See https://github.com/dib-lab/khmer/issues/1733#issuecomment-308496816.

I added it only reluctantly in the first place, would be happy to remove it if it's no longer a problem or if we find an alternative solution.

betatim:
Aha! I will pin it as well for the moment as I don't have a smart idea about RTD :-/

camillescott (approved):
looks all well and good to me now

PR #1746, opened 2017-07-25
[MRG] New consume function with mask

standage:
We are exploring some use cases in kevlar where k-mers are only stored if they are not present in another data set. This PR basically copy & pastes several of the existing bulk loading functions and adds a few lines to check against the mask. We discussed several ways forward in #1637. This is the "short term gain / long term loss" strategy that should hopefully allow us to see pretty quickly if/how this will work, but probably warrants more careful consideration for a long term solution.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?


src/oxli/hashtable.cc
```diff
@@ -115,6 +142,42 @@ void Hashtable::consume_seqfile(
...
+        } catch (NoMoreReadsAvailable) {
+            break;
+        }
+
+        read.set_clean_seq();
+        unsigned int this_n_consumed = 0;
+        KmerHashIteratorPtr kmers = new_kmer_iterator(read.cleaned_seq);
+        while(!kmers->done()) {
+            HashIntoType kmer = kmers->next();
+            if (mask->get_count(kmer) == 0) {
```
standage: This right here is the distinguishing conditional of this function.

src/oxli/hashtable.cc
```diff
@@ -528,6 +632,40 @@ template void Hashtable::consume_seqfile_banding<FastxReader>(
...
+    ReadParserPtr<FastxReader>& parser,
+    Hashtable* mask,
+    unsigned int &total_reads,
+    unsigned long long &n_consumed
+);
+
+//template void Hashtable::consume_seqfile_banding_with_mask<FastxReader>(
+//    std::string const &filename,
+//    unsigned int num_bands,
+//    unsigned int bands,
```
standage: For some reason, when this template instantiation is uncommented, clang complains about `explicit instantiation of undefined function template`. I've compared it to the other `consume_*` functions and can't see what the difference might be to cause this.

standage: Oops. Forgot the `Hashtable::` namespace prefix in the function definition. Fixed now.

standage:
Ok, this is ready for a review, and discussion about potential for a merge. cc @betatim @camillescott @luizirber @ctb

Again, there's a question of how much "debt" we want to accrue. We can't carry on proliferating consume functions like this indefinitely, but this feature would be immediately useful to me in kevlar. Eventually we'll need a more elegant solution.

camillescott:
Looks like it failed pep8 ;)

I'm kinda the lab architecture diva, but I'm coming around recently and down to merge as-is. Fix the pep8 stuff and I'll approve and merge.

standage:
@camillescott The pep8 monster has been fed another sacrifice.

standage:
Actually, hold off on this for a moment. I'm going to make a small change. Rather than ignoring k-mers with abundance > 0 in the mask, I'll introduce a parameter `t` such that k-mers with abundance > `t` are ignored.

standage:
Ok, at the C++ level there is now a parameter `threshold`. At the CPython level, the parameter is optional, set to 0 if not specified.

Should be a quick re-review @camillescott. :-)

betatim:
Why does codecov think that we have no coverage for the newly added code? There clearly are tests ... :confused:

standage:
¯\\\_(ツ)\_/¯

standage:
Review plz? :-)

standage:
@camillescott gave her informal blessing on this PR more than a week ago. I *did* make a change since then, but the change is minor and tested. I'm going to break protocol and merge this myself.

PR #1753, opened 2017-08-08
[MRG] Expose *_with_reads_parser method overloads to Python API

standage:
Each variant of the hashtable `consume_seqfile` method is overloaded to support either an input filename (provided as a `std::string`) or alternatively a read parser. However, most of these are not currently accessible via the Python API. This PR adds these to the CPython code.

While running the tests, I ran into "too many open files" errors. On further investigation, none of the existing CPython code explicitly closed the read parsers. I addressed this for both the new methods as well as the existing method.

One point of concern: the proliferation of `consume` functions is getting out of hand. We can only postpone refactoring for so long.

Ready for review and merge!

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?


include/khmer/_cpy_hashtable.hh
```diff
@@ -49,20 +49,38 @@ hashtable_consume_seqfile(khmer_KHashtable_Object * me, PyObject * args);
...
+);
+
+
+PyObject *
 hashtable_consume_seqfile_banding_with_mask(khmer_KHashtable_Object * me, PyObject * args);
 
 
 PyObject *
-hashtable_consume_seqfile_with_reads_parser(khmer_KHashtable_Object * me,
-        PyObject * args);
```
standage: Moved this function up for consistency's sake: for each overloaded method, the version that accepts a string is first, and the version that accepts a parser is second.

src/khmer/_cpy_hashtable.cc
```diff
@@ -464,52 +594,42 @@ hashtable_consume_seqfile_banding_with_mask(khmer_KHashtable_Object * me, PyObje
 }
 
 PyObject *
-hashtable_consume_seqfile_with_reads_parser(khmer_KHashtable_Object * me,
-        PyObject * args)
+hashtable_consume_seqfile_banding_with_mask_with_reads_parser(khmer_KHashtable_Object * me, PyObject * args)
```
standage: Looks like git is struggling with the similarity between all of the variants of this function and the fact that I cut n pasted this function further up, again for consistency.

standage:
Hey @camillescott and @ctb: I'm not grokking the errors from [the latest CI build](https://travis-ci.org/dib-lab/khmer/jobs/270131685) for this thread. Do you think this has to do with CPython inheritance issues?

standage:
With most recent Cython PRs merged, closing in favor of #1785.

PR #1784, opened 2017-09-13
Convert to unordered_map where appropriate

camillescott:
`std::unordered_map` has O(1) amortized random access; `std::map` is O(log(n)), due to maintaining internal ordering of its elements. Most of our use cases don't require that ordering, so this switches them to `unordered_map` where appropriate. On map sizes on the order of 5 million elements, this is a 20x speedup for element access. For large datasets using `bigcount`, I imagine the benefit will be... considerable.

cc @ctb @luizirber @betatim 

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/include/oxli/hashgraph.hh
+++ b/include/oxli/hashgraph.hh
@@ -43,7 +43,6 @@ Contact: khmer-project@idyll.org
 #include <fstream>
 #include <iostream>
 #include <list>
-#include <map>
 #include <queue>
 #include <set>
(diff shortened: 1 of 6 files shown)


standage:
I assume this is ready for review @camillescott?

camillescott:
Yupper, ready to roll

Issue #1751, opened 2017-08-01
Bulk loading fails with QFCounttable


standage:
I'm getting the following error message on Mac OS X.

```
>>> import khmer
>>> ct = khmer.QFCounttable(31, 2**16)
>>> ct.consume_seqfile('kevlar/notebook/human-sim-pico/father-reads-dump.fq.gz')
Assertion failed: (start_index <= empty_index && empty_index < qf->xnslots), function shift_remainders, file gqf.c, line 463.
Abort trap: 6
```

betatim:
I think this is the "QF is too full to continue" error message. We discussed this in https://github.com/dib-lab/khmer/pull/1675#issuecomment-310657888 and follow ups. You can keep filling a BF beyond reasonable limits and it won't ever crash. The QF data structure thinks that you should stop. The recommended course of action is to grow the QF, which is reasonable unless you are already using all the RAM you can.

Mostly a question of "better error message needed" and a good idea what to tell the user to do beyond "we are full, buy more RAM"

betatim:
I think this is the relevant comment https://github.com/splatlab/cqf/issues/4#issuecomment-310457762 for figuring out whether or not the QF is about to be full.

standage:
The thing is, the error message happens immediately, I might even guess on the first attempt to insert. I'll be honest and say I don't even know how big the table I've requested is (hence #1748).

standage:
Ok, the example above handles a Fastq file with 5 reads just fine. It still fails on a 13 Mb gzipped Fastq file, which is much bigger but of course orders of magnitude smaller than real Fastq files out in the wild.

I can increase the starting memory for a while, until I run into this error message.

```
>>> ct = khmer.QFCounttable(31, 2**32)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "khmer/_oxli/graphs.pyx", line 255, in khmer._oxli.graphs.QFCounttable.__cinit__ (khmer/_oxli/graphs.cpp:5067)
    def __cinit__(self, int k, int starting_size):
OverflowError: value too large to convert to int
```

We probably want to define this as an `unsigned long int`.

standage:
I modified the `.pyx` file and changed `int starting_size` to `unsigned long int starting_size`. This allows me to request larger starting sizes. But when I run the code above with starting size of `2**6` the process consumes > 10Gb of memory (paging to the SSD on my laptop) and then *still* fails with the Assertion error.

ctb:
[...]

PR #1789, opened 2017-09-17
[MRG] Switch QFCounttable to murmurhash

betatim:
Switch QFCounttable to use murmur hash. This makes it somewhat more predictable when the QF will be "full", see https://github.com/dib-lab/khmer/issues/1751#issuecomment-329535858

Should we modify the CQF code to raise an exception instead of using an assert or is there a way to "catch" an assert? This way we can print a better message for the user when we do exhaust the available slots. An alternative would be to check every time we insert a kmer if all slots are used :-/

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  changes were made?

### DIFF
--- a/include/oxli/hashtable.hh
+++ b/include/oxli/hashtable.hh
@@ -556,11 +556,11 @@ public:
 };
 
 // Hashtable-derived class with QFStorage.
-class QFCounttable : public oxli::Hashtable
+class QFCounttable : public oxli::MurmurHashtable
 {
 public:
     explicit QFCounttable(WordLength ksize, int size)
-        : Hashtable(ksize, new QFStorage(size)) { } ;
+        : MurmurHashtable(ksize, new QFStorage(size)) { } ;
 };
 
 // Hashtable-derived class with BitStorage.


src/oxli/kmer_hash.cc
```diff
@@ -146,29 +151,16 @@ std::string _revhash(HashIntoType hash, WordLength k)
...
+
+    auto from = out.begin();
+    auto to = out.end();
+
+    size_t len = to - from;
+    char c;
+    for (to--; from <= to; from++, to--) {
+        c = tbl[(int)*from];
+        *from = tbl[(int)*to];
+        *to = c;
```
betatim: @luizirber this comes from https://benchmarksgame.alioth.debian.org/u64q/program.php?test=revcomp&lang=gcc&id=2 and speeds things up a lot for me!

src/oxli/kmer_hash.cc
```diff
@@ -49,6 +49,11 @@ Contact: khmer-project@idyll.org
 
 using namespace std;
 
+#define tbl \
+  "                                                                "\
+  /*ABCDEFGHIJKLMNOPQRSTUVWXYZ      abcdefghijklmnopqrstuvwxyz    */\
+  " TVGH  CD  M KN   YSAABW R       TVGH  CD  M KN   YSAABW R"
```
standage: Is the leading whitespace intentional?

betatim: yes, you want to 'pad' the 'array' so that `tbl[ord("A")]` gives you the translation

betatim:

betatim:
I think this will fix #1751

betatim:
Woah! We might approach this being useful ;)

PR #1795, opened 2017-10-06
[MRG] add a test to make sure that LinearAssembler.assemble takes hashes

### LINKED ISSUE #1765: Should `LinearAssembler::assemble` be able to take a hash value?
I just debugged an issue where the behavior of `LinearAssembler.assemble()` changed -- I had previously been using hash values to seed assemblies of linear paths, but in recent versions of khmer master, `LinearAssembler` has changed to want either `Kmer` or DNA strings - and `assemble` now fails silently/returns nothing if it's not a Kmer and cannot be converted into a meaningful DNA string (e.g. `str(hash)` fails).

I don't think failing silently is a great idea :), but I'm also wondering if there's a good reason to require either `Kmer` or DNA strings here. Hash values are usable in many other functions on graphs and get automatically converted.

If we decide to allow hash values we should make sure to add that into the tests, too.

cc @camillescott

ctb:
I was unable to find a test that checked to make sure that `LinearAssembler.assemble(...)` can take a hash (see #1765) so I wrote one.

Fixes #1765 (or at least makes sure it stays fixed :).

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/tests/test_assembly.py
+++ b/tests/test_assembly.py
@@ -123,6 +123,15 @@ class TestLinearAssembler_RightBranching:
         assert len(path) == HDN.pos + K
         assert utils._equals_rc(path, contig[:len(path)])
 
+    def test_assemble_takes_hash(self, right_tip_structure):
+        # assemble from beginning of contig, up until branch point
+        graph, contig, L, HDN, R, tip = right_tip_structure
+        asm = khmer.LinearAssembler(graph)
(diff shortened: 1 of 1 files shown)


ctb:
Ready for review & merge!

betatim:
Beyonce strategy: if you liked it you should have put a test on it ;)

PR #1796, opened 2017-10-06
[MRG] Remove clearing of sequence

betatim:
Clearing the string containing the sequence is not needed and was
happening too early (before the end of the loop).

Noticed this because when working on the rolling hash. For one of the unit tests the first character in the sequence would suddenly be `\0` which "breaks everything".

Any ideas how we can test for this explicitly? I don't :(

- [x] Is it mergeable?
  `scripts/` is it tested?

### DIFF
--- a/src/oxli/hashtable.cc
+++ b/src/oxli/hashtable.cc
@@ -453,9 +453,6 @@ uint64_t * Hashtable::abundance_distribution(
 
     Read read;
 
-    string name;
-    string seq;
-
     // if not, could lead to overflow.
     if (sizeof(BoundedCounterType) != 2) {
         delete[] dist;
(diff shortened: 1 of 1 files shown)


src/oxli/hashtable.cc
```diff
@@ -482,9 +478,6 @@ uint64_t * Hashtable::abundance_distribution(
                 BoundedCounterType n = get_count(kmer);
                 dist[n]++;
             }
-
-            name.clear();
-            seq.clear();
```
betatim: `KmerHashIteratorPtr` has a pointer to `seq` which we invalidate by calling `clear()`. The standard apparently doesn't require that `clear` actually goes and sets all characters to `\0` so all it does is set the first one to `\0`. Not sure why we are calling `clear()` in the first place. If at all it should be just after the `while` loop. I think we never noticed before because we don't "subtract" the first character to compute the hash of the second kmer. This is new in the rolling hash ... was pretty puzzled as to why sometimes "AAAAAAAA" would result in a hash that changes as you iterate through the 4-mers in that sequence ... :)

standage: Yeah, I can't see any reason why this should be here.

betatim:
🍻

PR #1797, opened 2017-10-06
[MRG] Stopgap for conditional cython import

### LINKED ISSUE #1783: Issues with khmer as a dependency unless Cython explicitly listed as sub-dependency
Cython is not a run-time dependency for khmer, nor strictly a build-time dependency. Only developers that modify the Cython code need to be able to convert the Cython to C++ that can then be distributed and built as part of the normal installation process, without Cython.

That's the idea at least. In practice, it's not working out quite like that. We've had to add Cython to `doc/requirements.txt` to get the ReadTheDocs build to complete successfully. And now, I'm encountering another problem with kevlar citing khmer as a dependency. I've noticed that whenever I install kevlar on a fresh system, I can't just `pip install cython networkx pysam git+https://github.com/dib-lab/khmer.git`, I have to do `pip install cython` and then `pip install git+https://github.com/dib-lab/khmer.git` in a subsequent step. If not, all of the dependencies get treated equally, and for some reason khmer is always processed before Cython and we end up with this.

```
  File "<string>", line 1, in <module>
  File "/tmp/pip-i4qz_p93-build/setup.py", line 230, in <module>
      EXTENSION_MODS = cythonize(EXTENSION_MODS,
  NameError: name 'cythonize' is not defined
```

[asciicast](https://asciinema.org/a/137583)

In the khmer ReadTheDocs build, we got around this by adding Cython to the `doc/requirements.txt` so that the requirement is processed and installed before any khmer code is looked at. But in the kevlar ReadTheDocs build, this isn't possible. I have to place Cython and kevlar in the same list of dependencies, and khmer gets processed first. I can even add Cython as a dependency to kevlar, but khmer will still fail and bork the entire build.

Is there any way to fix this without adding Cython as an explicit dependency to khmer? Are there any arguments against doing so?

standage:
The current `setup.py` file is a bit schizophrenic with respect to Cython. The `cythonize` function is conditionally imported, but is unconditionally called. This makes it impossible for packaging and installation systems without cython installed to even parse the file without failing.

This pull request introduces a temporary fix to this problem by making the call to `cythonize` conditional and then adding Cython as an explicit setup requirement.

Eventually we'll want to do this more cleanly, but for now this PR fixes #1783 and similar problems where calling Cython without making sure it's installed can be problematic.

- [x] Is it mergeable?
  changes were made?

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -64,14 +64,14 @@ ez_setup.use_setuptools(version="3.4.1")
 
 CMDCLASS = versioneer.get_cmdclass()
 
-HAS_CYTHON = False
 try:
     import Cython
     from Cython.Build import cythonize
     HAS_CYTHON = True
+    cy_ext = 'pyx'
 except ImportError:
-    pass
-cy_ext = 'pyx' if HAS_CYTHON else 'cpp'
+    HAS_CYTHON = False
+    cy_ext = 'cpp'
 
 # strip out -Wstrict-prototypes; a hack suggested by
 # http://stackoverflow.com/a/9740721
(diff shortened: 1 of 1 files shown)


standage:
Ready for review! @luizirber @betatim @camillescott @ctb

betatim:
Comparing to http://cython.readthedocs.io/en/latest/src/reference/compilation.html#distributing-cython-modules this looks like a permanent/recommended fix to me.

Issue #1805, opened 2017-10-23
performance regression in Python call to `traverse_linear_path`


ctb:
I put some time and effort into tracking down an apparent slowdown in cDBG construction; it appears to be within this function, https://github.com/dib-lab/khmer/blob/master/khmer/_oxli/graphs.pyx#L489 - could be beneath it, not sure yet.

I see a factor of 5-10x slowdown when using this for spacegraphcats.

@camillescott et al could the `cdef set` stuff be causing lots of C <-> Python copying?

ctb:
Note: there is no difference in the underlying function in `src/oxli/hashgraph.cc` which makes me suspect it is in the Cython wrapper.

PR #1808, opened 2017-10-24
[MRG] Some performance hacks and minor fixes on graph interface

### LINKED ISSUE #1806: graph.neighbors() returns strings in latest master
...instead of hashes.

ctb:
Some spotty patches to keep objects from being evaluated or converted prematurely; fixes a major performance regression noted in spacegraphcats (#1805).

* make graph.neighbors(...) return Kmers;
* traverse_linear_path now returns HashSets rather than sets;
* HashSet __contains__ now uses std::set::find instead of full enumeration;

Fixes #1806 #1805 #1807.

- [x] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/khmer/_oxli/graphs.pyx
+++ b/khmer/_oxli/graphs.pyx
@@ -439,7 +439,7 @@ cdef class Hashgraph(Hashtable):
     def neighbors(self, object kmer):
         '''Get a list of neighbor nodes for this k-mer.'''
         cdef Traverser traverser = Traverser(self)
-        return [str(n) for n in traverser._neighbors(self._build_kmer(kmer))]
+        return list(traverser._neighbors(self._build_kmer(kmer)))
 
     def calc_connected_graph_size(self, object kmer, max_size=0,
                                   break_on_circumference=False):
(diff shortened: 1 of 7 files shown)


ctb:
OK, I think this is ready for review @luizirber @camillescott @betatim @standage . The extra uncovered line(s) are only in reporting/output.

khmer/_oxli/graphs.pyx
```diff
@@ -439,7 +439,7 @@ cdef class Hashgraph(Hashtable):
     def neighbors(self, object kmer):
         '''Get a list of neighbor nodes for this k-mer.'''
         cdef Traverser traverser = Traverser(self)
-        return [str(n) for n in traverser._neighbors(self._build_kmer(kmer))]
+        return list(traverser._neighbors(self._build_kmer(kmer)))
```
camillescott: I've been trying to push patterns like this toward iterator interfaces with the Cython stuff. Not a deal breaker, but something to consider.

src/oxli/hashgraph.cc
```diff
@@ -828,14 +828,16 @@ const
     while(!kmers.done()) {
         n++;
         if (n % 10000 == 0) {
-            std::cout << "... find_high_degree_nodes: " << n << "\n";
-            std::cout << std::flush;
+            std::cout << "\r... find_high_degree_nodes: " << n;
         }
         Kmer kmer = kmers.next();
         if ((traverser.degree(kmer)) > 2) {
```
camillescott: So, depending on how we want to define a cDBG, this could introduce a bug. Consider:
```
(HDN)--->(neighbor_A)---...
     \-->(neighbor_C)---...
```
This is a degree 2 node, but it should probably be counted as a HDN in the context of a cDBG (in the context of my streaming compactor, I'm counting it as one).

Maybe this deserves its own issue/thread. Thoughts?

khmer/_oxli/graphs.pyx
ctb: hey maaan I'm just riffing off your code :).

yes, on board with that.

ctb:
thankee @camillescott !

src/oxli/hashgraph.cc
ctb: Yep! Punting to issue #1809.

Issue #1810, opened 2017-10-27
pip install not compiling correctly


arivers:
I installed khmer via pip a few weeks ago without any problem but when I try today I get the error:
 
```
 gcc -pthread -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -fPIC -DVERSION=2.1 -DSEQAN_HAS_BZIP2=1 -DSEQAN_HAS_ZLIB=1 -UNO_UNIQUE_RC -Ilib -Ithird-party/zlib -Ithird-party/bzip2 -Ithird-party/seqan/core/include -Ithird-party/smhasher -I/home/adam.rivers/.conda/envs/vicaenv/include/python3.5m -c khmer/_khmer.cc -o build/temp.linux-x86_64-3.5/khmer/_khmer.o -O3 -std=c++11 -pedantic -fopenmp
  cc1plus: warning: command line option "-Wstrict-prototypes" is valid for Ada/C/ObjC but not for C++
  cc1plus: error: unrecognized command line option "-std=c++11"
  error: command 'gcc' failed with exit status 1
```

I'm using gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-16) on centos-release-6-7.el6.centos.12.3.x86_64

ctb:
Hi @arivers your message got truncated!

arivers:
Khmer 2.1.1 does compile with pip when using clang (clang-800.0.42.1) on OSX 10.13.

arivers:
Hi Titus,

I edited my original post to add additional information about the compilation error. Let me know if  you need anything more.

ctb:
thank you!

betatim:
[...]

PR #1811, opened 2017-11-01
[MRG] Mention C++11 requirement

betatim:
In response to #1810. Make a more prominent mention of C++11 and what that means in terms of compiler version.

- [x] Is it mergeable?

### DIFF
--- a/doc/user/install.rst
+++ b/doc/user/install.rst
@@ -40,8 +40,8 @@
 Installing and running khmer
 ============================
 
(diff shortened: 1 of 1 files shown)


standage:
Daheck is going on with Docker and our Travis builds?!?!

betatim:
I think it is trying to tell us "something timed out", at least on other repos where we use docker that seems to be a thing at the moment. I restarted the builds, if that doesn't fix it I'll actually think about the error ;)

PR #1813, opened 2017-11-03
Adjust print statement in check_space

souravsingh:
- [x] Is it mergeable?
- [x] `make test` Did it pass the tests?
- [x] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested?
- [x] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
  formatted?
- [ ] Did it change the command-line interface? Only backwards-compatible
  additions are allowed without a major version increment. Changing file
  formats also requires a major version number increment.
- [ ] For substantial changes or changes to the command-line interface, is it
  documented in `CHANGELOG.md`? See [keepachangelog](http://keepachangelog.com/)
  for more details.
- [x] Was a spellchecker run on the source code and documentation after
  changes were made?
- [x] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)

PR #1814, opened 2017-11-07
[MRG] Debugging docker

betatim:
Trying to find out why docker still fails ...

- [x] Is it mergeable?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -87,9 +87,6 @@ script:
   - make py-demos
   - make cpp-demos
   - cd examples/stamps && ./do.sh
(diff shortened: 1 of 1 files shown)


betatim:
Removing the docker build from travis. It seems we can't find a way that works reliably with the development model we have. Most of us push changes to branches which are in this repository, however outsiders keep the branch in their fork. I think this is the reason for the failures in #1813.

Right now no one can remember why we build the container so I am proposing we remove it (and add it again later when we need it) as right now it causes us a lot of trouble and it isn't super easy to fix.

betatim:

ctb:
Thanks!

Issue #1800, opened 2017-10-19
Idea: creating N banded count tables in a single pass over reads


standage:
Accumulate hashed k-mers in some kind of buffer<sup>1</sup>. The first time the buffer reaches a user-specified size, go through each band one-by-one, create a Counttable for the band, insert the appropriate hashed k-mers into the Counttable, and save to disk. Once all Counttables are saved, clear the buffer and continue. Each subsequent time the buffer reaches the user-specified size, load the Counttables from disk one-by-one, insert the appropriate hashed k-mers, and re-write to disk.

Peak memory usage should be the size of the buffer + the size of one Counttable.

Please tell me how irresponsible and inefficient this is going to be! :-)

------

<sup>1</sup> Probably implemented as a `std::unordered_map< int, std::list<HashIntoType> >`, where the keys are band indices and values are lists of hashed k-mers.

drtamermansour:
I think it would be more efficient if you save your Counttables of each buffer in a temp space then do one final merge step of all Counttables. This should save the time of reading and writing Counttables several times to the desk. As usual, I am copying this from Jellyfish which is my raw model for mixing between in-memory and on-desk data structures to control the peak memory usage.

standage:
Forgive me while I ask for a couple clarifications.

> ...save your Counttables of each buffer in a temp space...This should save the time of reading and writing Counttables several times to the disk.

I'm not sure I understand "temp space". Is it in memory or disk?

- If it's memory, that's problematic. The idea is to have only one Counttable in memory at a time.
- If it's on disk, how does this differ from the approach I have suggested?

> ...then do one final merge step of all Counttables.

My intent is actually not to merge the Counttables, but to do N parallel operations independently on different subsets of the data.

drtamermansour:
I am mean desk temp space.

You says

> Once all Counttables are saved, clear the buffer and continue. Each subsequent time the buffer reaches the user-specified size, load the Counttables from disk one-by-one, insert the appropriate hashed k-mers, and re-write to disk

My suggested change has 2 advantages: 
1) smaller size of the Counttables that you load and write with each step
2) you can multi-thread the process efficiently 

I will try to model your algorithm "according to my understanding":
let us assume that 
1. You have a single band
2. The available buffer size would allow reading (n) hashed kmers (hk) that belong to this band.
3. You will insert the hashed kmers in a Counttable (C)
4. You need to repeat the cycle of "load the Counttable, insert hashed kmers, and re-write" (z) times

[...]

PR #1818, opened 2017-11-16
[WIP] Prototype: count k-mers in banded mode in a single pass

standage:
This PR is a Python prototype of the idea described in #1800. In brief, the idea is to count k-mers in the input reads and populate N counttables (each with ≈1/N of the k-mers) in a single pass.

This prototype implements a k-mer buffer that, when "full", writes to disk and flushes/resets. Once all the input reads are processed, the temp files are read back from disk 1-by-1 to populate a Counttable for each band.

The prototype works as expected, but the performance is painfully slow on large data sets. It would need a Cython or C++ implementation with threading support to be viable methinks. I would welcome any thoughts. cc @camillescott @betatim @ctb @luizirber

- [ ] Is it mergeable?
  `scripts/` is it tested?
  formatted?
  for more details.
  changes were made?
  tested for streaming IO?)

### DIFF
--- a/sandbox/count-band-single-pass.py
+++ b/sandbox/count-band-single-pass.py
@@ -1,0 +1,117 @@
+#!/usr/bin/env python
+
+import argparse
+import os
+from math import floor
+import pickle
+import sys
+
+import khmer
+
+
+class BandedHashBuffer(object):
+    """
+    Store k-mers in separate bands, using disk to keep memory usage low.
+
+    The `self.hashlists` variable is the primary member. It is a dictionary
+    that stores a list of hash values for each band.
+
+    When a hash value is added to the buffer, it is stored in the list
+    corresponding to the appropriate band. Once the buffer accumulates a user-
+    specified number of hash values, each list is serialized and written to
+    temporary files on disk, and the buffer is flushed and reset.
+
+    Once all the input is consumed, the temporary files are re-opened and used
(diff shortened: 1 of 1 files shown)


standage:
```
curl -L https://osf.io/f5trh/download?version=1 -o reads.fq.gz
sandbox/count-band-single-pass.py \
    --ksize 25 --num-bands 4 --buffersize 1e7 --memory 1e7 \
    --outfmt "bandtest{}.ct" \
    reads.fq.gz
```

The procedure above will result in the same output as the Python commands below.

```python
for i in range(4):
    counts = khmer.Counttable(25, 1e7 / 4, 4)
    counts.consume_seqfile_banding('reads.fq.gz', 4, i)
    of = 'bandtest{}.ct'.format(i)
    counts.save(of)
```

betatim:
For my education: why split into bands but then create N counttables? I understand "let's make N bands because we have far too much data anyway" but that doesn't fit with then making N bands. Tim is confused.com

standage:
We are investigating some use cases where we compute only on a single band, but there is at least one use case (motivated by kevlar and variant discovery) where we want to compute on each band independently. In this case, the banding provides us a way of randomly sharding the k-mers into N disjoint sets, each of which only requires 1/N of the memory required to store the entire data set.

This PR is looking for an efficient way to compute counts for all bands in a single first pass, rather than N passes.

PR #1824, opened 2017-12-27
[MRG] Unicode --> ASCII for contributor names

standage:
I hate that it has come to this, but it looks like unless we want to do a lot of troubleshooting and hacking on pip itself, the easiest way to get rid of khmer's build issues with the Python3 toolchain is to remove non-ASCII characters from the author list (temporarily hopefully). See https://github.com/dib-lab/khmer/issues/1565#issuecomment-354156725.

- [x] Is it mergeable?

### DIFF
--- a/authors.csv
+++ b/authors.csv
@@ -5,7 +5,7 @@ Alex Hyer,theonehyer@gmail.com
 Alexander Johan Nederbragt,lex.nederbragt@ibv.uio.no
 Ali Aliyari,aaliyari@ucdavis.edu
 Amanda Charbonneau,charbo24@msu.edu
-Andreas Härpfer,ahaerpfer@gmail.com
+Andreas Harpfer,ahaerpfer@gmail.com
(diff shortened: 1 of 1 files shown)


standage:
Ready for review and merge! @ctb @betatim @luizirber @camillescott

authors.csv
```diff
@@ -5,7 +5,7 @@ Alex Hyer,theonehyer@gmail.com
 Alexander Johan Nederbragt,lex.nederbragt@ibv.uio.no
 Ali Aliyari,aaliyari@ucdavis.edu
 Amanda Charbonneau,charbo24@msu.edu
-Andreas Härpfer,ahaerpfer@gmail.com
+Andreas Harpfer,ahaerpfer@gmail.com
```
betatim: "ae" instead of a. That is a kinda standard hack in Germany for when you need to create an Umlaut but have a US keyboard. Looks like from the email address it is what he does too.

PR #1859, opened 2018-05-02
CQF is replaced by MQF

shokrof:
MQF, Mixed Quotient Filter, is a new variant of [CQF](https://github.com/splatlab/cqf). MQF has changed the counting eschme of the parent data structure by adding fixed size counters and using different encoding for the dynamic one.
This improvement makes MQF more memory efficient for a wider range of zipfian distributions. Also, MQF expanded the metadata component of the filter to enable tagging of its kmers. In addition to the basic functions provided by the original CQF, the new datastructure comes with an expanding list of kmer processing functions including functions for merging and resizing the filter, functions for comparing kmer sets, and functions for indexing and meta-analysis of  multiple kmer groups. More details about these functions can be found [here](https://github.com/dib-lab/khmer/tree/MQFIntegration/third-party/mqf)   
- [x] Is it mergeable?
- [x] `make test` Did it pass the tests?
- [x] `make clean diff-cover` If it introduces new functionality in
  `scripts/` is it tested? No new functionality is added
- [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
  formatted? 
      **Khmer source files are changed after running this command.**
- [x] Did it change the command-line interface? Only backwards-compatible
  additions are allowed without a major version increment. Changing file
  formats also requires a major version number increment. 
- [x] For substantial changes or changes to the command-line interface, is it
  documented in `CHANGELOG.md`? See [keepachangelog](http://keepachangelog.com/)
  for more details.
- [x] Was a spellchecker run on the source code and documentation after
  changes were made?
- [x] Do the changes respect streaming IO? (Are they
  tested for streaming IO?)

PR #1865, opened 2018-05-19
switch to using c++ compiler instead of g++ to fix cpp-demos on mac os x

ctb:
Try fixing compile errors for `cpp-demos` on Mac OS X, ref https://github.com/dib-lab/khmer/pull/1859#issuecomment-390262027, by switching to using the system C++ rather than g++.

This will be a pull request into #1859 after Travis finishes checking it.

I've also updated from the latest master branch, so this will update `MQFintegration` to be mergable.


ctb:
Hmm, switching to pull into master so that we get continuous integration goin'.

ctb:
(closing and reopening seems to be the easiest way to trigger a new build)

ctb:
hi again @shokrof, suggest opening yet another new pull request for you to experiment, rather than using this one. I'd like to avoid confusing each other with weird test changes :).  See also https://github.com/dib-lab/khmer/pull/1859#issuecomment-390642149

shokrof:
I am thinking of developing a bot to brute force all the options for the compiler. It may have better intuition than me.

shokrof:
Bug is fixed in this pull request.
https://github.com/dib-lab/khmer/pull/1873

PR #1867, opened 2018-05-25
[MRG] "consume_masked" mode for consume_seqfile_with_mask

standage:
In branch `feature/mask-comp` I implemented a new `complement` option for the `consume_seqfile_with_mask` method. By default, the function will not count any k-mers present in the mask, but with `complement=True` the function will count *only* the k-mers present in the mask/filter.

In that branch I also putzed around with some Python code (utils and scripts) to get that functionality working in `abundance-dist-single.py`.

This PR splits the C++/Cython changes from the Python changes, the latter of which aren't ready for review yet. The C++/Cython functionality is blocking my progress on a feature in kevlar, whereas I don't have a pressing need to bake this in to khmer scripts at the moment.

      changes were made?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,16 +1,27 @@
-- [ ] Is it mergeable?
-- [ ] `make test` Did it pass the tests?
-- [ ] `make clean diff-cover` If it introduces new functionality in
-  `scripts/` is it tested?
-- [ ] `make format diff_pylint_report cppcheck doc pydocstyle` Is it well
-  formatted?
-- [ ] Did it change the command-line interface? Only backwards-compatible
-  additions are allowed without a major version increment. Changing file
-  formats also requires a major version number increment.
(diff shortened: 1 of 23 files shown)


ctb:
Hi @standage the code looks good but I am not sure why you chose `complement` here. I can't find a definition of the word that makes sense with the usage here - help?

standage:
I explained this briefly in the first paragraph of the PR, but I am open to using different terminology for the variable names if it communicates intent more clearly.

The old behavior of `consume_seqfile_with_mask` is IMO pretty self explanatory: consume all k-mers except those in the mask. The new behavior is an option that complements this behavior, consuming all k-mers that *are* in the mask.

Let me know what you think regarding terminology.

ctb:
I finally remembered to search for "set complement" and that is the right definition :). Glad I found it! What about `consume_masked=True|False` (False by default)?

standage:
> What about `consume_masked=True|False` (False by default)?

Roger that. I like it.

standage:
Thank *you*

PR #1869, opened 2018-06-01
[MRG] Endow *table sketch types with 'primes' argument

standage:
The `primes` argument of Countgraph and Nodegraph is useful for creating a sketch with the same exact size as another sketch. Somehow this functionality never got ported over to the *table sketch types, and this PR fixes that oversight.

Ready for review and merge!

----------------------------------------

      changes were made?

### DIFF
--- a/khmer/_oxli/graphs.pyx
+++ b/khmer/_oxli/graphs.pyx
@@ -391,31 +391,43 @@ cdef class QFCounttable(Hashtable):
 
 cdef class Counttable(Hashtable):
 
-    def __cinit__(self, int k, uint64_t starting_size, int n_tables):
-        cdef vector[uint64_t] primes
+    def __cinit__(self, int k, uint64_t starting_size, int n_tables,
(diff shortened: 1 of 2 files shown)


khmer/_oxli/graphs.pyx
```diff
@@ -391,31 +391,43 @@ cdef class QFCounttable(Hashtable):
 
 cdef class Counttable(Hashtable):
 
-    def __cinit__(self, int k, uint64_t starting_size, int n_tables):
-        cdef vector[uint64_t] primes
+    def __cinit__(self, int k, uint64_t starting_size, int n_tables,
+                  primes=[]):
```
luizirber: I'm not sure how Cython treats this, but in Python defining arguments as lists or dicts is quite dangerous (they get allocated at definition time, and not in every function call). I suggest setting it to `None` and doing a check in the body:
```
if primes is None:
    primes = []
```

PR #1871, opened 2018-06-15
use full path for temporary filename

olgabot:
In `trim-low-abund.py`, use the full path of the input filename rather than only the basename, in case of identical basenames which causes errors.

- [x] ~Is any new functionality in tested? (This can be checked with
      generated following a successful CI build.)~
      changes were made?
- [x] Have any changes to the command-line interface been explicitly described?
      Only backwards-compatible additions are allowed without a major version
      increment. Changing file formats also requires a major version number
      increment.
- [ ] Have any substantial changes been documented in `CHANGELOG.md`? See
      [keepachangelog](http://keepachangelog.com/) for more details.
- [x] Do the changes respect streaming I/O? (Are they tested for streaming I/O?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ under semantic versioning, but will be in future versions of khmer.
   or generators where appropriate.
 - All constructors have been removed from khmer/__init__.py.
 - GraphLabels does not inherit from Hashgraph.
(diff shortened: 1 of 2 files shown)


standage:
Closes #1870.

standage:
Thanks for you contribution @olgabot!

PR #1874, opened 2018-06-18
[WIP] Experiment compacting counttables into a nodetable given a max abundance criterion

standage:
Scenario: I want to query some big ol' counttables, and my k-mers of interest are those that are abundance <= X. Rather than keeping several of these big counttables in memory at once, I'd like to be able to initialize and update a nodetable (with the same number of buckets) to indicate which k-mers satisfy the given criterion. That way I only have to keep one counttable in memory at a time.

This PR introduces two new methods for nodetables: `compose_init`, and `compose_update`. These methods do bit flipping directly on the byte array.

The build is currently failing. I'm guessing there's an issue at the C++ level, and maybe at the Cython level as well. This could really benefit from a preliminary review from @luizirber or @camillescott.

----------------------------------------

      changes were made?
- [ ] Have any substantial changes been documented in `CHANGELOG.md`? See
      [keepachangelog](http://keepachangelog.com/) for more details.

### DIFF
--- a/include/oxli/hashtable.hh
+++ b/include/oxli/hashtable.hh
@@ -624,6 +624,24 @@ class Nodetable : public oxli::MurmurHashtable
 public:
     explicit Nodetable(WordLength ksize, std::vector<uint64_t> sizes)
         : MurmurHashtable(ksize, new BitStorage(sizes)) { } ;
+
+    void compose_init(Counttable& other, BoundedCounterType max)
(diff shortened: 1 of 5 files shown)


Westfargo0147 (changes requested):
Thank

Westfargo0147 (changes requested):
Fix

PR #1884, opened 2018-12-31
Fix build

standage:
The current build process checks for Cython and will not crash during build if Cython is not available. However, if Cython is not installed previously, none of the Cython code gets built and `import khmer` fails once the build is complete.

This PR swings the pendulum the other way, and doesn't even try to build the Cython code. It directly compiles the Cythonized `.cpp` files.

Good news: this means that khmer will actually build correctly when khmer latest is pip-installed in a clean environment (for example, in a [ReadTheDocs build](https://github.com/dib-lab/kevlar/blob/2cfb9782297aa380b2074c3116e4633f512d0ffe/docs/requirements.txt) where Cython cannot be installed before khmer).

Bad news: Any updates to the Cython code now have to be manually Cythonized to C++ to be propagated. This should be fairly easy to automate though.

For more discussion see #1733. There are lots of things we can and should do to clean up the build, but all are daunting thankless tasks that we would all rather avoid, myself included.

Ready for review! @ctb @luizirber @camillescott


standage:
Kevlar currently has this branch listed as a dependency. Could I get a review so I can refer to the master branch for an upcoming release? Thanks

standage:
I am merging this pull request. I will create an issue to describe what should ideally be done.

